PR #941, opened 2026-06-05
[PR #616] feat(jira): auto-discover projects — remove mandatory jira_project_keys

il10241024:
> 🔗 **Mirrored PR** [cyberfabric/cyber-insight#616](https://github.com/cyberfabric/cyber-insight/pull/616) | **Author:** mozhaev-dev | **Opened:** 2026-06-02T14:53:33Z | **Status:** open
> *GitHub API does not allow setting PR author or timestamps — attribution preserved here.*

---

## Problem
The Jira connector required `jira_project_keys` (comma-separated list) to be manually maintained in the K8s secret. As Jira projects are frequently added/removed, this became operationally painful.

## Solution
Switch `jira_issue` stream to **SubstreamPartitionRouter**: on every sync the connector queries `/rest/api/3/project/search` to get all accessible project keys, then runs one JQL request per project:

```
project = "<KEY>" AND updated >= "..." AND updated <= "..."
```

New projects are picked up automatically on the next scheduled sync. Deleted/archived projects simply disappear from the parent stream and are no longer queried.

## Changes
| File | What changed |
|------|-------------|
| `connector.yaml` | Added `SubstreamPartitionRouter` with inline `jira_project_keys` parent stream to `jira_issue.retriever`; updated JQL to `stream_slice.project_key`; removed `jira_project_keys` from spec |
| `descriptor.yaml` | Removed `jira_project_keys` from `secret.required_fields` |
| `jira.yaml.example` | Removed `jira_project_keys` entry |

## State migration
Existing connections will lose their incremental cursor state and perform a full re-sync from `jira_start_date` on first run after upgrade. This is acceptable — current data has no production value.

## Architecture
```
sync start
  └─ jira_project_keys (parent stream, internal)
       GET /rest/api/3/project/search  →  [PROJ-A, PROJ-B, PROJ-C, ...]
            │
            ├─ jira_issue [PROJ-A] × [t₀..t₁] [t₁..t₂] ...
            ├─ jira_issue [PROJ-B] × [t₀..t₁] [t₁..t₂] ...
            └─ jira_issue [PROJ-C] × [t₀..t₁] ...
```

## Summary by CodeRabbit

* **Refactor**
  * Removed project-key allowlist — connector now auto-discovers projects and runs per-project syncs; upgrades may drop per-source incremental cursors and trigger full re-syncs from the configured start date.
  * Sync window end moved forward by a 14-hour buffer.

* **Documentation**
  * Docs, ADRs, README and examples updated to reflect automatic project discovery and revised configuration guidance.

* **Chores**
  * Connector version bumped to 2.0.0; example secret manifest updated (project key field removed).

---


mitasovr:
Superseded by #1316, which reworks this idea against the current manifest (crediting the original design from this PR / cyberfabric/cyber-insight#616).

Why a rework rather than a rebase — the connector changed significantly since this was written:

1. #1283 introduced `jira_issue_keys`, a lightweight substream parent whose JQL also references `jira_project_keys`. Removing the config key without migrating that stream (which this PR predates) would silently zero out `jira_issue_history`/`jira_comments`/`jira_worklogs` — the JQL renders as `project IN ()`, Jira returns 400, the error handler IGNOREs it, and all three substreams produce no partitions while the sync stays green.
2. #1308 (`default_concurrency: 4`) is required for the wider per-project partition fan-out — at concurrency 1 the concurrent CDK self-deadlocks once a sync generates ≥ ~10k partitions.
3. The `+14h end_datetime` timezone workaround advances the cursor past the actual query time, permanently skipping records updated between the query moment and the future-dated window end. #1316 addresses the same timezone tail with `lookback_window: PT14H` instead.
4. #1316 also adds an incremental discovery gate (`expand=insight` → client-side cursor on `insight.lastIssueUpdateTime`), so after the first full sync only projects with issue changes are scanned — verified live: 3/211 project partitions touched on a resume read.

Thanks for the groundwork — the inline discovery parent and per-project JQL shape carried over directly.

PR #1277, opened 2026-06-09
feat(observability): observability.mode contract + bundled-stack recipe

mitasovr:
## What

Adds a first-class **observability** switch to the umbrella, modelled on the existing infra `deploy`-flag pattern. The product governs only what services **emit** and where they point — it never bundles the collector itself (same model as Airbyte/Argo).

```yaml
observability:
  mode: none      # none | external | bundled
  logs: { level: info, format: json }
  traces: { enabled: false }
  otlp: { endpoint: "", protocol: grpc }
```

- `none` (default) — services only write structured JSON to stdout; the host cluster collects. **Safe drop into an unknown existing cluster.**
- `external` — services export OTLP to an operator-supplied endpoint (the customer's own stack).
- `bundled` — services export to an in-cluster Loki/Alloy/Grafana stack stood up out-of-band.

JSON-to-stdout is **always on** — that is the product contract; `mode` only decides where (if anywhere) Insight also exports.

## Changes

- `charts/insight/values.yaml` — `observability` block.
- `charts/insight/templates/_helpers.tpl` — `insight.observability.{mode,otlpEndpoint}` helpers + fail-fast in `insight.validate` (endpoint required for `external`/`bundled`, mirroring the `deploy=false → host required` contract).
- `charts/insight/templates/platform-config.yaml` — publishes `INSIGHT_LOG_*` / `OTEL_*` into the `{release}-platform` ConfigMap (standard OTel env names → any SDK picks them up via `envFrom`, no per-service wiring).
- `charts/insight/values.schema.json` — schema with enums.
- `deploy/observability/` — reference Loki + Alloy + Grafana values + README (the `bundled` recipe).
- `deploy/scripts/install-observability.sh` — reference installer (Loki → Alloy → Grafana), `${NAMESPACE}` substituted via `sed` like `install-argo.sh`.

## Validation

`helm template` across all modes:
- `none` → only `INSIGHT_LOG_*` rendered, no OTLP vars.
- `external` without endpoint → fails with a clear message.
- `bundled` + endpoint (+ traces) → full `OTEL_*` set rendered.
- invalid mode → rejected by `values.schema.json`.

`bash -n` + YAML parse + `${NAMESPACE}` substitution checks pass for the deploy recipe.

## Scope / follow-ups (not in this PR)

This is the **emission contract + bundled recipe**. The source-side instrumentation it unlocks is deliberately separate:
- reconcile `lib/log.sh` → stdout JSON + `reconcile.completed` events
- dbt `--log-format json` + a `dbt.completed` event
- `sync.completed` from the airbyte-sync workflow
- OTel SDK wiring in the .NET/Python services + Faro in the frontend (the tracing phase)

## Summary by CodeRabbit

* **New Features**
  * Added observability configuration support, enabling users to configure log levels and OTLP export endpoints.
  * Introduced optional observability stack deployment (Loki, Alloy, Grafana) for centralized log collection and visualization.

* **Documentation**
  * Updated system layer deployment documentation with observability configuration guidance and setup instructions.


deploy/scripts/install-observability.sh
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/env bash
+#
+# Install/upgrade the BUNDLED observability stack (LGTM, logs first):
```
cyberantonz: Why this script exists at all? We use gitops approach to install everything, isn't it?

charts/insight/templates/_helpers.tpl
```diff
@@ -156,6 +156,33 @@ App services are mandatory umbrella components — no deploy flag.
 {{- define "insight.identity.host"            -}}{{- printf "%s-identity"             .Release.Name -}}{{- end -}}
 {{- define "insight.frontend.host"            -}}{{- printf "%s-frontend"             .Release.Name -}}{{- end -}}
 
+{{/*
+==============================================================================
+ OBSERVABILITY (logs & traces emission)
+==============================================================================
+Mirrors the infra `deploy=false → host required` contract:
+  - mode=none      → no collector; services log JSON to stdout only.
```
cyberantonz: What differs in external and bundled mode in this umbrella chart? Why those modes exist at all?

cyberantonz (commented):
need answers

deploy/scripts/install-observability.sh
mitasovr: By analogy with Airbyte/Argo. This repo already ships `deploy/scripts/install-airbyte.sh` and `deploy/scripts/install-argo.sh` even though the gitops repo drives our cluster installs — gitops is still the install mechanism for *our* clusters.

This script is the same thing for observability: the product-side **reference recipe** for `mode: bundled`. It is what the local/dev path uses, and what external operators (who do not have access to our private gitops repo) read to learn how to stand up the bundled stack. The gitops `system/` layer consumes the very same `deploy/observability/*-values.yaml`. So nothing changes about "we install via gitops" — this is the canonical, consumer-visible recipe, exactly like Airbyte/Argo have one.

charts/insight/templates/_helpers.tpl
mitasovr: Same bundled-vs-external split we already use for the infra deps and for Airbyte (embedded vs external Postgres/S3):

- `bundled` — Insight stands up its own Loki/Alloy/Grafana; services export in-cluster.
- `external` — the host cluster already runs observability (their OTel Collector / Datadog / Splunk); services export OTLP to the operator-supplied endpoint and we install nothing.
- `none` — stdout only; the host cluster's node agent scrapes it.

Inside *this* umbrella chart `external` and `bundled` do render the same `OTEL_*` vars — the only difference is the endpoint published into `{release}-platform` (bundled → our in-cluster Alloy Service; external → the customer's endpoint). The umbrella never installs the collector in either mode, just like it never installs Airbyte. The actual install/skip divergence lives at the deploy layer (gitops inventory toggles / install-observability.sh).

The modes exist so that dropping Insight into an existing cluster does not duplicate the customer's logging stack — the same reason ClickHouse/MariaDB expose `deploy: true/false`.

mitasovr: Reworked — dropped the `mode` flag entirely (commit 8a248f73). It is now purely endpoint-driven: `observability.otlp.endpoint` empty → services log to stdout only (no `OTEL_*` published); set → they export OTLP there. Whether the bundled stack actually gets installed is the gitops `inventory.system.{loki,alloy,grafana}` toggle. So no redundant enum in the chart anymore — exactly your point.

deploy/scripts/install-observability.sh
mitasovr: Removed `install-observability.sh` (commit 8a248f73). Cluster install is `make`-driven in gitops (`system-loki`/`system-alloy`/`system-grafana`) and will be documented in the gitops-sample; local dev does not need it. The `deploy/observability/*-values.yaml` stay as reference only, same as `deploy/argo` / `deploy/airbyte`.

mitasovr: Update: went one step further in e2ea3d5b — `deploy/observability/` is removed entirely (my earlier "values stay as reference" is withdrawn). With the script gone it had no consumer and duplicated the deploy pipeline. Single source for collector install material is the gitops side, publicly mirrored in `docs/deploy/system/` in this repo.

charts/insight/templates/_helpers.tpl
mitasovr: Also trimmed the contract further in e2ea3d5b: dropped `logs.format` (JSON-to-stdout IS the contract, not a setting) and `traces.enabled` (`OTEL_TRACES_EXPORTER` hardcoded "none" until the tracing phase). What remains is exactly the two operational knobs: `logs.level` + `otlp.{endpoint,protocol}`.

cyberantonz (approved):

Issue #1260, opened 2026-06-08
EPIC: BFF cookie-auth migration


cyberantonz:
## Goal

Replace the current Bearer-token + sessionStorage SPA auth with the BFF (Backend-for-Frontend) pattern specified in [`docs/components/backend/api-gateway/`](docs/components/backend/api-gateway/). End state:

- SPA holds **no tokens**. Only the opaque `__Host-sid` cookie.
- BFF (api-gateway) owns OIDC code+PKCE, Redis sessions, refresh rotation, CSRF.
- Router (api-gateway) mints short-lived EdDSA JWTs for downstream services from the validated session.
- Frontend uses `credentials: 'include'`, drops `oidc-client-ts`.

## Phase 1 status

Phase 1 landed in commit [`1196a69`](../commit/1196a69) on `feat/cookie-auth`. Delivered:

- `GET /auth/login` — code+PKCE start, `bff:login_state:{state}` (5 min TTL).
- `GET /auth/callback` — atomic GET+DEL of login state, token exchange, ID-token validation (sig via JwksKeyProvider, then iss/aud/nonce/exp/nbf/azp + algo allow-list), cookie fixation guard, session creation in single MULTI/EXEC, `__Host-sid` set, 302 to sanitized `return_to`.
- `GET /auth/me` — cookie validation, both `expires_at` + `absolute_expires_at` enforced, returns `{user, tenant, expires_at, refresh_at, csrf_token}`; 401+clear-cookie on miss.
- Cookie hardening per `cpt-insightspec-nfr-bff-cookie-attrs` (snapshot-tested `Set-Cookie`).
- Redis schema `bff:{session,user_sessions,sid_index,login_state}` per DD-BFF-04.
- 87 unit tests + 1 ignored Redis integration round-trip.

## Implementation order

Per resolved open question Q5: **Backend phases 2–3 first, then frontend.** Router can proceed in parallel after BFF Phase 1.

### Backend — `constructorfabric/insight`

- [ ] BFF Phase 2 — `/auth/refresh` + `/auth/logout` (#1261)
- [ ] BFF — Wire Identity Service `sub → user_id` in `/auth/callback` + `/auth/me` (#1262)
- [ ] BFF Phase 3a — `/auth/sessions` (list, DELETE, DELETE all) + `/auth/oidc/back-channel-logout` (#1263)
- [ ] BFF Phase 3b — `/auth/csrf` + CSRF middleware (#1264)
[...]

PR #1279, opened 2026-06-09
BFF Phase 1: OIDC login + server-side sessions

cyberantonz:
## Summary

Phase 1 of the BFF cookie-auth migration (issue #1260 EPIC). Adds a new BFF (Backend-for-Frontend) ModKit module to the api-gateway binary plus a shared `redis-client` module that publishes a `RedisShared` handle on the ClientHub.

Specs: [`docs/components/backend/api-gateway/bff/PRD.md`](docs/components/backend/api-gateway/bff/PRD.md) §5.1, §5.2, §5.4 and [`bff/DESIGN.md`](docs/components/backend/api-gateway/bff/DESIGN.md) §3.6 (Login). FR IDs: `cpt-insightspec-fr-bff-oidc-login`, `cpt-insightspec-fr-bff-session-cookie`, `cpt-insightspec-fr-bff-session-store`.

### Endpoints (all public; BFF owns its own cookie-based auth)

- **GET /auth/login** — kicks off OIDC code+PKCE; stores state/nonce/PKCE verifier in `bff:login_state:{state}` (5 min TTL); 302 to the IdP authorize endpoint.
- **GET /auth/callback** — atomic GET+DEL of login state; exchanges code at the IdP token endpoint; validates the ID token (signature via `modkit_auth::JwksKeyProvider`, then our own iss/aud/nonce/exp/nbf/azp checks plus an algorithm allow-list); revokes any attacker-planted incoming cookie (fixation guard); creates a session in Redis under one atomic MULTI/EXEC; sets `__Host-sid` and 302s the browser to a sanitized `return_to`.
- **GET /auth/me** — reads cookie, looks up session, enforces both `expires_at` and `absolute_expires_at`; returns `{user, tenant, expires_at, refresh_at, csrf_token}`; 401 + clear-cookie otherwise.

### Cookie hardening (`cpt-insightspec-nfr-bff-cookie-attrs`)

Single helper emits `__Host-sid`, `HttpOnly`, `Secure`, `SameSite=Strict`, `Path=/`, no `Domain`. Snapshot tests pin the exact `Set-Cookie` string. Reader rejects duplicate `__Host-sid` cookies and strips RFC 6265 quoted values.

### Redis schema (DD-BFF-04, prefix `bff:`)

- `session:{sid}` HASH
- `user_sessions:{user_id}` ZSET (score = `expires_at`)
- `sid_index:{iss}:{idp_sid}` SET
- `login_state:{state}` HASH

Phase 2/3 keys (`swap`, `logout_jti`, `lock:janitor`) declared in the same builder file for schema unity but not used yet.

### Identity

`user_id = OIDC sub` verbatim (single OIDC issuer per installation by deployment design). Identity Service mapping is a separate scope; `idp_iss` and `idp_sub` are persisted on every session so the future `(scope="oidc", id=sub)` resolver call has its data.

### Out of scope (intentional, follow-ups)

`/auth/refresh`, `/auth/logout`, `/auth/sessions`, `/auth/csrf`, `/auth/oidc/back-channel-logout`, CSRF middleware, janitor, rate-limit middleware, dev-fake-OIDC stub, Helm + config example.

### Side fix-up (second commit)

Two trivial clippy-1.95 lints (`map(f).unwrap_or(x)` → `map_or(x, f)`, `Duration::from_secs(3600)` → `Duration::from_hours(1)`) plus the Phase 1 integration test's `now: 1` → fixed future epoch so EXPIREAT lands ahead of Redis's wall clock instead of evicting immediately. No behavioural change.

## Test plan

- [x] `cargo fmt -p insight-api-gateway` clean
- [x] `cargo clippy --workspace --all-targets` clean (workspace-pedantic config, `pedantic = deny`, `unwrap_used = deny`, `expect_used = deny`)
- [x] `cargo test -p insight-api-gateway` — 87 / 87 pass
- [x] `BFF_TEST_REDIS_URL=redis://localhost:16379/15 cargo test -p insight-api-gateway -- --include-ignored` — 88 / 88 pass against Redis 7
- [ ] Reviewer: cookie attributes verified via snapshot tests in `cookies.rs`
- [ ] Reviewer: grep diff for `id_token` and confirm it never reaches logs or response bodies (only Redis HASH field + future Phase 2 `id_token_hint`)

Phase 2 (`/auth/refresh` + `/auth/logout`) follows as a stacked PR (#1278) once this lands.

## Summary by CodeRabbit

* **New Features**
  * Backend-for-Frontend (BFF) auth: /auth/login, /auth/callback, /auth/me with OIDC (PKCE), discovery, token exchange and session refresh.
  * Server-side Redis sessions, secure __Host-sid cookie handling, CSRF tokens, login-state lifecycle, jittered refresh scheduling, and audit-event emission.
  * Centralized canonical API error responses, Redis key namespace, and a Redis client module.

* **Tests**
  * Extensive unit tests for config, crypto/secrets, cookies, session store, OIDC flows, audit, routing, and helpers.

### DIFF
--- a/src/backend/services/api-gateway/src/bff/config.rs
+++ b/src/backend/services/api-gateway/src/bff/config.rs
@@ -1,0 +1,249 @@
+//! BFF module configuration.
+//!
+//! All knobs live under `modules.bff.config` in the api-gateway YAML.
+//! Defaults match `cpt-insightspec-fr-bff-*` requirements unless explicitly
+//! noted; production values should be set via Helm.
+
+use serde::Deserialize;
+
+/// BFF module configuration.
+#[derive(Debug, Clone, Default, Deserialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct BffConfig {
+    /// OIDC settings.
+    pub oidc: OidcConfig,
+    /// Session lifecycle settings.
+    pub session: SessionConfig,
+    /// CSRF settings.
+    pub csrf: CsrfConfig,
+    /// Public origin where the BFF serves the SPA's redirect target.
+    /// Used to build absolute callback URLs and the post-login redirect.
+    /// Required.
+    pub public_origin: String,
+    /// Default tenant assigned to every authenticated user until the
+    /// Identity Service mapping lands. Single-tenant deployments only.
+    pub default_tenant_id: String,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct OidcConfig {
+    /// OIDC issuer URL — `<issuer>/.well-known/openid-configuration` is
+    /// fetched at module init.
+    pub issuer_url: String,
+    /// Confidential client ID.
+    pub client_id: String,
+    /// Confidential client secret.
+    pub client_secret: String,
+    /// Scopes to request (`openid` is appended automatically if missing).
+    pub scopes: Vec<String>,
+    /// Optional override for the OIDC `audience` claim validation. Defaults
+    /// to `client_id`.
+    pub audience: Option<String>,
+}
+
+impl Default for OidcConfig {
+    fn default() -> Self {
+        Self {
+            issuer_url: String::new(),
+            client_id: String::new(),
+            client_secret: String::new(),
+            scopes: vec![
+                "openid".to_owned(),
+                "profile".to_owned(),
+                "email".to_owned(),
+            ],
+            audience: None,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct SessionConfig {
+    /// Session cookie TTL. Spec range: 30 s – 1 h. Default 120 s.
+    pub ttl_seconds: u64,
+    /// Hard cap on session lifetime across refreshes. Spec range:
+    /// 1 h – 24 h. Default 8 h.
+    pub absolute_lifetime_seconds: u64,
+    /// `refresh_at = expires_at - safety_margin + jitter`. Default 30 s.
+    pub refresh_safety_margin_seconds: u64,
+    /// Total jitter window applied to `refresh_at` (uniform in `±half/2`).
+    /// Default 10 s.
+    pub refresh_jitter_seconds: u64,
+    /// Grace TTL for `bff:swap:{old_sid}` after a refresh-rotation. Default
+    /// 250 ms.
+    pub refresh_grace_ms: u64,
+}
+
+impl Default for SessionConfig {
+    fn default() -> Self {
+        Self {
+            ttl_seconds: 120,
+            absolute_lifetime_seconds: 8 * 3600,
+            refresh_safety_margin_seconds: 30,
+            refresh_jitter_seconds: 10,
+            refresh_grace_ms: 250,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Default, Deserialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct CsrfConfig {
+    /// Allowlist of `Origin` values accepted as the secondary CSRF defense.
+    /// Empty means fail-closed: every state-changing `/auth/*` request must
+    /// carry a valid `X-CSRF-Token`.
+    pub origins: Vec<String>,
+}
+
+impl BffConfig {
+    /// Validate the loaded config against spec ranges. Called once at
+    /// module init.
+    pub fn validate(&self) -> anyhow::Result<()> {
+        if self.public_origin.is_empty() {
+            anyhow::bail!("bff: public_origin is required");
+        }
+        url::Url::parse(&self.public_origin)
+            .map_err(|e| anyhow::anyhow!("bff: public_origin is not a valid URL: {e}"))?;
+
+        if self.oidc.issuer_url.is_empty() {
+            anyhow::bail!("bff: oidc.issuer_url is required");
+        }
+        if self.oidc.client_id.is_empty() {
+            anyhow::bail!("bff: oidc.client_id is required");
+        }
+        if self.oidc.client_secret.is_empty() {
+            anyhow::bail!("bff: oidc.client_secret is required");
+        }
+
+        // cpt-insightspec-nfr-bff-session-ttl
+        if !(30..=3600).contains(&self.session.ttl_seconds) {
+            anyhow::bail!(
+                "bff: session.ttl_seconds must be in [30, 3600], got {}",
+                self.session.ttl_seconds
+            );
+        }
+        if !(3600..=86_400).contains(&self.session.absolute_lifetime_seconds) {
+            anyhow::bail!(
+                "bff: session.absolute_lifetime_seconds must be in [3600, 86400], got {}",
+                self.session.absolute_lifetime_seconds
+            );
+        }
+        if self.session.refresh_safety_margin_seconds >= self.session.ttl_seconds {
+            anyhow::bail!(
+                "bff: session.refresh_safety_margin_seconds ({}) must be < session.ttl_seconds ({})",
+                self.session.refresh_safety_margin_seconds,
+                self.session.ttl_seconds
+            );
+        }
+        if self.session.refresh_jitter_seconds >= self.session.ttl_seconds {
+            anyhow::bail!(
+                "bff: session.refresh_jitter_seconds ({}) must be < session.ttl_seconds ({})",
+                self.session.refresh_jitter_seconds,
+                self.session.ttl_seconds
+            );
+        }
+        if self.session.refresh_grace_ms == 0 || self.session.refresh_grace_ms > 5_000 {
+            anyhow::bail!(
+                "bff: session.refresh_grace_ms must be in (0, 5000], got {}",
+                self.session.refresh_grace_ms
+            );
+        }
+
+        Ok(())
+    }
+
+    /// Final list of OIDC scopes with `openid` guaranteed present.
+    #[must_use]
+    pub fn effective_scopes(&self) -> Vec<String> {
+        let mut scopes = self.oidc.scopes.clone();
+        if !scopes.iter().any(|s| s == "openid") {
+            scopes.insert(0, "openid".to_owned());
+        }
+        scopes
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn good_cfg() -> BffConfig {
+        BffConfig {
+            public_origin: "https://insight.example.com".to_owned(),
+            default_tenant_id: "00000000-0000-0000-0000-000000000001".to_owned(),
+            oidc: OidcConfig {
+                issuer_url: "https://idp.example.com".to_owned(),
+                client_id: "insight-bff".to_owned(),
+                client_secret: "secret".to_owned(),
+                ..OidcConfig::default()
+            },
+            ..BffConfig::default()
+        }
+    }
+
+    #[test]
+    fn validate_accepts_good_config() {
+        good_cfg().validate().expect("should pass");
+    }
+
+    #[test]
+    fn validate_rejects_short_session_ttl() {
+        let mut c = good_cfg();
+        c.session.ttl_seconds = 10;
+        assert!(c.validate().is_err());
+    }
+
+    #[test]
+    fn validate_rejects_long_session_ttl() {
+        let mut c = good_cfg();
+        c.session.ttl_seconds = 7200;
+        assert!(c.validate().is_err());
+    }
+
+    #[test]
+    fn validate_rejects_short_absolute_lifetime() {
+        let mut c = good_cfg();
+        c.session.absolute_lifetime_seconds = 300;
+        assert!(c.validate().is_err());
+    }
+
+    #[test]
+    fn validate_rejects_safety_margin_geq_ttl() {
+        let mut c = good_cfg();
+        c.session.refresh_safety_margin_seconds = 120;
+        assert!(c.validate().is_err());
+    }
+
+    #[test]
+    fn effective_scopes_inserts_openid_when_missing() {
+        let mut c = good_cfg();
+        c.oidc.scopes = vec!["profile".to_owned()];
+        assert_eq!(
+            c.effective_scopes(),
+            vec!["openid".to_owned(), "profile".to_owned()]
+        );
+    }
+
+    #[test]
+    fn effective_scopes_keeps_openid_when_present() {
+        let c = good_cfg();
+        let s = c.effective_scopes();
+        assert_eq!(s.iter().filter(|x| x.as_str() == "openid").count(), 1);
+    }
+
+    #[test]
+    fn validate_rejects_empty_public_origin() {
+        let mut c = good_cfg();
+        c.public_origin = String::new();
+        assert!(c.validate().is_err());
+    }
+
+    #[test]
+    fn validate_rejects_empty_issuer() {
+        let mut c = good_cfg();
+        c.oidc.issuer_url = String::new();
+        assert!(c.validate().is_err());
+    }
+}
--- a/src/backend/services/api-gateway/src/bff/config.rs
+++ b/src/backend/services/api-gateway/src/bff/config.rs
@@ -62,18 +62,18 @@ impl Default for OidcConfig {
 #[serde(default, deny_unknown_fields)]
 pub struct SessionConfig {
     /// Session cookie TTL. Spec range: 30 s – 1 h. Default 120 s.
-    pub ttl_seconds: u64,
+    pub ttl_seconds: u32,
     /// Hard cap on session lifetime across refreshes. Spec range:
     /// 1 h – 24 h. Default 8 h.
-    pub absolute_lifetime_seconds: u64,
+    pub absolute_lifetime_seconds: u32,
     /// `refresh_at = expires_at - safety_margin + jitter`. Default 30 s.
-    pub refresh_safety_margin_seconds: u64,
+    pub refresh_safety_margin_seconds: u16,
     /// Total jitter window applied to `refresh_at` (uniform in `±half/2`).
     /// Default 10 s.
-    pub refresh_jitter_seconds: u64,
+    pub refresh_jitter_seconds: u16,
     /// Grace TTL for `bff:swap:{old_sid}` after a refresh-rotation. Default
     /// 250 ms.
-    pub refresh_grace_ms: u64,
+    pub refresh_grace_ms: u32,
 }
 
 impl Default for SessionConfig {
@@ -104,8 +104,9 @@ impl BffConfig {
         if self.public_origin.is_empty() {
             anyhow::bail!("bff: public_origin is required");
         }
-        url::Url::parse(&self.public_origin)
+        let parsed = url::Url::parse(&self.public_origin)
             .map_err(|e| anyhow::anyhow!("bff: public_origin is not a valid URL: {e}"))?;
+        validate_public_origin(&parsed)?;
 
         if self.oidc.issuer_url.is_empty() {
             anyhow::bail!("bff: oidc.issuer_url is required");
@@ -115,6 +116,9 @@ impl BffConfig {
         }
         if self.oidc.client_secret.is_empty() {
             anyhow::bail!("bff: oidc.client_secret is required");
+        }
+        if self.default_tenant_id.trim().is_empty() {
+            anyhow::bail!("bff: default_tenant_id is required");
         }
 
         // cpt-insightspec-nfr-bff-session-ttl
@@ -130,14 +134,14 @@ impl BffConfig {
                 self.session.absolute_lifetime_seconds
             );
         }
-        if self.session.refresh_safety_margin_seconds >= self.session.ttl_seconds {
+        if u32::from(self.session.refresh_safety_margin_seconds) >= self.session.ttl_seconds {
             anyhow::bail!(
                 "bff: session.refresh_safety_margin_seconds ({}) must be < session.ttl_seconds ({})",
                 self.session.refresh_safety_margin_seconds,
                 self.session.ttl_seconds
             );
         }
-        if self.session.refresh_jitter_seconds >= self.session.ttl_seconds {
+        if u32::from(self.session.refresh_jitter_seconds) >= self.session.ttl_seconds {
             anyhow::bail!(
                 "bff: session.refresh_jitter_seconds ({}) must be < session.ttl_seconds ({})",
                 self.session.refresh_jitter_seconds,
@@ -163,6 +167,37 @@ impl BffConfig {
         }
         scopes
     }
+}
+
+/// Defense-in-depth validation for `public_origin`: HTTPS only, except for
+/// local-dev hosts (`localhost`, `127.0.0.1`). Reject query/fragment/userinfo
+/// and any path beyond `/` since the value is concatenated with `"/auth/..."`
+/// at the call site. Runtime traffic is still enforced HTTPS-only at the
+/// ingress (NFR `cpt-insightspec-nfr-bff-https-only`); this guards the config.
+fn validate_public_origin(u: &url::Url) -> anyhow::Result<()> {
+    let host = u.host_str().unwrap_or("");
+    let is_local = host == "localhost" || host == "127.0.0.1";
+    match u.scheme() {
+        "https" => {}
+        "http" if is_local => {}
+        s => anyhow::bail!(
+            "bff: public_origin must use https (got `{s}://{host}`); http allowed only for localhost"
+        ),
+    }
+    if !u.username().is_empty() || u.password().is_some() {
+        anyhow::bail!("bff: public_origin must not include userinfo");
+    }
+    if u.query().is_some() {
+        anyhow::bail!("bff: public_origin must not include a query string");
+    }
+    if u.fragment().is_some() {
+        anyhow::bail!("bff: public_origin must not include a fragment");
+    }
+    let path = u.path();
+    if !(path.is_empty() || path == "/") {
+        anyhow::bail!("bff: public_origin must not include a path (got `{path}`)");
+    }
+    Ok(())
 }
 
 #[cfg(test)]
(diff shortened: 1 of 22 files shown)


src/backend/services/api-gateway/src/bff/handlers/callback.rs
```diff
@@ -0,0 +1,261 @@
...
+        }
+    }
+    headers
+        .get("x-real-ip")
+        .and_then(|v| v.to_str().ok())
+        .unwrap_or("")
+        .to_owned()
+}
+
+fn unix_now() -> i64 {
```
cyberantonz: Strange code. You have similar example in rust book.
```rust
use std::time::SystemTime;

match SystemTime::now().duration_since(SystemTime::UNIX_EPOCH) {
    Ok(n) => println!("1970-01-01 00:00:00 UTC was {} seconds ago!", n.as_secs()),
    Err(_) => panic!("SystemTime before UNIX EPOCH!"),
}
```

So, you can return `n.as_secs()` in Ok(n) and panic at error. I do not expect anything before system time.

src/backend/services/api-gateway/src/bff/handlers/callback.rs
```diff
@@ -0,0 +1,261 @@
...
+    if cleaned.is_empty() {
+        "unknown".to_owned()
+    } else {
+        cleaned
+    }
+}
+
+/// Best-effort source IP. Honors `X-Forwarded-For` (first hop) when the
+/// gateway is behind a trusted proxy — the ingress in our deployment.
+fn client_ip_from_headers(headers: &HeaderMap) -> String {
```
cyberantonz, line 187: Why do you not use axum-client-ip?

src/backend/services/api-gateway/src/bff/handlers/login.rs
```diff
@@ -0,0 +1,178 @@
...
+///   * any control character (CR/LF/TAB/etc) — defense in depth
+///   * anything above the length cap
+///
+/// On rejection or absence we fall back to `/`.
+pub(super) fn sanitize_return_to(raw: Option<&str>) -> String {
+    let raw = raw.unwrap_or("/").trim();
+    if raw.is_empty() || !raw.starts_with('/') {
+        return "/".to_owned();
+    }
+    if raw.len() > 1024 {
```
cyberantonz: You need to introduce module level (or config with default parameter) constant. Avoid using such "magic" values.

src/backend/services/api-gateway/src/bff/handlers/me.rs
```diff
@@ -0,0 +1,124 @@
...
+    let st = state;
+
+    let sid = read_session_cookie(headers.get(header::COOKIE))
+        .ok_or(BffError::Unauthorized("no session cookie"))?;
+
+    let Some(record) = st.store.get_session(&sid).await? else {
+        return Ok(unauthorized_clear_cookie());
+    };
+
+    let now = i64::try_from(
```
cyberantonz: Same issue with time, you may create separate function for it (as i see in callback.rs) and use it everywhere, do not repeat yourself.

src/backend/services/api-gateway/src/bff/handlers/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+    safety_margin_seconds: u64,
+    jitter_window_seconds: u64,
+) -> i64 {
+    use rand::Rng;
+
+    let base = expires_at.saturating_sub(i64::try_from(safety_margin_seconds).unwrap_or(i64::MAX));
+    if jitter_window_seconds == 0 {
+        return base;
+    }
+    let half = i64::try_from(jitter_window_seconds / 2).unwrap_or(i64::MAX);
```
cyberantonz: It is better to use `jitter_window_seconds: u8`, who in real life can use values more than 10 minutes window for jitter at all? Then you need not to make such try_from...

src/backend/services/api-gateway/src/bff/handlers/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+/// rest of the state.
+#[must_use]
+pub fn jittered_refresh_at(
+    expires_at: i64,
+    safety_margin_seconds: u64,
+    jitter_window_seconds: u64,
+) -> i64 {
+    use rand::Rng;
+
+    let base = expires_at.saturating_sub(i64::try_from(safety_margin_seconds).unwrap_or(i64::MAX));
```
cyberantonz: safety_margin_seconds can be u16. 8 hours more than enough.  Or even u8.

src/backend/services/api-gateway/src/bff/secrets.rs
```diff
@@ -0,0 +1,104 @@
+//! CSPRNG helpers for opaque session IDs and CSRF tokens.
```
cyberantonz, line 1: Not invented here? You are learned on MS code??

```
use csrf::{AesGcmCsrfProtection, CsrfProtection};
```

src/backend/services/api-gateway/src/bff/cookies.rs
```diff
@@ -0,0 +1,260 @@
...
+/// header. Callers never see `Max-Age` arithmetic or `HeaderName`s.
+///
+/// ```ignore
+/// (StatusCode::OK, no_store(), SessionCookie::set(&sid, expires_at, now), Json(view))
+///     .into_response()
+/// ```
+#[must_use]
+pub struct SessionCookie<'a> {
+    value: &'a str,
+    max_age_seconds: i64,
```
cyberantonz: Do you really need i64 here????

src/backend/services/api-gateway/src/bff/oidc_client.rs
```diff
@@ -0,0 +1,764 @@
+//! Minimal OIDC confidential client.
```
cyberantonz: Is there any good reason to invent own OIDC client ? There is stable library for it.

```
use openidconnect::
```

src/backend/services/api-gateway/src/bff/redis_keys.rs
```diff
@@ -0,0 +1,85 @@
+//! Centralized Redis key builders for the BFF.
```
cyberantonz, line 1: Please validate key names with DESIGN.md document.

src/backend/services/api-gateway/src/bff/session_store.rs
```diff
@@ -0,0 +1,484 @@
...
+    /// Empty string means "IdP did not supply a sid claim". Empty values
+    /// skip the sid_index write.
+    pub idp_sid: &'a str,
+    pub id_token: &'a str,
+    pub email: &'a str,
+    pub display_name: &'a str,
+    pub user_agent: &'a str,
+    pub ip: &'a str,
+    pub now: i64,
+    pub session_ttl_seconds: u64,
```
cyberantonz: How long the session may be alive? Use some more real world value. u32 more than enough

src/backend/services/api-gateway/src/bff/errors.rs
```diff
@@ -0,0 +1,143 @@
...
+                "Bad Gateway",
+                "urn:insight:error:idp_unreachable",
+                detail.clone(),
+            ),
+            Self::StoreUnavailable(detail) => (
+                StatusCode::SERVICE_UNAVAILABLE,
+                "Service Unavailable",
+                "urn:insight:error:session_store_unavailable",
+                detail.clone(),
+            ),
```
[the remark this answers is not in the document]
cyberantonz: Already addressed by commit `01b31fd` ("emit BFF errors via the constructorfabric canonical envelope"). Both `BffError::Idp` and `BffError::StoreUnavailable` now collapse to `CanonicalError::service_unavailable()`, whose default detail is `"Service temporarily unavailable"` — the diagnostic stays in the server-side `tracing::warn!` only. New unit tests `bff::errors::{idp_maps_to_service_unavailable, store_unavailable_maps_to_service_unavailable}` pin that the diagnostic never leaks into the response body.

src/backend/services/api-gateway/src/bff/handlers/callback.rs
```diff
@@ -0,0 +1,267 @@
...
+            idp_sid: exch.claims.sid.as_deref().unwrap_or(""),
+            id_token: &exch.id_token,
+            email: &resolved.email,
+            display_name: &resolved.display_name,
+            user_agent: &user_agent,
+            ip: &ip,
+            now,
+            session_ttl_seconds: st.cfg.session.ttl_seconds,
+            absolute_lifetime_seconds: st.cfg.session.absolute_lifetime_seconds,
+            incoming_sid: incoming_sid.as_deref(),
```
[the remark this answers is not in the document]
cyberantonz: Skipping in Phase 1 — the premise that "/auth/me cannot clear a bad cookie" doesn't hold. `unauthorized_clear_cookie()` emits `SessionCookie::clear()` on every unauth path in `/auth/me`, including the `read_session_cookie → None` branch (i.e., the cookie was missing OR malformed — both look the same from the parser, so any bad value is cleared on the next call).

For the callback fixation guard, the only way an "invalid" `__Host-sid` reaches the BFF requires same-origin XSS or a misbehaving proxy; given `__Host-` + `HttpOnly` + `SameSite=Strict` + `Secure`, no standard fixation vector goes through a malformed cookie value, so skipping revocation in that edge case is acceptable.

The tri-state refactor would ripple through `CreateSessionRequest`, every caller, and tests for a residual risk the cookie attributes already mitigate. Happy to add a future-additive metric/log when the parser rejects a present `Cookie` header if telemetry is wanted — no API change.

src/backend/services/api-gateway/src/bff/secrets.rs
cyberantonz: Thanks for the nudge, but this is intentional rather than NIH. DESIGN §4.2 and PRD §5.8 pin the CSRF defense to (1) `SameSite=Strict` primary, (2) double-submit `X-CSRF-Token` compared constant-time against `session.csrf_token` on the Redis session record, (3) `Origin` allowlist fallback. That's a session-store-backed model, so the token just needs to be unguessable random — `OsRng` + URL-safe base64 is exactly that.

The `csrf` crate's `AesGcmCsrfProtection` is stateless self-contained ciphertext (binding + timestamp + counter encrypted under a server key); it solves a different problem (validation without a store) and wouldn't slot into our server-side session model without redundancy.

Also, the CSRF middleware itself lands in Phase 3b per the EPIC — this PR only ships the CSPRNG building blocks (`new_session_id`, `new_csrf_token`, `new_state`, `new_nonce`, `ct_eq`). Happy to revisit if the design pivots.

src/backend/services/api-gateway/src/bff/config.rs
```diff
@@ -0,0 +1,249 @@
...
+        if self.oidc.issuer_url.is_empty() {
+            anyhow::bail!("bff: oidc.issuer_url is required");
+        }
+        if self.oidc.client_id.is_empty() {
+            anyhow::bail!("bff: oidc.client_id is required");
+        }
+        if self.oidc.client_secret.is_empty() {
+            anyhow::bail!("bff: oidc.client_secret is required");
+        }
+
```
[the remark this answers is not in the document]
cyberantonz: Addressed in `7e58560` — `validate()` now bails when `default_tenant_id.trim().is_empty()`, with `validate_rejects_empty_default_tenant_id` and `..._blank_...` tests pinning the new behavior.

src/backend/services/api-gateway/src/bff/config.rs
```diff
@@ -0,0 +1,249 @@
...
+impl BffConfig {
+    /// Validate the loaded config against spec ranges. Called once at
+    /// module init.
+    pub fn validate(&self) -> anyhow::Result<()> {
+        if self.public_origin.is_empty() {
+            anyhow::bail!("bff: public_origin is required");
+        }
+        url::Url::parse(&self.public_origin)
+            .map_err(|e| anyhow::anyhow!("bff: public_origin is not a valid URL: {e}"))?;
+
```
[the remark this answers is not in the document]
cyberantonz: Addressed in `7e58560` — `public_origin` is now tightened to https-only with a localhost/127.0.0.1 carve-out for local dev. Rejects userinfo, query strings, fragments, and any non-root path. Runtime HTTPS-only (`cpt-insightspec-nfr-bff-https-only`) is still enforced at the ingress; this is the config invariant. Tests cover both pass and fail paths.

src/backend/services/api-gateway/src/bff/handlers/mod.rs
cyberantonz: Addressed in `7e58560` — `jittered_refresh_at(... safety_margin_seconds, jitter_window_seconds)` and the matching `SessionConfig` fields are now `u16`, so the `try_from(...).unwrap_or(i64::MAX)` boilerplate is gone (replaced with `i64::from(...)`). The now-unrepresentable overflow-safety-margin test was dropped.

src/backend/services/api-gateway/src/bff/handlers/mod.rs
cyberantonz: Addressed in `7e58560` — `safety_margin_seconds` is now `u16` (paired with jitter narrowing). u8 felt tight if anyone wants a wider jitter window, but for both safety_margin and jitter, u16 gives ~18× headroom over the 3600s TTL cap with no `try_from` boilerplate.

src/backend/services/api-gateway/src/bff/session_store.rs
cyberantonz: Addressed in `7e58560` — `CreateSessionRequest.{session_ttl_seconds, absolute_lifetime_seconds}` and the matching `SessionConfig` fields are now `u32` (~136 years ceiling, comfortably above any real session TTL). `u32 → i64` is infallible, so the two `i64::try_from(...).unwrap_or(i64::MAX)` callsites in `session_store.rs` are now plain `i64::from(...)`.

src/backend/services/api-gateway/src/redis_client.rs
```diff
@@ -0,0 +1,154 @@
...
+use tracing::info;
+
+/// Module configuration.
+#[derive(Debug, Clone, Default, Deserialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct RedisClientConfig {
+    /// Redis connection URL (e.g. `redis://localhost:6379/0`). Empty means
+    /// "no Redis configured" — modules that depend on this client will fail
+    /// at their own init when they call `client_hub.get::<RedisShared>()`.
+    pub url: String,
```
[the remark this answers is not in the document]
cyberantonz: Addressed in `b154e03` — `RedisClientConfig.url` doc now reads: "Redis connection URL (e.g. `redis://localhost:6379/0`). Required: module initialization fails fast if this is empty." — aligned with what `init()` actually does.

src/backend/services/api-gateway/src/bff/redis_keys.rs
cyberantonz: Addressed in `b154e03` — added `bff_rl_auth(ip)` and `bff_rl_login_state_count()` builders for the two DESIGN §4.4 rate-limit keys the module was missing, and replaced the raw `"bff:rl:login_state_count"` literal in `session_store::login_state::touch` with the new builder. Schema is back to one place. `keys_use_bff_prefix` extended to cover both new builders.

src/backend/services/api-gateway/src/bff/oidc_client.rs
```diff
@@ -0,0 +1,764 @@
...
+        if doc.issuer.is_empty()
+            || doc.authorization_endpoint.is_empty()
+            || doc.token_endpoint.is_empty()
+            || doc.jwks_uri.is_empty()
+        {
+            return Err(BffError::Idp(
+                "discovery doc missing required fields".into(),
+            ));
+        }
+        Ok(doc)
```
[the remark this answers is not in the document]
cyberantonz: Addressed in `6cbb23b` — `DiscoveryDoc::fetch` now compares the advertised `issuer` against the configured `issuer_url` (trailing slash normalized on both sides) per OIDC Core §3.1.2.2 and bails with `BffError::Idp` on mismatch. Extracted to a pure `check_issuer_match` so the test pair (`check_issuer_match_accepts_equal_normalized`, `check_issuer_match_rejects_foreign_issuer`) doesn't need a mock HTTP server.

src/backend/services/api-gateway/src/bff/oidc_client.rs
```diff
@@ -0,0 +1,764 @@
...
+                .and_then(serde_json::Value::as_i64)
+                .ok_or_else(|| BffError::Idp(format!("id_token claim `{k}` missing or not int")))
+        };
+        let aud = match v.get("aud") {
+            Some(serde_json::Value::String(s)) => vec![s.clone()],
+            Some(serde_json::Value::Array(arr)) => arr
+                .iter()
+                .filter_map(|x| x.as_str().map(str::to_owned))
+                .collect(),
+            _ => return Err(BffError::Idp("id_token `aud` claim missing".into())),
```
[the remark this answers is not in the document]
cyberantonz: Addressed in `6cbb23b` — the `aud` array branch now uses `map(...).collect::<Result<Vec<_>, _>>()?` and rejects non-string entries (no more `filter_map` silently dropping numbers/objects). Empty arrays now collapse to the same `aud missing` error. Two new tests pin both rejections.

src/backend/services/api-gateway/src/bff/handlers/callback.rs
```diff
@@ -0,0 +1,267 @@
...
+        let safe = sanitize_oauth_error_code(err);
+        crate::bff::audit::emit(
+            AuthEventKind::LoginFail,
+            &AuthEvent {
+                reason: Some(&safe),
+                ..Default::default()
+            },
+        );
+        return Err(BffError::Idp(format!("idp error: {safe}")));
+    }
```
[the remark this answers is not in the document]
cyberantonz: Addressed in `29f4a01` — the error branch at `/auth/callback` now best-effort burns `bff:login_state:{state}` before returning. `login_state::take` is atomic GET+DEL so it's safe to call unconditionally; failure is logged and does not block the user-facing `BffError::Idp` response. Honors the one-shot spec contract even when the IdP rejected the request.

src/backend/services/api-gateway/src/bff/handlers/callback.rs
cyberantonz: Addressed in `29f4a01` — extracted to `pub(crate) fn bff::handlers::unix_now() -> i64` with bare `panic!` on pre-epoch / past-`i64::MAX` clocks (workspace-wide `unwrap_used`/`expect_used` denies satisfied without an `#[allow]`). Loud failure is correct for an operator-grade invariant violation. Replaces the local `unix_now` and drops the dead `SystemTime`/`UNIX_EPOCH` import from `callback.rs`.

src/backend/services/api-gateway/src/bff/handlers/me.rs
cyberantonz: Addressed in `29f4a01` — `/auth/me` now uses the shared `bff::handlers::unix_now` helper that was extracted alongside `callback.rs` (commit also folds in the `oidc_client.rs` copy). Drops `SystemTime`/`UNIX_EPOCH` from `me.rs` and replaces the silent `.map_or(0, ...)` with the panic-on-pre-epoch invariant.

src/backend/services/api-gateway/src/bff/handlers/login.rs
cyberantonz: Addressed in `80dd2b0` — lifted to a module-level `const RETURN_TO_MAX_LEN: usize = 1024;` with a doc comment noting it aligns with typical browser URL length caps. Used at both the check site in `sanitize_return_to` and in the `sanitize_return_to_caps_length` test, so the literal is in one place. Not promoted to a config knob — this is a defense-in-depth cap, not operator-tunable policy.

src/backend/services/api-gateway/src/bff/handlers/login.rs
```diff
@@ -0,0 +1,183 @@
...
+    // mid-path can still mislead URL parsers; if a real SPA path needs
+    // a backslash, it can be percent-encoded *and* we reject `%5c` below.
+    if raw.contains('\\') || raw.chars().any(|c| (c as u32) < 0x20) {
+        return "/".to_owned();
+    }
+    // Reject percent-encoded slashes/backslashes that some browsers
+    // decode before resolving Location. Case-insensitive match.
+    let lower = raw.to_ascii_lowercase();
+    if lower.contains("%2f") || lower.contains("%5c") {
+        return "/".to_owned();
```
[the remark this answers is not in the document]
cyberantonz: Addressed in `80dd2b0` — the check is now `raw.chars().any(char::is_control)`, which covers Unicode category Cc end-to-end: C0 (U+0000 to U+001F), DEL (U+007F), and C1 (U+0080 to U+009F). New `sanitize_return_to_rejects_del_and_c1_controls` test pins DEL, NEL, and APC. The doc comment was already promising "any control character", so the sanitizer behavior now matches it.

src/backend/services/api-gateway/src/bff/cookies.rs
cyberantonz: Addressed in `bb0f118` — narrowed the stored `SessionCookie.max_age_seconds` to `u32` (always clamped to >= 0; bounded above by the absolute 24h spec cap). The public `set(sid, expires_at: i64, now: i64)` stays as-is — epoch subtraction is naturally signed and clamping the negative case is what the helper exists for. The `u32` conversion happens at the boundary (`try_from(...).unwrap_or(u32::MAX)`); inside `build_cookie` the `time::Duration::seconds` call goes through `i64::from(u32)` (infallible). Snapshot tests unaffected.

src/backend/services/api-gateway/src/bff/handlers/callback.rs
cyberantonz: Skipping in this PR — pre-flight findings made this no longer a drop-in:

1. `modkit::bootstrap::run_server` does not expose a root-Router layer hook, so the only place we could install `ClientIpSource::*.into_extension()` is on the BFF sub-router via `register_rest`. Workable, but means `ClientIp` would 500 outside `/auth/*` and we would need to remember to add the same layer to any future module that wants it.

2. `axum-client-ip` v1 deliberately removed leftmost-XFF extraction (see its V1 breaking-changes section: `InsecureClientIp` and "leftmost" IP logic were dropped as a fix for [issue #32](https://github.com/imbolc/axum-client-ip/issues/32)). v1 only offers `RightmostXForwardedFor`, `XRealIp`, or per-CDN headers like `CfConnectingIp`. Our current hand-rolled `client_ip_from_headers` takes the leftmost XFF entry — that matches our k8s nginx-ingress topology (ingress appends the real client IP to XFF). `RightmostXForwardedFor` would pick the ingress proxy itself, not the user; the typed-extractor benefit is offset by getting the wrong IP.

3. The "right" v1-friendly knob for our deployment is probably `XRealIp` (nginx-ingress sets `X-Real-IP` to the connecting client), but that is a topology/operator decision rather than a code-cleanup. Switching makes assumptions about every deployment, including local-dev and any future non-nginx ingress.

Net: this slipped from "quick win" to "needs a deployment-topology decision," so it should land as a follow-up after the operator decides which trusted header to canonicalize on — not bundled into the Phase-1 BFF PR. Filed a TODO mentally; happy to open a tracking issue if useful.

src/backend/services/api-gateway/src/bff/oidc_client.rs
cyberantonz: Addressed in `8e4a3f6` — migrated to `openidconnect = 4`, dropping ~550 lines of hand-rolled discovery + token-exchange + ID-token-validation code. The crate now owns:

- Discovery + issuer match (OIDC Core §3.1.2.2 — closes the carryover of comment #6).
- ID-token signature verification + alg allow-list (`set_allowed_algs(RS{256,384,512}, ES{256,384}, EdDSA)`); HMAC / none stay excluded.
- `iss` / `aud` / `nonce` / `exp` / `nbf` claim checks (closes the carryover of comment #7 — the crate parses `aud` strictly and rejects non-string entries).
- Multi-aud `azp` check via the same verifier.

API delta is small: `OidcClient::new / authorize_url / exchange_code` keep the same shape. `authorize_url` is infallible now (no IdP-side data is touched). `PkcePair` exposes only `verifier: String` — the crate constructs the SHA-256 challenge from the verifier each time via `PkceCodeChallenge::from_code_verifier_sha256`, so the BFF only ever persists one secret in `bff:login_state`. `IdTokenClaims` is a small projection over `openidconnect::IdTokenClaims<BffAdditionalClaims, CoreGenderClaim>` carrying just the fields the handlers actually consume (iss, sub, sid, email, name). A small `BffAdditionalClaims { sid: Option<String> }` keeps the back-channel-logout claim addressable through the crate's `AdditionalClaims` trait.

Mitigation noted by the upstream docs: the reqwest client now disables 3xx redirect following on token / userinfo / JWKS fetches (SSRF against the provider's hostname).

The 24 hand-rolled validation tests are gone — they covered the crate's responsibilities now. The remaining `pkce_verifier_is_url_safe_no_pad` / `pkce_verifier_is_random` / `additional_claims_round_trips_sid` / `additional_claims_omits_sid_when_absent` tests stay so a future crate bump can't silently change the verifier encoding or the `sid` claim wire shape.

Workspace lints clean (`pedantic = deny`, `unwrap_used = deny`, `expect_used = deny`). All 81 unit tests pass, including the Redis round-trip integration test.

mitasovr:
**Finding (HIGH): `decode_session` silently coerces missing required fields to empty/zero defaults**

`src/backend/services/api-gateway/src/bff/session_store.rs` — `decode_session`:

```rust
let get = |name: &str| -> String {
    pairs
        .iter()
        .find(|(k, _)| k == name)
        .map(|(_, v)| v.clone())
        .unwrap_or_default()
};
let parse_i64 = |name: &str| -> Result<i64, BffError> {
    let raw = get(name);
    if raw.is_empty() {
        return Ok(0);
    }
    ...
};
```

Every string field falls back to `""` and every numeric field to `0` when absent. As a result, a partial or corrupt Redis HASH (partial write, mis-migration, manual edit) decodes into a "present" `SessionRecord` instead of an error:

- A record with a valid `expires_at` but missing identity fields passes all `/auth/me` checks and returns an **empty `csrf_token`** to the SPA.
- In Phase 2, the router JWT-cache would mint claims from empty `sub`/`tenant`.

The test `decode_session_handles_missing_optional_fields` currently pins this behavior as if `user_id`/`expires_at` were optional — but they are required for correct session semantics.

**Proposal**: keep the empty-pairs → `Ok(None)` path (missing key), but after decoding a non-empty hash, bail with `BffError::Internal` when any of `user_id`, `tenant_id`, `csrf_token` is empty or `expires_at == 0`. Flip the existing test to pin rejection of such records.

cyberantonz:
@coderabbitai review

mitasovr (approved):

PR #1281, opened 2026-06-09
feat(ingestion): data-quality checks as dbt tests with scheduled findings

aleksdotbar:
Run data-quality checks as dbt tests and emit one structured JSON finding per check to stdout for the central log store. store_failures keeps the violating rows for drill-down; checks are severity=warn so a finding never blocks ingestion. A standalone daily CronWorkflow runs the data_quality-tagged catalog, decoupled from the per-connector pipeline.

Checks read silver/gold only (present regardless of connector set) — enforced by audit_dq_no_bronze.py — so the catalog is connector-agnostic. Includes five starter checks (collab source-uniqueness, meeting grain, duration caps; HR working-hours bound; gold ic_kpis bounds), an ADR, and an authoring guide.

## Summary by CodeRabbit

* **New Features**
  * Introduced a scheduled data-quality framework that runs non-blocking dbt checks and emits one structured JSON finding per check for centralized logging.

* **Tests**
  * Added multiple data-quality tests covering collaboration metrics, KPI bounds, and HR working-hours, preserving full failing rows for audit.

* **Documentation**
  * Added an ADR and a testing README describing design, conventions, and authoring guidance.

* **Chores**
  * Added deployment templates and chart values to schedule and run the data-quality workflows.


mitasovr:
> [!NOTE]
> **Resolved — both points are already implemented in the current HEAD (as of `067236a5`, 2026-06-10).** I reviewed against a stale diff snapshot; apologies for the noise.
>
> - Catalog scoping is hardcoded: the cron runs `dbt test --selector data_quality` via the named selector in `selectors.yml`, with no operator-overridable selector and no empty default.
> - The tag is effectively reserved for singular tests: `selectors.yml` sets `indirect_selection: empty` (a mis-tagged model won't drag its generic tests in), and `tests/README.md` documents the tag as singular-test-only.
>
> Original comment left below for history.

---

<details>
<summary>Original comment (stale)</summary>

Two requests on the catalog-scoping mechanics:

**1. Make `data_quality` a reserved tag — tests only, never models.**

dbt tags are a single namespace shared by all node types, and `dbt test` applies *indirect selection*: if a **model** ever carries `tag:data_quality`, all of its generic tests (`not_null`/`unique`/`relationships`, `severity=error`, some referencing connector-specific relations that may not exist on a tenant) silently join the scheduled run — turning the monitoring cron red and breaking the "curated opt-in catalog" contract. Nothing prevents that today except convention.

Please:
- state explicitly in `tests/README.md` that `data_quality` is reserved for singular tests and must never be applied to models/seeds/snapshots;
- enforce it the same way the no-bronze rule is enforced — extend `audit_dq_no_bronze.py` (or add a sibling check) to fail when any non-test node carries the tag, so the rule can't rot.

**2. Drop the empty default for the selector — better yet, hardcode it.**

`data-quality-cron.yaml` renders `value: {{ default "" .Values.ingestion.dataQuality.select | quote }}`. An empty selector makes the shell drop `--select` entirely (`${DBT_SELECT:+...}`), so a bare `dbt test` runs the *whole* test catalog — including untagged structural tests at `severity=error` and tests against connector-specific tables that are absent on tenants without that connector (`UNKNOWN_TABLE` → run errors). I.e. the fallback value is guaranteed-broken on an arbitrary tenant, which is exactly the silent-default failure mode our no-defaults rule exists for.

Since catalog membership is already defined by the tag itself, the selector doesn't need to be an operator knob at all: hardcode `tag:data_quality` in the CronWorkflow and remove `ingestion.dataQuality.select` from values. Manual scoped/debug runs are still covered by submitting the `dbt-test` WorkflowTemplate directly with an explicit `dbt_select`. If per-env scoping is genuinely needed later, it should come back as `{{ required ... }}`, not `default ""`.

</details>

mitasovr (approved):

PR #1282, opened 2026-06-09
feat(chatgpt-team): ChatGPT Team connector (browser-proxy) — Bronze→Gold + analytics-api

mozhaev-dev:
Closes #497

## ChatGPT Team connector — browser-proxy architecture, Bronze → Gold + analytics-api

New declarative connector for **ChatGPT Team / Business** usage analytics, modelled on the existing `claude-team` connector. Delivers the full pipeline: ingestion → Silver → Gold (`ai_bullet_rows`) → analytics-api `query_ref` + `metric_catalog`.

> **Depends on** the secure-enclave MR `feat/chatgpt-team-proxy` (the customer-hosted browser proxy this connector reads through).

### Why a browser proxy
ChatGPT has no public admin/usage API. Like `claude-team`, the customer self-hosts a headless-Chromium proxy (Playwright + stealth) that holds the workspace session and exposes a thin bearer-auth `/api/*` surface; the session **never enters Insight infra**. This connector talks only to that proxy (`proxy_url` / `proxy_auth_token`), never to chatgpt.com directly. See `docs/components/connectors/ai/chatgpt-team/specs/` (PRD / DESIGN / ADR-001).

### What's included
- **Connector** (`connectors/ai/chatgpt-team`) — 5 streams: `seats`, `chat_activity` (incremental per-day), `codex_user_daily` (incremental per-day), `subscription_usage`, `subscription_balance`. The two subscription streams need a workspace `chatgpt_org_id`; it is **optional/blank** for analytics-viewer accounts, so `403/404` are IGNORE'd to keep the sync green.
- **Silver** — `chatgpt_team__ai_dev_usage` (Codex → `class_ai_dev_usage`, `tool='codex'`) and `chatgpt_team__ai_assistant_usage` (chat → **`class_ai_assistant_usage`**, `tool='chatgpt'`, `surface='chat'`). Wired into `silver/ai/*` via `union_by_tag` + `depends_on`; `chatgpt_team` added to the source `accepted_values` (fixes a `dbt test` blocker).
- **Gold** — `ai_bullet_rows` migration adds a Codex branch (`codex_active` / `codex_lines` / `codex_sessions`) and a chat branch (`chatgpt_active` / `chatgpt`).
- **analytics-api** — un-stubs `codex_active` / `chatgpt` in the AI Bullet `query_ref`; adds `codex_lines` / `codex_sessions` (sumIf) + `chatgpt_active` (countIf); `'chatgpt_active'` added to `ACTIVE_LIST`. Seeds `metric_catalog` + product-default `metric_threshold` for the 5 keys. Unit tests included.
- **descriptor** bumped **1.0.1 → 1.1.0** per ADR-0015 (Silver dbt models were added to the dbt scope, so reconcile must re-run the dbt step).

### Verification (local e2e on kind-insight)
- **Bronze** distinct unique_key: seats 41, chat_activity 238, codex_user_daily 55; subscription streams empty (no billing/org_id — expected).
- **Silver**: `class_ai_dev_usage` 55 rows, `class_ai_assistant_usage` 79 rows. Field fix found live: seats `added_at ← created_time`.
- **dbt test**: PASS=45. **`cargo test`** (analytics-api migrations): 11 passed.
- **Gold/`query_ref`** aggregation validated by direct ClickHouse queries (codex_active 55, chatgpt 621, etc.).

### Not yet runtime-validated in the live UI
Rendering in the product still needs an analytics-api image rebuild+redeploy (runs the SeaORM migration → `query_ref` + catalog) and an FE build. The FE-side change is a separate PR on `insight-front`.

### Known follow-ups
1. Subscription streams (`subscription_usage` / `subscription_balance`) unverified — no billing-capable account / `org_id` available in test.
2. `docs/components/connectors/ai/README.md` flags the dir name as stale (ChatGPT Team → **Business**, 2025-08-29) — worth reconciling docs.

## Summary by CodeRabbit

* **New Features**
  * Added ChatGPT Team connector using a customer-hosted browser proxy to ingest seats, daily per-user chat and Codex usage, and subscription data; produces Bronze→Silver outputs for assistant and developer usage.

* **Documentation**
  * Added ADR, design spec, PRD, README, connector descriptor, dbt schemas, and secret example documenting proxy deployment, config, streams, and operational guidance.

* **Chores / Migrations**
  * Added DB migrations and metric catalog seeds, plus metric/rollup label fixes and view updates to surface new ChatGPT Team metrics.


src/ingestion/connectors/ai/chatgpt-team/connector.yaml
```diff
@@ -0,0 +1,513 @@
...
+          current_streak: { type: ["number", "null"] }
+          text_tokens: { type: ["number", "null"] }
+          lines_added: { type: ["number", "null"] }
+    retriever:
+      type: SimpleRetriever
+      requester:
+        <<: *base_requester
+        path: "/api/wham/analytics/usage-leaderboard"
+        request_parameters:
+          window_days: "30"
```
mitasovr: **Data-correctness question — `window_days: "30"` vs `step: P1D`.**

The datetime cursor walks one day per request (`start_date` = `end_date`), but the request also pins `window_days: 30`. If `usage-leaderboard` honors `window_days` rather than the explicit date range, every "daily" row is actually a 30-day rolling aggregate attributed to a single date — and the Gold sums (`codex_lines`, `codex_sessions`) would be overcounted by up to ~30×.

The row counts in the PR verification don't rule this out (the row count would be identical; only the values differ). Could you confirm on the live workspace that a single-day response matches the admin UI for that day, and drop `window_days` if the endpoint doesn't require it?

src/ingestion/connectors/ai/chatgpt-team/connector.yaml
```diff
@@ -0,0 +1,513 @@
...
+    cursor_datetime_formats:
+      - "%Y-%m-%d"
+    datetime_format: "%Y-%m-%d"
+    start_datetime:
+      type: MinMaxDatetime
+      datetime: "{{ config.get('start_date') or day_delta(-7, format='%Y-%m-%d') }}"
+      datetime_format: "%Y-%m-%d"
+      # Earliest data point observed in the data_collector audit for the
+      # test org — going earlier returns empty pages. Verify per tenant.
+      min_datetime: "2026-01-14"
```
mitasovr: `min_datetime: "2026-01-14"` is the earliest data point of the **test org**, baked into the connector for all tenants. A tenant with older history can't backfill past this date, and the failure mode is silent (empty pages look expected).

Suggest removing the floor or making it config-driven — the "verify per tenant" comment can't be acted on once this ships.

src/ingestion/connectors/ai/chatgpt-team/connector.yaml
```diff
@@ -0,0 +1,513 @@
...
+    end_datetime:
+      type: MinMaxDatetime
+      datetime: "{{ day_delta(-1, format='%Y-%m-%d') }}"
```
mitasovr, line 68: No `lookback_window` on the day cursor: the Airbyte state advances past `end_datetime = yesterday`, so if chatgpt.com analytics arrive late — or "yesterday" is still accumulating at the 04:00 UTC sync (likely, since the source's reporting timezone is unverified) — those days stay permanently undercounted. The 3-day lookback in the dbt models doesn't help, because Bronze never receives the revised rows.

`claude-team` has the same gap, so this could be a shared follow-up: `lookback_window: P2D` on both day cursors.

src/ingestion/connectors/ai/chatgpt-team/connector.yaml
```diff
@@ -0,0 +1,513 @@
...
+  # analytics-viewer account with no billing visibility) the request hits
+  # /subscriptions//usage → 404; a billing-gated session returns 403. Both
+  # are benign here: IGNORE them so the stream commits zero records and the
+  # sync stays GREEN (the seat/chat/Codex streams are unaffected). A real
+  # org_id with billing access starts populating rows automatically.
+  billing_error_handler: &billing_error_handler
+    type: DefaultErrorHandler
+    response_filters:
+      - type: HttpResponseFilter
+        http_codes: [403, 404]
```
mitasovr, line 103: IGNORE on 404 also masks a mistyped `chatgpt_org_id` or a broken proxy route **forever** — the classic green-but-empty failure. Keeping it for the blank-org_id case is fine, but please add a runbook/monitoring note: e.g. flag when the subscription streams return 0 rows for N consecutive days while `chatgpt_org_id` is set.

src/ingestion/connectors/ai/chatgpt-team/README.md
```diff
@@ -0,0 +1,84 @@
...
+./src/ingestion/tools/declarative-connector/source.sh validate-strict ai/chatgpt-team
+./src/ingestion/tools/declarative-connector/source.sh check           ai/chatgpt-team <tenant>
+```
+
+> ⚠️ **Unverified against a live instance.** Endpoint shapes, fields, and the
+> proxy's access-token flow are derived from the `data_collector` prototype
+> and must be confirmed once workspace credentials are available
+> (spec OQ-CGT-3/4/6).
+
+## Silver Targets
```
mitasovr, line 74: This section contradicts the PR itself: the Silver models ship **in this same PR**, and they feed `class_ai_assistant_usage` (not `class_ai_tool_usage` — that's the superseded v1.0 spec name). "Only the bronze→RMT promotion (`dbt/`) exists today" is already stale.

src/ingestion/connectors/ai/chatgpt-team/README.md
```diff
@@ -0,0 +1,84 @@
...
+1. A **workspace Owner/Admin** account on the ChatGPT Team/Enterprise workspace
+   (the analytics/subscription endpoints are admin-only).
+2. A deployed `chatgpt-team-proxy` with the session installed
+   (`POST /admin/session-key`) and reachable from Insight.
+3. The workspace `account_id` and `org_id` (from `chatgpt.com/api/auth/session`).
+
+## K8s Secret
+
+See [`src/ingestion/secrets/connectors/chatgpt-team.yaml.example`](../../../secrets/connectors/chatgpt-team.yaml.example).
+Required: `chatgpt_account_id`, `proxy_url`, `proxy_auth_token`.
```
mitasovr, line 32: `chatgpt_org_id` is listed as Required here (and the comment in `chatgpt-team.yaml.example` also says "Required"), but `descriptor.yaml` correctly puts it under `optional_fields` — and the whole 403/404-IGNORE design depends on it being optional. Please align the README and the secret example with the descriptor.

src/ingestion/connectors/ai/chatgpt-team/dbt/chatgpt_team__ai_dev_usage.sql
```diff
@@ -0,0 +1,91 @@
...
+    -- becomes a no-op once promote_bronze_to_rmt merges (ADR-0002), but guards
+    -- against duplicate raw rows from multiple sync attempts.
+    SELECT *
+    FROM {{ source('bronze_chatgpt_team', 'chatgpt_team_codex_user_daily') }}
+    ORDER BY _airbyte_extracted_at DESC
+    LIMIT 1 BY tenant_id, source_id, email, date
+)
+WHERE email IS NOT NULL
+  AND trim(email) != ''
+  AND date IS NOT NULL
```
mitasovr, line 88: Unlike the chat model (which filters out all-zero rows), this model has **no activity filter**: if `usage-leaderboard` ever returns zero-usage users, each such row still emits `codex_active = 1` in the Gold view and inflates the DAU marker.

Either add a filter (`n_turns > 0 OR credits > 0 OR n_threads > 0 OR lines_added > 0`) or document that the endpoint is verified to return only active users.

src/ingestion/connectors/ai/chatgpt-team/dbt/chatgpt_team__ai_assistant_usage.sql
```diff
@@ -0,0 +1,96 @@
...
+    CAST(_airbyte_extracted_at AS Nullable(DateTime64(3)))          AS collected_at,
+    toUnixTimestamp64Milli(_airbyte_extracted_at)                   AS _version
+FROM (
+    -- Bronze dedup: keep the latest extract per (email, date). Defensive depth —
+    -- becomes a no-op once promote_bronze_to_rmt merges (ADR-0002), but guards
+    -- against duplicate raw rows from multiple sync attempts.
+    SELECT *
+    FROM {{ source('bronze_chatgpt_team', 'chatgpt_team_chat_activity') }}
+    ORDER BY _airbyte_extracted_at DESC
+    LIMIT 1 BY tenant_id, source_id, email, date
```
mitasovr: Nit: the dedup is `LIMIT 1 BY tenant_id, source_id, email, date` on the **raw** email, while `unique_key` uses `lower(trim(email))` — two case-variant spellings of the same address survive dedup and then collide on `unique_key`, failing the unique test. Same pattern in `chatgpt_team__ai_dev_usage.sql`.

Cheap fix: `LIMIT 1 BY tenant_id, source_id, lower(trim(email)), date`.

src/ingestion/connectors/ai/chatgpt-team/descriptor.yaml
```diff
@@ -0,0 +1,46 @@
...
+# Strict semver per ADR-0015. 1.0.0: initial MVP — browser-proxy
+# architecture (mirrors claude-team). Reads chatgpt.com web data via the
+# customer-deployed chatgpt-team-proxy; requires `proxy_url` (no default),
+# `proxy_auth_token`, and the workspace `chatgpt_account_id` / `chatgpt_org_id`.
+# 1.0.1: tolerate 403/404 on subscription streams (blank org_id).
+# 1.1.0: add Silver dbt models (chatgpt_team__ai_dev_usage,
+#   chatgpt_team__ai_assistant_usage) — MINOR bump per ADR-0015 so reconcile
+#   re-runs the dbt step (tag:chatgpt-team+ now spans the new staging models
+#   and their downstream silver.class_ai_* unions).
+version: "1.1.0"
```
mitasovr: Nit: a brand-new connector shipping at `1.1.0` with a narrated 1.0.0/1.0.1 history (which never existed on `main`) reads oddly under ADR-0015 — the first released version could simply be `1.0.0`. Harmless either way, since the reconcile loop only compares versions going forward.

mitasovr (commented):
## Review summary

Strong, well-documented PR. I verified the highest-risk invariants locally and they all hold:

- **Positional column contracts** (`union_by_tag` does positional UNION ALL): `chatgpt_team__ai_dev_usage` matches `claude_team__ai_dev_usage` (all 26 columns, order and types), and `chatgpt_team__ai_assistant_usage` matches the `chat` CTE of `claude_enterprise__ai_assistant_usage`. ✅
- Gold view migration: Branches 1–3 are byte-identical to `20260601000000_ai-claude-team-metrics.sql`. ✅
- Catalog seed idempotency: `metric_catalog` has UNIQUE on `metric_key` and `metric_threshold` uses sentinel columns, so `ON DUPLICATE KEY UPDATE` will actually fire and refresh the two ComingSoon rows. ✅
- Silver `tool` accepted_values already include `'chatgpt'` / `'codex'`. ✅

Inline comments cover the actionable items. The three I'd consider blocking: the `window_days=30` semantics question (potential ~30× overcount), the README/descriptor contradiction about `chatgpt_org_id` and the stale Silver Targets section, and the hardcoded `min_datetime`.

---

## Proxy review (`secure-enclave/proxies/chatgpt_team` — no MR exists, so leaving findings here)

The code itself is in good shape: session and access_token live only in memory and are never logged (only `token_len`), bearer comparison is constant-time, shutdown is graceful, and the transport seam is clean. The token-refresh logic (proactive 40-min TTL + one retry on 401) is sound. Findings, in priority order:

1. **TLS is not enforced anywhere.** The server is plain `http.createServer`, and `deploy/` ships a Service but no Ingress/TLS. Insight will call the customer-hosted proxy across the internet with `Authorization: Bearer <token>` — without mandatory TLS termination, the token and workspace data travel in cleartext. Either ship an Ingress+TLS example or make the README require an https `proxy_url` explicitly.
2. **Liveness probe on `/health` (the deployment's own TODO).** `/health` returns 503 until the session is installed; with `failureThreshold: 100 × 30s` the pod gets killed ~50 min after start if the operator hasn't posted the session yet, breaking their port-forward and looping. A `tcpSocket` liveness or a dedicated `/live` endpoint is a 10-line fix — worth doing before customer handoff.
3. **A dead session is invisible to monitoring.** When the session expires, `refreshAccessToken` fails silently, upstream fetches return 401 to the connector, but `isReady()` stays `true` and `/health` stays green. DESIGN tells the operator to "rotate when /health returns 503" — that signal never fires for token death. Suggest surfacing auth staleness in `/health` after N consecutive refresh failures/401s.
4. **Image reference in `deploy/deployment.yaml`:** `ghcr.io/cyberfabric/chatgpt-team-proxy:latest` — (a) the org has moved, it should be `ghcr.io/constructorfabric/chatgpt-team-proxy`; (b) for a customer-hosted component, pin a version/digest instead of `:latest` + `imagePullPolicy: Always` (reproducibility / supply chain).
5. `setSessionKey` during in-flight fetches does `clearCookies` + `goto` on the shared page, destroying live `page.evaluate` contexts. Rotation is rare and manual, so acceptable — but document "rotate outside sync windows" or serialize rotation behind in-flight requests.
6. No 429 handling/throttling: a full backfill from 2026-01-14 is hundreds of sequential requests (day-walk × pages of 100). Airbyte's retry on 429/5xx probably suffices, but Cloudflare rate limits are worth watching on first real backfill.
7. `deploy/secret.yaml` says "minimum 32 characters" while `config.ts` enforces ≥16 — align the two.
8. Minor: `POST /api/*` returns 404 (405 would be accurate); `parseCookies` forces `httpOnly: false` on all cookies.

mitasovr (approved):

PR #1288, opened 2026-06-11
fix(ingestion): salesforce connector + dbt

aleksdotbar:
- Rewrite the salesforce connector to REST `/queryAll` only; remove the Bulk API
  path.
- Per-request OAuth token refresh — syncs longer than the SF session timeout no longer fail on expired tokens.
- `num_workers` typed as string (K8s Secret stringData), default 1;
  `stop_sync_on_stream_failure=false`; drop `salesforce_streams` override.
- Align salesforce dbt staging models with native bronze column types
  (DateTime64/Date32) and drop User fields the org never exposes.
- Map only 003-prefixed WhoId to `contact_id`.
- Per tenant+source incremental watermarks in salesforce staging and CRM silver
  classes; drop populated silver placeholders (row-count guard locked in stale
  schemas).

## Summary by CodeRabbit

* **Changes**
  * Salesforce ingestion is now REST `/queryAll` by default (Bulk API 2.0 behavior removed).
  * Default REST concurrency reduced to 1.
  * Incremental syncing now uses per-tenant/per-source high-water mark tracking for more reliable catch-up.
  * Stream selection now follows a curated connector-managed set (catalog-driven only).

* **Bug Fixes / Reliability**
  * Long-running syncs refresh authentication tokens per request.
  * Improved handling for 401 and request-limit failures so individual stream errors don’t halt the entire sync.

* **Documentation**
  * Updated connector configuration, concurrency/window settings, and build/deploy scripts.


mitasovr:
## Code review notes

Solid, well-motivated surgery — the Bulk→REST removal is clean (no dangling references to `BulkNotSupportedException`, `availability_strategy`, `salesforce_streams`, or `job_tracker` anywhere on the branch), and the per-tenant watermark + token-refresh fixes are real bugfixes. A few things to address before merge:

### 1. `_get_updated_state` is dead code — remove it rather than patching it
[`streams.py:582`](https://github.com/constructorfabric/insight/blob/feat/salesforce-sync-resilience/src/ingestion/connectors/crm/salesforce/source_salesforce/streams.py#L582)

CodeRabbit flagged a state-regression bug here, but the method is never called — state is owned by `ConcurrentCursor` via `StreamFacade`, so the regression is unreachable at runtime. The real problem is the opposite: this PR removed the `_slice = None` class attribute, yet line 596 still reads `self._slice`. If anything ever does call this method, it'll raise `AttributeError`. Please delete `_get_updated_state` entirely instead of applying CodeRabbit's suggested clamp fix.

### 2. Bronze type change needs a rollout plan for already-synced tenants
[`salesforce__crm_accounts.sql:33`](https://github.com/constructorfabric/insight/blob/feat/salesforce-sync-resilience/src/ingestion/connectors/crm/salesforce/dbt/salesforce__crm_accounts.sql#L33) (and the other staging models)

Staging now reads native bronze types (`CreatedDate AS created_at`, `toUnixTimestamp64Milli(SystemModstamp)`, `DurationInMinutes * 60` without `toInt64OrNull`). On any environment where `bronze_salesforce` tables were already created by the old Bulk/CSV path with `String` columns, these models will break. There's no migration or README note describing the cutover. What's the plan for tenants that already synced under the old connector — drop `bronze_salesforce` + full resync, or `ALTER` the columns?

### 3. `coalesce(toUnixTimestamp64Milli(SystemModstamp), 0)` can silently drop rows — what's the intended behavior?
[`salesforce__crm_accounts.sql:36`](https://github.com/constructorfabric/insight/blob/feat/salesforce-sync-resilience/src/ingestion/connectors/crm/salesforce/dbt/salesforce__crm_accounts.sql#L36) (same pattern in every staging + silver model)

A row with a NULL `SystemModstamp` gets `_version = 0`. After the first incremental run advances the high-water mark, that row can never satisfy `_version > hwm` again — it's silently excluded from all subsequent runs. In Salesforce `SystemModstamp` is effectively always populated, so the practical risk is low, but the failure mode is invisible.

Question: is `0` the deliberate choice here, or should a NULL modstamp fall back to something monotonic (e.g. `CreatedDate`/`collected_at`) so the row stays eligible? At minimum I'd add a comment documenting that `_version = 0` rows are intentionally one-shot. How do you want to handle this?

### 4. `_DEFAULT_CONCURRENCY = 1` on the concurrent CDK — watch the slice count
[`source.py:58`](https://github.com/constructorfabric/insight/blob/feat/salesforce-sync-resilience/src/ingestion/connectors/crm/salesforce/source_salesforce/source.py#L58)

We've previously hit a concurrent-CDK self-deadlock at low concurrency with a large partition count (jira/confluence froze at default_concurrency:1 with ≥10k partitions — fingerprint "Records read: N" stuck). Here ~25 slices × 10 streams ≈ 250 partitions, which should be safe. But the risk profile returns if anyone shrinks `salesforce_stream_slice_step` (e.g. P30D → P1D produces thousands of slices) while workers stay at 1. The `max(1, concurrency_level // 2)` guard is correct and necessary — without it the default of 1 would pass 0 initial partitions. Just flagging the slice-step interaction as a documented gotcha.

### 6. Dropping the row-count guard in the placeholder macro — reasoning is sound
[`drop_silver_placeholders_at_start.sql:89`](https://github.com/constructorfabric/insight/blob/feat/salesforce-sync-resilience/src/ingestion/dbt/macros/drop_silver_placeholders_at_start.sql#L89)

The rationale in the macro comment is convincing — a placeholder can fill mid-run (staging materializes after the hook has already run, the silver model inserts into the placeholder with `on_schema_change=ignore`), and contents are always rebuildable from staging, so a row-count guard would block the drop forever. The residual risk is that *any* populated table carrying the `INSIGHT_PLACEHOLDER_v1` marker now gets dropped, but since only `create-bronze-placeholders.sh` ever sets that marker, this is acceptable. No change requested — just confirming the two-factor logic checks out.

---

Also: CI is red for process reasons unrelated to this PR — the E2E failure is the `identity` DB fixture (`test_migrations_create_insight_database`), fixed in #1287 which isn't in this branch's base, and DCO needs `Signed-off-by`. A `git rebase --signoff` onto `main` should clear both.

mitasovr (approved):

mitasovr (approved):

PR #1290, opened 2026-06-11
fix(ingestion): require PyYAML in parse_descriptor (drop hand-rolled fallback)

SharedQA:
## Summary

`parse_descriptor.py` shipped a hand-rolled fallback YAML parser for hosts without PyYAML — that path only ever ran under the retired host-python `dev-up.sh`. Every caller now runs where PyYAML is present: the reconcile toolbox image ships it transitively via `dbt-clickhouse`, and CI installs it explicitly. The fallback was dead weight that could silently diverge from `yaml.safe_load` (the original CONN-11 symptom: a block-style `secret.required_fields` came back as `{}`).

So instead of hardening the fallback, this drops it and makes PyYAML a hard requirement that fail-fasts with an install hint.

## Changes
- `parse_descriptor.py`: remove the fallback; `import yaml` failure → `exit 2` with `pip install pyyaml` guidance. ~82 → ~45 lines.
- `test_parse_descriptor.py`: full-corpus parse, dotted-path `_walk` semantics (incl. `secret.required_fields`), and a fail-fast test (shadows `yaml` → asserts exit 2). **34 tests pass** against the current connector corpus.

## Dropped since the first revision
- The earlier `_helpers.tpl` hunk (omit the `app.kubernetes.io/version` label when `appVersion == "---"`) has been **removed — it is now dead code**:
  - #1303 fixed the root cause of the `"---"` appVersion (yq v4 emitting `---` document separators), restored a real committed `appVersion`, and added a fail-fast publish guard so `"---"` can never ship again.
  - #1431 removed `dev-up.sh` — the only install-from-source path that consumed the committed placeholder.
  - The committed `Chart.yaml` `appVersion` on main is now a real build tag, so the `if ne .Chart.AppVersion "---"` condition is unreachable.
- The earlier `dev-up.sh` / `.env.*.example` edits were already dropped (those files were removed from main in the dev-up retirement).

This PR is now scoped purely to the `parse_descriptor.py` fallback removal + its tests.


dev-up.sh
```diff
@@ -101,6 +101,18 @@ OIDC_AUDIENCE="${OIDC_AUDIENCE:-}"
...
 IDENTITY_TENANT_DEFAULT_ID="${IDENTITY_TENANT_DEFAULT_ID:-}"
 
+# Insight tenant slug for the reconcile loop (Airbyte source/connection
+# naming per ADR-0005). The chart `required`s ingestion.reconcile.tenantId,
+# so the umbrella install dies without it. Defaulting is safe ONLY for a
+# single-developer Kind cluster; on a shared remote cluster a silently
+# wrong slug would misname Airbyte sources/connections for everyone, so
+# remote mode fail-fasts instead.
+if [[ "$CLUSTER_MODE" == "local" ]]; then
+  TENANT_ID="${TENANT_ID:-local}" # RULE-DEFAULTS-OK: dev-only local Kind cluster; 'local' is the documented single-developer slug (.env.local.example)
```
mitasovr: usually we avoid using default values

SharedQA:
Both CodeRabbit nitpicks (E702 semicolon split, RUF021 parentheses) addressed in 2bb50eb. No behavior change; all 44 parser tests pass.

dev-up.sh
SharedQA: Agreed — dropped the default entirely in 7d0ee87. `dev-up.sh` now fail-fasts on missing TENANT_ID in every mode (`${TENANT_ID:?...}`); the env file is the single source of truth, with `.env.local.example` shipping `local` for single-developer Kind.

ktursunov (changes requested):
Dev-up.sh script is reworked into dev compose script - https://github.com/constructorfabric/insight/pull/1431
Please revalidate locally with it's merged. 

About yaml parser fallback - suggest to use library parser instead of self written alternative as more stable

SharedQA:
Reworked per review: rather than harden the hand-rolled fallback, it's gone — PyYAML is now required (it's already present everywhere reconcile runs: the toolbox image via dbt-clickhouse, and CI installs it explicitly). That removes the divergence-from-`yaml.safe_load` risk entirely instead of chasing parity. Rebased onto the dev-up retirement, so the `dev-up.sh`/`.env` edits are dropped; the chart `appVersion` label fix stays. Tests repurposed to cover the corpus + the PyYAML-absent fail-fast.

ktursunov:
**Test placement: `parse_descriptor` is a `reconcile-connectors` helper, but its test lands in the e2e rig**

Nice cleanup — dropping the hand-rolled fallback and fail-fasting on missing PyYAML is the right call, and the tests themselves (corpus parse, dotted-path `_walk`, the CONN-11 `secret.required_fields` repro, fail-fast) are good.

One placement nit on `test_parse_descriptor.py`: it's a pure unit test of a `reconcile-connectors` CLI helper (`src/ingestion/reconcile-connectors/python/parse_descriptor.py` — stdlib + PyYAML only, no infra), but it's landing under `src/ingestion/tests/e2e/meta/`, which is the e2e rig's own turf:

- `pytest.ini` documents `smoke` as *"framework smoke tests"*, and its `meta/` siblings (`test_dbt_runner`, `test_ref_resolver`, `test_expect_engine`) all test the **harness**, not product/connector code.
- It's collected only via `testpaths = metrics meta api` and run through `./e2e.sh` → `docker compose run --rm runner pytest`, i.e. inside the containerized e2e runner. So a zero-dependency helper test gets gated behind the full e2e pipeline instead of running as a fast standalone check.

I'd colocate it with the code it tests, under `src/ingestion/reconcile-connectors/` (e.g. a `python/tests/` dir). We already have precedent for package-local tests — the connectors (e.g. `connectors/git/gitlab/`) ship their own `tests/` + `pyproject.toml` with `testpaths = ["tests"]`.

The real cost to be upfront about: `reconcile-connectors` has no `pyproject.toml` / test harness / CI collection today, so relocating the file needs a runner wired up (a `tests/` dir + a CI job that collects it) or the test silently stops running. Two options:

1. Establish that minimal harness here and move the test — it's the natural long-term home for `reconcile-connectors/python/*` tests, which currently have none.
2. If that's out of scope for this fix, keep it in `meta/` for now but track the relocation as a follow-up.

Happy to take the relocation + harness setup as a follow-up if you'd rather keep this PR tight.

cyberantonz (approved):

ktursunov (approved):

PR #1314, opened 2026-06-12
Migrate to gears

cyberantonz:
## Summary by CodeRabbit

* **Documentation**
  * Updated configuration docs and examples to use gear-based naming and env-vars (APP__gears__...), and revised gateway/auth setup guidance.

* **Chores**
  * Migrated backend to a new toolkit and updated workspace dependencies.
  * Refined configuration layouts and Helm/secret examples; introduced zero-config tenant derivation for gateway.

* **Behavior**
  * Proxy error mapping simplified to return a canonical service-unavailable response; authenticated-404 fallback removed.


mitasovr (approved):

SharedQA:
@cyberantonz — flagging an overlap so we don't double-edit the same files. The gears migration here touches files that QA PRs already in review also change:

- **#1327** (api-gateway unit tests) edits the same `auth_info.rs`, `main.rs`, `proxy.rs`, and asserts the **pre-gears** config shape (old `APP__modules__` / `from_config` layout). After the `APP__gears__` rename here, those assertions need updating.
- **#1333** (analytics-api unit tests) covers `domain/query.rs`, which this PR also changes.
- **#1317** (CI quality gates) has no code overlap, but its gates will run against the post-gears tree.

Suggested order: land #1314 first, then rebase #1327/#1333 so the tests assert the new gears config/keys instead of the old shape. Happy to do that rebase once this merges, or to review the auth-config/query changes against the existing tests beforehand if that's easier. Just want to avoid conflicting edits on the same files.

PR #1327, opened 2026-06-13
test(api-gateway): unit tests for OIDC auth-config + CLI (crate ~0%→71%)

SharedQA:
Unit-test coverage for the `api-gateway` crate, the workspace's biggest unit gap (was 2 tests, only `proxy.rs` partial).

## Result (measured, `cargo-llvm-cov`)
| File | Before | After (lines) |
|---|---|---|
| `auth_info.rs` | 0% | **70.5%** |
| `main.rs` (CLI) | 0% | **64.7%** |
| `proxy.rs` | ~73% | 73% |
| `core_types.rs` | 0% | 0% (runtime-only) |
| **crate TOTAL** | ~near-0 | **70.7% lines** (regions 73.0%, fns 74.0%) |

27 tests pass; `cargo clippy --all-targets --all-features` clean (no `unwrap`/`expect` in tests, per the workspace lint).

## What's covered
- **OIDC auth-config transform** — extracted `AuthInfoResponse::from_config` (scopes split on whitespace, `response_type` pinned to `code`, field passthrough); config `default` + `deny_unknown_fields`; JSON wire-shape roundtrip.
- **CLI parsing** — default→`run`, all subcommands, `-v/-vv` count, `--config/-c`, `--print-config`, unknown-subcommand rejection, clap self-check.

## Not unit-tested
`main()`→`run_server`/`run_migrate`, `register_rest`, and `core_types::init` need the live `modkit` runtime (`ModuleCtx`, client hub) — that's **integration** scope (tracked separately), not unit. This PR covers the crate's pure/testable surface.

Opened as draft for review before it's marked ready.

## Release Notes

* **Tests**
  * Added unit coverage for authentication config generation (including scope parsing and a consistent response type), CLI flag/subcommand parsing, and proxy routing/forwarding behavior (upstream failures mapped to 503, correct prefix/query preservation, and hop-by-hop header handling).
* **Chores**
  * Updated development dependencies for the API gateway service.


SharedQA:
@coderabbitai review

cyberantonz:
Please wait for a while, the project is migrating to gears version, so there will be some changes.

SharedQA:
Overlaps #1314 (migrate to gears) on `auth_info.rs`/`main.rs`/`proxy.rs`. These tests assert the pre-gears `APP__modules__`/`from_config` shape; once #1314 lands, rebase to assert the `APP__gears__` config instead. Tracking so we reconcile rather than conflict.

src/backend/services/api-gateway/src/auth_info.rs
```diff
@@ -139,3 +149,98 @@ impl RestApiCapability for AuthInfoModule {
...
+            issuer_url: "https://idp.example/oauth2".to_owned(),
+            client_id: "spa-client".to_owned(),
+            redirect_uri: "https://app.example/callback".to_owned(),
+            scopes: scopes.to_owned(),
+        }
+    }
+
+    #[test]
+    fn from_config_splits_space_separated_scopes() {
+        let r = AuthInfoResponse::from_config(&cfg("openid profile email"));
```
cyberantonz, line 170: What do this test do? It check that you can fill rust structure with values? It is guaranteed with language itself.

SharedQA: The quoted lines are the `cfg()` fixture (the Arrange step) — the assertion is the next line: `assert_eq!(r.scopes, vec!["openid", "profile", "email"])`. The test isn't checking struct construction. `AuthInfoConfig.scopes` is a single space-separated `String` and `from_config` splits it into a `Vec<String>` via `split_whitespace().collect()`, so the test pins the OAuth2 scope-parsing. It would still compile with the wrong delimiter or with `split(' ')` (which emits empty tokens on tabs/double spaces) — that's why the sibling `from_config_collapses_extra_whitespace_and_tabs` exists too.

Agree with the general principle though: `config_default_is_all_empty` is the one that mostly exercises `#[derive(Default)]`. Happy to drop it if you'd prefer.

src/backend/services/api-gateway/src/auth_info.rs
```diff
@@ -139,3 +149,90 @@ impl RestApiCapability for AuthInfoModule {
...
+    fn config_default_is_all_empty() {
+        let c = AuthInfoConfig::default();
+        assert!(c.issuer_url.is_empty());
+        assert!(c.client_id.is_empty());
+        assert!(c.redirect_uri.is_empty());
+        assert!(c.scopes.is_empty());
+    }
+
+    #[test]
+    fn config_parses_known_fields() -> R {
```
cyberantonz, line 213: This and next tests check serde library behaviour. Is it ok? I think we can trust serde.

cyberantonz (commented):
Some issues

ktursunov:
OIDC and API Gateway are reworked in scope of https://github.com/constructorfabric/insight/issues/1583

PR #1328, opened 2026-06-13
test(e2e): analytics-api tenant-resolution failure matrix

SharedQA:
First Rust-side **integration** test on the bronze→API e2e rig.

## Scope
The original idea was a "JWT auth-failure matrix," but the e2e rig **bypasses the API gateway** (it spawns `analytics-api` directly with an `X-Insight-Tenant-Id` header — JWT auth is a gateway concern). The validatable equivalent at the analytics-api layer is **tenant resolution**, which is the actual tenant-isolation / data-correctness boundary.

## What it asserts (`meta/test_tenant_resolution.py`)
- valid tenant → **accepted** (200);
- missing / empty / nil-UUID / malformed tenant → **rejected (4xx)** — never silently served, which would let a caller resolve against data with no valid tenant scope.

## Validation
Runs in the e2e rig: `./e2e.sh test meta/test_tenant_resolution.py -m smoke`. Local rig run (docker compose: ClickHouse + MariaDB + a release build of analytics-api) is in progress; result will be posted here. CI is the authoritative runner.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for tenant-resolution middleware behavior, validating proper handling of missing, empty, malformed, and whitespace-containing tenant ID headers across all protected routes.


SharedQA:
Validated locally via the e2e rig: 5 passed in 86s (`./e2e.sh test meta/test_tenant_resolution.py -m smoke`). Valid tenant accepted (200); missing, empty, nil-UUID and malformed tenant each rejected with 4xx. The tenant middleware enforces tenant resolution as expected.

src/ingestion/tests/e2e/meta/test_tenant_resolution.py
```diff
@@ -0,0 +1,62 @@
...
+    ],
+)
+def test_unresolvable_tenant_is_rejected(
+    analytics_api: AnalyticsApiProcess, case: str, headers: dict[str, str]
+) -> None:
+    """Tenant isolation: a missing / empty / nil / malformed tenant must be
+    REJECTED with a 4xx — never silently served, which would let a caller read
+    data without a resolvable tenant scope."""
+    status = _health_status(analytics_api, headers)
+    assert 400 <= status < 500, (
```
ktursunov: Assert must contain exact expected status code

src/ingestion/tests/e2e/meta/test_tenant_resolution.py
```diff
@@ -0,0 +1,62 @@
...
+    assert status == 200, f"valid tenant should be accepted, got {status}"
+
+
+@pytest.mark.parametrize(
+    "case,headers",
+    [
+        ("missing", {}),
+        ("empty", {TENANT_HEADER: ""}),
+        ("nil-uuid", {TENANT_HEADER: NIL_TENANT}),
+        ("malformed", {TENANT_HEADER: "not-a-uuid"}),
```
ktursunov: Suggest split test case by each of the parameter - those are different cases and might trigger different responses - depends on requirements

src/ingestion/tests/e2e/meta/test_tenant_resolution.py
SharedQA: Done in d016d0ac — switched to `assert status == 400`. analytics-api maps an unresolvable/missing/nil/malformed tenant to the canonical `invalid_argument` envelope (`TENANT_UNRESOLVED`), which is 400 — pinned in `auth::tenant_unresolved_response` and the `api/tenant_resolution_tests.rs` parity tests, so the exact code is the right thing to assert and a drift to 401/403 now fails the test.

Also swept the rest of our test branches for the same anti-pattern (range/predicate status asserts) — this was the only test instance; the one other `is_success()` hit is production control-flow in `infra/identity/mod.rs`, which should stay tolerant of any non-2xx.

src/ingestion/tests/e2e/meta/test_tenant_resolution.py
SharedQA: Done in e84d0772 — split into four named tests (`test_missing_…`, `test_empty_…`, `test_nil_uuid_…`, `test_malformed_…`) sharing one exact-400 assertion helper.

I traced each input through `auth::read_session_tenant`: missing header (no first value), empty/malformed (`Uuid::parse_str` fails), and nil (parseable but dropped by the `!is_nil()` filter) all resolve to `None` → `resolve_tenant(None)` → (no configured default in this rig) → canonical `invalid_argument` = 400. So today they converge on the same code, but they're genuinely separate requirements, and per-case tests mean a future spec divergence shows up against the exact case rather than hiding in a matrix.

One more distinct branch `read_session_tenant` guards that isn't covered yet: a duplicate `X-Insight-Tenant-Id` header ("refuse to pick a winner" → None → 400). Happy to add it as a fifth case — it needs the helper to send multi-valued headers, so I left it out of this change to keep the diff focused.

src/ingestion/tests/e2e/meta/test_tenant_resolution.py
```diff
@@ -0,0 +1,87 @@
...
+# default in this rig) → the canonical `invalid_argument` envelope = 400
+# (`auth::tenant_unresolved_response`; pinned by `api/tenant_resolution_tests.rs`).
+# Asserting `== 400` (not a 4xx range) catches a drift to 401/403 that would mean
+# the isolation boundary or the envelope moved.
+
+
+def _assert_tenant_rejected_400(
+    api: AnalyticsApiProcess, headers: dict[str, str], why: str
+) -> None:
+    status = _health_status(api, headers)
```
ktursunov: Only health status is covered, api/analytics/v1/metrics/queries is not covered

ktursunov: api/analytics/v1/catalog/get_metrics - also should be covered

SharedQA: Both covered as of 31ad7417. The health-only helper this thread pointed at is gone — every case now runs through a single ROUTES table:

- /health
- POST /v1/metrics/{id}/query
- POST /v1/metrics/queries
- POST /v1/catalog/get_metrics

`_assert_accepted_everywhere` and `_assert_tenant_unresolved_everywhere` both loop that table, so /v1/metrics/queries and /v1/catalog/get_metrics are each exercised by all six tests (valid-tenant accept + the five reject cases: missing / empty / nil / malformed / multi-valued header), not just /health. Rejections assert the exact envelope (400 + application/problem+json + field_violations[0] = {tenant_id, TENANT_UNRESOLVED}), so an incidental 400 from body/Content-Type validation cannot pass for a tenant rejection. Live E2E suite is green on the latest commit.

cyberantonz (approved):

Issue #1332, opened 2026-06-13
data(threshold): eq operator uses f64::EPSILON — magnitude-dependent equality, likely wrong for large metric values


SharedQA:
## Description
While adding unit tests for `domain/threshold.rs`, the `eq` branch of `threshold_matches` looks semantically suspect:

```rust
"eq" => (value - threshold).abs() < f64::EPSILON,
```

`f64::EPSILON` (~2.22e-16) is the gap between 1.0 and the next double — it is **only** a meaningful tolerance near magnitude 1. For a metric value like `1000.0`, the representable gap (ULP) is ~`1e-13`, so two values that should compare equal can differ by **more than** `EPSILON` → `eq` returns **false** for values that are effectively equal. It is neither exact equality nor a relative tolerance.

For threshold rules on real metric values (hours, counts, ratios), this will mis-evaluate "equal" comparisons at common magnitudes.

## Why this matters for testing
A unit test written to make the current code green would **lock in** this behaviour. The added test instead **characterises** it and this issue flags it, so the intended semantics get decided rather than frozen.

## Decision needed
Define `eq` semantics:
- exact equality for integer-valued metrics (counts), or
- a relative tolerance (`abs(a-b) <= rel * max(abs(a),abs(b))`) / explicit absolute tolerance.

Then update `threshold_matches` + the test to the chosen rule. Low severity today (threshold eval is not yet wired — the fn is `#[allow(dead_code)]`), but it should be fixed before thresholds go live.

Issue #1331, opened 2026-06-13
data(time): align time formats — toDate() timezone-truncation risk + metric_date exposed as String at the API


SharedQA:
## Question raised
Are insert and fetch time formats aligned? Are we using epoch?

## What the audit found (across dbt + gold views)
- **`_version` = `toUnixTimestamp64Milli(now64())` → epoch-milliseconds (Int64).** Consistent and monotonic — this part is correct (it's the RMT "latest wins" key).
- **Business dates** use `toDate(<source ts>)` → ClickHouse `Date`. `toDate` truncates a timestamp to a calendar day **in the session/column timezone**. If the ClickHouse session TZ (or the column's TZ) is not pinned to UTC everywhere, the same event can bucket into a **different day** depending on TZ — which makes "same period" queries disagree and compounds the day-boundary edges.
- **Fetch boundary:** the gold views expose `metric_date` as **`toString(<date>)`** — a **String**, not a typed `Date`. The API's OData `$filter` (e.g. `metric_date ge '2026-03-01'`) compares it **lexicographically**. That happens to work for zero-padded `YYYY-MM-DD`, but it's fragile: any format drift (non-padded, different separator, a `DateTime` string) breaks range filters silently, with no type to catch it.

## Risks
1. **TZ-dependent day bucketing** (`toDate` without a pinned UTC session) → same-period totals differ.
2. **String `metric_date` at the API** → relies on exact ISO `YYYY-MM-DD` + lexicographic ordering; no type guard.

## Recommendation (single aligned convention)
- `_version`: keep epoch-millis (`toUnixTimestamp64Milli(now64())`).
- Business dates: compute all `toDate(...)` against a **pinned UTC** session/column TZ; document that business dates are **UTC calendar days**.
- API boundary: treat `metric_date` as **ISO-8601 `YYYY-MM-DD` (UTC)**; ideally type it as `Date` and let the contract enforce it (ties to the dbt/API contract work).

## Tests
- TZ-stability: insert an event near a day boundary, assert it buckets to the expected UTC day regardless of session TZ.
- Format: assert every gold `metric_date` matches `^\d{4}-\d{2}-\d{2}$`; assert an OData range `$filter` returns the same rows as a typed-date range.

Related: contract work (#1323/#1326), #1330 (dedup).

Issue #1330, opened 2026-06-13
data(gold): IC dashboard views read silver ReplacingMergeTree without FINAL → duplicate / unstable metrics (same period, different numbers day-to-day)


SharedQA:
## Symptom (reported)
Fetching metrics for the **same period** on different days returns **different numbers**.

## Root cause
`src/ingestion/scripts/migrations/20260422000000_gold-views.sql` (the IC-dashboard gold views — `*_bullet_rows`, focus-metric rollups) reads silver tables **without `FINAL`**:

- **9 silver reads, 0 with `FINAL`** (e.g. `FROM silver.class_collab_email_activity`, `class_collab_meeting_activity`, `class_focus_metrics`, `class_collab_chat_activity`, `class_collab_document_activity`).

All `silver.class_*` are `ReplacingMergeTree(_version)`. RMT only collapses duplicate `unique_key` rows during **asynchronous background merges** — never guaranteed at query time. A `SELECT` without `FINAL` therefore sees **un-merged duplicates**, which inflate the aggregates; and because merges run on ClickHouse's own schedule, the **same query returns different values on different days** as duplicates get collapsed over time. This is the reported instability.

## Same pattern is already applied correctly elsewhere
`20260512000000_crm-gold-views.sql` does it right — **14/14 silver reads use `FINAL`**, with a header comment: *"Background merges aren't synchronous, so a SELECT without FINAL can see un-merged duplicates. We use FINAL on every silver read here."* The older IC views (20260422) predate that discipline.

## Fix
Add read-time dedup to **every** silver RMT read in the IC gold views — mirror crm-gold-views: hoist each silver source into a single `*_dedup AS (SELECT * FROM silver.class_X FINAL)` CTE referenced by downstream CTEs (so the version-resolve runs once per source).

## Test (QA, deterministic)
1. **Dedup:** seed a silver `class_*` table with two rows for the same `unique_key` differing only by `_version` (the pre-merge duplicate state), query the gold metric, assert the value reflects **one** row (latest `_version`), not the sum of both.
2. **Stability:** query the same metric/period twice without an intervening merge → identical result. (Forcing `OPTIMIZE ... FINAL` between reads must NOT change the value once FINAL is in the view.)

Belongs in the e2e rig as a `*_dedup_smoke` fixture. Related: union_by_tag read-time-dedup convention; #1318/#1319 (data-correctness).

PR #1333, opened 2026-06-13
test(analytics-api): unit tests for pure-logic modules (threshold/query/metric/config)

SharedQA:
Raises analytics-api unit coverage on the **pure-logic** modules (the part that doesn't need a DB/runtime).

## Measured (cargo-llvm-cov, instrumented)
- Workspace: **58.53% → 59.86% lines** (+1.33pts) from these tests. (api-gateway's larger jump is separate — #1327.)
- `threshold.rs` 0% → covered; `query.rs`/`metric.rs`/`config.rs` raised.
- 19 tests; analytics-api **257 pass**; `clippy --all-targets` clean (no unwrap/expect in tests).

## Test validity
Expected values are derived from **documented intent** — OData semantics, the in-code PATCH triple-state comment, the `default_*` fns — **not** from running the code. Negative/boundary cases included (unknown operator, missing required config, null-vs-absent PATCH, empty/oversized inputs).

**One unit was found semantically suspect during review and flagged rather than locked in:** `threshold_matches`'s `eq` uses `f64::EPSILON` (magnitude-dependent equality) — see **#1332**. The test characterises current behaviour; the issue asks for the intended rule.

## Coverage ceiling
Further unit gains are limited: the remaining uncovered analytics-api/oidc/clickhouse code is runtime/IO (`validate_token` needs JWKS, DB clients need ClickHouse/MariaDB, `main`/module wiring need the modkit runtime). Reaching ~80% requires the **instrumented e2e profile** (the rig running instrumented binaries — milestone 2.15), not more unit tests.

Related correctness findings from this pass: #1330 (duplicate/unstable data), #1331 (time formats), #1332 (threshold eq).

## Summary by CodeRabbit

* **Tests**
  * Added unit test coverage for analytics configuration precedence and required-field validation.
  * Added serialization/deserialization tests for metric update PATCH semantics and query/batch request behavior, including defaulting and JSON status formatting.
  * Added threshold evaluation and request deserialization tests covering supported operators, epsilon equality, invalid operator/level handling, and partial updates.


SharedQA:
Overlaps #1314 (migrate to gears) on `domain/query.rs`. Reconcile these unit tests with the gears changes once #1314 merges.

src/backend/services/analytics-api/src/domain/metric.rs
```diff
@@ -80,3 +80,69 @@ pub struct TableColumn {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    type R = Result<(), Box<dyn std::error::Error>>;
+
+    #[test]
+    fn update_description_is_triple_state() -> R {
```
cyberantonz, line 97: Not sure what this test for. Put something to structure and check it is present? Strange idea.

src/backend/services/analytics-api/src/domain/metric.rs
```diff
@@ -80,3 +80,69 @@ pub struct TableColumn {
...
+            id: Uuid::nil(),
+            insight_tenant_id: Uuid::nil(),
+            name: "m".to_owned(),
+            description: None,
+            query_ref: "SELECT 1".to_owned(),
+            is_enabled: true,
+            created_at: ts,
+            updated_at: ts,
+        };
+        let json = serde_json::to_string(&m)?;
```
cyberantonz: What this test for? You do not sure that serde do what it should do? 
I bet the serde developers have such kind of tests in their test cases.

src/backend/services/analytics-api/src/domain/metric.rs
```diff
@@ -80,3 +80,69 @@ pub struct TableColumn {
...
+            !json.contains("description"),
+            "None description omitted: {json}"
+        );
+        assert!(json.contains("\"query_ref\":\"SELECT 1\""));
+        Ok(())
+    }
+
+    #[test]
+    fn metric_summary_never_exposes_query_ref() -> R {
+        let s = MetricSummary {
```
cyberantonz:
```rust
pub struct MetricSummary {
    pub id: Uuid,
    pub name: String,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub description: Option<String>,
}
```

There is no query ref in the structure. What is this test for?

src/backend/services/analytics-api/src/domain/query.rs
```diff
@@ -85,3 +85,73 @@ pub enum BatchQueryResult {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    type R = Result<(), Box<dyn std::error::Error>>;
+
+    #[test]
+    fn default_top_is_25() {
```
cyberantonz: Why this test exists? The function is just return 25 and we need to check it. It is strange

src/backend/services/analytics-api/src/config.rs
```diff
@@ -98,3 +98,57 @@ impl AppConfig {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    type R = Result<(), Box<dyn std::error::Error>>;
+
+    #[test]
+    fn default_helpers() {
+        assert_eq!(default_bind_addr(), "0.0.0.0:8081");
```
cyberantonz: Why we need to test those values at all?

cyberantonz (commented):
Needs work

src/backend/services/analytics-api/src/domain/metric.rs
SharedQA: That one was just round-tripping through serde — it couldn't fail unless serde itself broke. Removed in 27045b4f.

src/backend/services/analytics-api/src/domain/metric.rs
SharedQA: No point re-testing serde. Dropped it. The only metric test I kept is the triple-state one: description is Option<Option<_>> with a custom deserializer (absent / null / value), so that branching is actually ours, not the library's.

src/backend/services/analytics-api/src/domain/metric.rs
SharedQA: Good point — MetricSummary doesn't even have a query_ref field, so the assert was dead weight. Removed.

src/backend/services/analytics-api/src/domain/query.rs
SharedQA: Agreed, asserting default_top() returns 25 proves nothing. Gone. The default still gets exercised through real deserialization in query_request_applies_defaults_when_empty.

src/backend/services/analytics-api/src/config.rs
SharedQA: Same here — that was just testing the defaults equal their literals. Removed default_helpers and the defaults-extraction test; kept only the override-precedence and missing-clickhouse_url cases, which check real behavior.

ktursunov:
Already covered with unit tests coverage

Issue #1322, opened 2026-06-13
data(bronze): source freshness on all sources + contracts on *__bronze_promoted views


SharedQA:
**Layer: BRONZE.** Independent of silver/gold — a dev can take this alone.

Measured by the dbt-coverage gate (`scripts/ci/dbt_coverage.py`, run in CI via the Data Contracts workflow):
- **Source freshness: 0 / 130 sources** declare `loaded_at_field` + `freshness`. Add freshness to every bronze source so a stalled connector is detected, not silently served.
- **`*__bronze_promoted` views: 0 contracted.** Add `contract: { enforced: true }` + column `data_type`s + `not_null`/`unique` on `unique_key` (the first typed boundary).

Done when: every bronze source has a freshness block; every `*__bronze_promoted` model is contracted + key-tested; `dbt_coverage.py` shows bronze at 100%.

**Not blocked by the gears-rust migration.** This is plain dbt/OpenAPI work; it does not require adopting any gears toolkit/lints. Gears adoption is a separate track.

Gate + baseline: #1320. Related: #1318, #1319.

Issue #1321, opened 2026-06-13
EPIC: data-correctness contract coverage (bronze · silver · gold) → 100%


SharedQA:
Umbrella for getting the metric path fully under contract + tested + monitored. **Split per layer so devs work each independently**; the CI wire-up is its own task. **None of this depends on the gears-rust migration** (plain dbt/OpenAPI; gears is a separate track).

QA owns the measurement (the `dbt_coverage.py` gate + Data Contracts CI workflow, #1320); Dev owns filling each layer.

## Children
- [ ] #1322 — **bronze**: source freshness (0/130) + `*__bronze_promoted` contracts
- [ ] #1323 — **silver**: enforced contracts + key tests (baseline 1/124, 45/124 key-tested)
- [ ] #1324 — **gold (API)**: API contract (openapi+schemathesis+oasdiff) + golden-value pins (gold = views, not dbt)
- [ ] #1326 — **gold (UI render)**: displayed value = documented transform of the API value (rounding/format) — `DOM == format(api)`, not `DOM == api`
- [ ] #1325 — **CI wire-up**: flip the gates to blocking per layer (ratchet)

## Baseline (measured in CI, no DB needed — `dbt parse`)
contracts **1/124** · key tests **45/124 (36%)** · source freshness **0/130**.

Related defects this prevents recurring: #1318 (type drift), #1319 (wired-empty connectors).

SharedQA:
## Layer-by-layer test/gate implementation plan & status

Tracking the medallion-layer data-quality gates here. ✅ shipped (PR) · 🟡 partial/in-progress · ⬜ planned.

**Bronze**
- ✅ Data presence / volume audit — #1335 (`data_presence_audit.py`)
- ✅ Freshness SLA (warn 36h / error 72h) — #1346 (dbt source freshness, 21 connectors)
- ⬜ Schema contracts on `*__bronze_promoted` — #1322 (needs warehouse CI)
- ⬜ Volume/anomaly baseline (today 0 vs usual 1000) — extend #1335

**Silver**
- ✅ Connector→silver coverage gate — #1335 (`connector_silver_coverage.py`)
- ✅ Nullable dedup-key guard (`order_by` NULL→double-count) — #1348 (`nullable_key_audit.py`)
- ✅ Uniqueness / not-null on keys — existing dbt (429 unique / 474 not_null)
- ⬜ Referential integrity (`relationships`, e.g. `person_key`→`class_people`) — #1323
[...]

Issue #1319, opened 2026-06-13
data: wired-but-empty connectors (Slack, Zoom, Zulip-Proxy) break silver build and yield no metrics — seed or deregister


SharedQA:
## Summary
Several connectors are wired into the pipeline but carry no source data. Their staging snapshots are therefore skipped, and downstream silver models fail or no-op. Observed in the `dbt-sim-silver` run as the cascade `Unknown table expression identifier 'staging.<conn>__users_snapshot'` for `slack__users_fields_history`, `zoom__users_fields_history`, `zulip_proxy__users_fields_history`, plus the `union_by_tag: no source tables` first-run errors for ai/crm/git `class_*` targets that have no contributing connector.

## Why it matters
- **No metrics** are produced for those sources (the 5 wired-empty connectors are invisible, not erroring loudly in production).
- The build noise (30+ red models) **masks real defects** — e.g. #1318 was buried among them.
- Per the data-correctness contract, a source silently producing nothing makes a contributor's activity look lower than it actually is.

## Decision needed (Product)
For each wired-empty connector, choose one and record it:
1. **Seed** real data via a sandbox/service account (see QA Requirements → integration sandboxes), so the snapshot + silver build, or
2. **Deregister**: remove the connector's models from the run / mark out of scope, so the pipeline doesn't carry dead models.

## Engineering follow-ups (regardless of the above)
- `union_by_tag` should emit an empty typed relation on a first run with zero sources instead of hard-erroring, **or** the sim/run should `--select`-exclude silver targets with no configured connector — so a partial environment is green and real failures stand out.
- Zero-row / freshness alerts in the production sync DAG (30-day P0-3) so an empty connector pages instead of failing silently.

Issue #949, opened 2026-06-05
[PR #256] feat(ingestion): bronze freshness SLA + daily freshness CronWorkflow


il10241024:
> 🔗 **Mirrored PR** [cyberfabric/cyber-insight#256](https://github.com/cyberfabric/cyber-insight/pull/256) | **Author:** dzarlax | **Opened:** 2026-05-04T08:37:13Z | **Status:** open
> *GitHub API does not allow setting PR author or timestamps — attribution preserved here.*

---

## Summary

Detect "data hasn't arrived for several days" automatically, without per-connector boilerplate.

- **Single source of truth** in `src/ingestion/dbt/dbt_project.yml`: every bronze source inherits `loaded_at_field: _airbyte_extracted_at` + `warn_after: 30h` / `error_after: 48h`. New connectors are covered the moment they declare a `bronze_*` source — nothing to copy-paste.
- **Per-table opt-outs** only for slow-moving lookup/catalog streams (`jira_statuses`, `claude_admin_workspaces`, `bamboohr.meta_fields`, `confluence.wiki_spaces`, etc.) where a quiet day is legitimate. Lives next to the table in `schema.yml`.
- **One generic CronWorkflow** `dbt-source-freshness-check` runs `dbt source freshness --select source:*` daily at 13:00 UTC (after every connector's sync window of 02:00–11:00 UTC). Parses `target/sources.json`, logs each breaching source with `max_loaded_at` + lag, and POSTs a generic JSON payload to `ingestion.freshness.notificationWebhookUrl`.

## Exit semantics

| Status | Meaning | Workflow exit | What to do |
|---|---|---|---|
| `pass` | within SLA | 0 | nothing |
| `warn` | 30–48h lag (one missed run) | 0 (visible in log + payload) | investigate during business hours |
| `error` | >48h lag | 1 (Argo Failed) | page |
| `runtime error` | dbt couldn't even check (CH down, schema drift) | 1 (Argo Failed) | page |

`error` and `runtime error` flip the workflow to Failed so Argo retains the run in `failedJobsHistoryLimit`. Warn-only runs stay Successful — the breach is still printed and POSTed in the payload, but on-call doesn't get paged on a single missed sync.

## Notification payload

Provider-agnostic JSON — Zulip / Mattermost / generic webhook / a small relay all fit:

```json
{
  "topic": "ingestion-freshness",
  "cluster": "prod",
  "summary": "[prod] 3 bronze source(s) breaching freshness SLA",
[...]

PR #1346, opened 2026-06-15
test(data): source freshness on all bronze connectors (#1322)

SharedQA:
## What
Adds dbt **source freshness** to all 21 `bronze_*` connector sources — `warn_after: 36h`, `error_after: 72h`, `loaded_at_field: _airbyte_extracted_at`.

## Why
A stopped/slow sync is currently invisible: the dashboard just shows stale numbers with no error. `dbt source freshness` turns that into a signal. Implements the bronze half of [#1322](https://github.com/constructorfabric/insight/issues/1322) (EPIC [#1321](https://github.com/constructorfabric/insight/issues/1321)).

## Notes
- Freshness is its own command (`dbt source freshness`), not part of `dbt build`/`test` — so sources not yet ingested (#1319) don't break the normal run.
- Verified `_airbyte_extracted_at` exists on the ingested bronze tables.
- **Next in this branch:** enforced contracts (`data_type`) on silver models — those need the warehouse-gated CI dbt run to validate the type declarations, so they land once `data-checks.yml` runs against a populated ClickHouse (can't be validated on a machine without dbt).

Draft until the freshness thresholds are confirmed with the team.

## Summary by CodeRabbit

* **New Features**
  * Enabled DBT data freshness monitoring across 20+ connectors, so stale or missing upstream data now surfaces as freshness warnings/errors within configured time windows.
  * Added source-level defaults (based on ingestion/extraction time) and table-level overrides for better alignment to the most relevant business or event timestamps, with tuned thresholds (commonly ~36h/72h, and stricter settings like ~72–168h for specific datasets).

### DIFF
--- a/src/ingestion/connectors/ai/chatgpt-team/dbt/schema.yml
+++ b/src/ingestion/connectors/ai/chatgpt-team/dbt/schema.yml
@@ -2,6 +2,12 @@ version: 2
 
 sources:
   - name: bronze_chatgpt_team
+    # Source freshness (TAF #1322): the newest Airbyte row must be recent,
+    # else the sync silently stopped and dashboards go stale with no error.
+    freshness:
+      warn_after: {count: 36, period: hour}
+      error_after: {count: 72, period: hour}
+    loaded_at_field: _airbyte_extracted_at
     schema: bronze_chatgpt_team
     tables:
       - name: chatgpt_team_seats
--- a/src/ingestion/connectors/ai/chatgpt-team/dbt/schema.yml
+++ b/src/ingestion/connectors/ai/chatgpt-team/dbt/schema.yml
@@ -2,19 +2,39 @@ version: 2
 
 sources:
   - name: bronze_chatgpt_team
-    # Source freshness (TAF #1322): the newest Airbyte row must be recent,
-    # else the sync silently stopped and dashboards go stale with no error.
+    # Source freshness (TAF #1322). Source-level default = _airbyte_extracted_at
+    # (sync-liveness) for the roster table; the WINDOWED daily/snapshot tables
+    # override below to anchor on their business date. ChatGPT Team re-emits a
+    # fixed window every sync, so extracted_at is false-green for them — VERIFIED
+    # on live data (extracted_at ~8.6h fresh while `date` 60–85h stale). Report
+    # tier absorbs the vendor publish lag.
     freshness:
       warn_after: {count: 36, period: hour}
       error_after: {count: 72, period: hour}
     loaded_at_field: _airbyte_extracted_at
     schema: bronze_chatgpt_team
     tables:
-      - name: chatgpt_team_seats
-      - name: chatgpt_team_chat_activity
-      - name: chatgpt_team_codex_user_daily
-      - name: chatgpt_team_subscription_usage
-      - name: chatgpt_team_subscription_balance
+      - name: chatgpt_team_seats   # roster — sync-liveness default is correct
+      - name: chatgpt_team_chat_activity      # VERIFIED windowed
+        loaded_at_field: parseDateTimeBestEffortOrNull(date)
+        freshness:
+          warn_after: {count: 72, period: hour}
+          error_after: {count: 120, period: hour}
+      - name: chatgpt_team_codex_user_daily   # VERIFIED windowed
+        loaded_at_field: parseDateTimeBestEffortOrNull(date)
+        freshness:
+          warn_after: {count: 72, period: hour}
+          error_after: {count: 120, period: hour}
+      - name: chatgpt_team_subscription_usage    # inferred windowed (snapshot_date)
+        loaded_at_field: parseDateTimeBestEffortOrNull(snapshot_date)
+        freshness:
+          warn_after: {count: 72, period: hour}
+          error_after: {count: 120, period: hour}
+      - name: chatgpt_team_subscription_balance  # inferred windowed (snapshot_date)
+        loaded_at_field: parseDateTimeBestEffortOrNull(snapshot_date)
+        freshness:
+          warn_after: {count: 72, period: hour}
+          error_after: {count: 120, period: hour}
 
 models:
   - name: chatgpt_team__ai_dev_usage
--- a/src/ingestion/connectors/ai/claude-admin/dbt/schema.yml
+++ b/src/ingestion/connectors/ai/claude-admin/dbt/schema.yml
@@ -2,6 +2,12 @@ version: 2
 
 sources:
   - name: bronze_claude_admin
+    # Source freshness (TAF #1322): the newest Airbyte row must be recent,
+    # else the sync silently stopped and dashboards go stale with no error.
+    freshness:
+      warn_after: {count: 36, period: hour}
+      error_after: {count: 72, period: hour}
+    loaded_at_field: _airbyte_extracted_at
     schema: bronze_claude_admin
     tables:
       - name: claude_admin_users
--- a/src/ingestion/connectors/ai/claude-admin/dbt/schema.yml
+++ b/src/ingestion/connectors/ai/claude-admin/dbt/schema.yml
@@ -2,8 +2,12 @@ version: 2
 
 sources:
   - name: bronze_claude_admin
-    # Source freshness (TAF #1322): the newest Airbyte row must be recent,
-    # else the sync silently stopped and dashboards go stale with no error.
+    # Source freshness (TAF #1322). INFERRED, NOT live-verified — bronze_claude_admin
+    # has no rows in any reachable install (config is not a reliable signal, cf.
+    # M365). Source-level default = _airbyte_extracted_at (sync-liveness) for the
+    # roster/lookup tables (full-refresh → extracted_at stays fresh, so it still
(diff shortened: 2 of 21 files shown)


dzarlax:
**Coordinating — I have a ready "runtime/alerting" layer that builds directly on this PR.**

This PR nails the *declaration* half of #1322 (per-source `freshness` blocks so `dbt_coverage.py` counts them toward 130/130). I've got the complementary *runtime* half already written against the old tree, and it sits cleanly on top of yours rather than competing with it:

- `dbt-source-freshness-check` **CronWorkflow** — runs `dbt source freshness --select source:*` daily after the sync window, parses `target/sources.json`, logs each breaching source with `max_loaded_at` + lag, and flips the workflow to Failed only on `error`/`runtime error` (warn stays green, still printed + emitted in payload).
- **Driver-based notification** (webhook / zulip / slack / teams) off a provider-agnostic JSON payload.
- **Runtime trap-detector** (`freshness-trap-detect.py`) for the "dbt couldn't even check" case (CH down, schema drift).
- PRD/DESIGN/FEATURE for a new `ingestion-monitoring` domain.

None of that exists in #1322's scope (which is declaration + coverage gate), so it's additive. Two things to align:

1. **Thresholds** — I had 30h/48h, you have 36h/72h. Happy to defer to yours; the cron reads whatever the sources declare, so no duplication.
2. **Sequencing** — my cron consumes the per-source declarations this PR adds, so it should land *after* this merges. I'd open it as a follow-up PR referencing #1321 as the runtime/alerting half of #1322.

One design note for the record: my original took the inherited-default route (`loaded_at_field` + thresholds in `dbt_project.yml`, per-table opt-outs). That's functionally equivalent but **wouldn't satisfy `dbt_coverage.py`**, which counts per-source declarations — so your per-source approach is the right call and I'll build on it as-is.

Want me to open the follow-up PR once this merges, or fold the cron into this branch?

SharedQA:
Thanks — agreed on all of it, and good that the runtime half is already written.

On the two alignment points:

1. **Thresholds — keep 36h/72h here.** The bronze syncs are daily, so 36h warn is roughly a 1.5-day grace (one missed sync isn't yet alarming) and 72h error covers a weekend gap before anyone gets paged. Since your cron reads whatever the sources declare, there's a single source of truth and no duplication — if we want to tighten later, it changes in one place.

2. **Sequencing — open it as a separate follow-up after this merges, not folded in.** This PR is deliberately declaration-only: static, no warehouse needed, mergeable as-is. Your runtime layer is a different concern with its own domain (`ingestion-monitoring`) and its own PRD/DESIGN, so keeping them apart keeps both reviewable and lets this gate land without waiting on the alerting design. The dependency direction is right too — your cron consumes these per-source blocks, so this lands first and yours sits on top.

For tracking: the runtime/alerting CronWorkflow is the "scheduled bronze freshness" item in the stabilization plan (the monitoring line, #949), so it slots in there as the operational counterpart to this declaration gate.

And confirmed your read on `dbt_coverage.py` — it counts per-source declarations, which is exactly why we went per-source instead of the `dbt_project.yml` inherited default. Please go ahead and open the follow-up once this merges.

dzarlax:
**Heads-up before this lands: flat `_airbyte_extracted_at` is false-green on the windowed/vendor-analytics connectors — verified on live data (Eternal installation).**

The check correctly catches "sync fully stopped" for true incremental connectors. But for any connector whose API returns a **fixed reporting window** that the sync re-fetches every run (M365 Graph reports, ChatGPT Team, Claude Team), `_airbyte_extracted_at` is re-stamped on every successful sync regardless of whether the business data advanced — so it measures *sync liveness*, not *data freshness*.

Measured just now on the Eternal ClickHouse (`now() - max(...)`):

| bronze source | `max(_airbyte_extracted_at)` age | `max(<business_date>)` age | distinct business dates re-emitted in the last extraction batch |
|---|--:|--:|--:|
| `m365.{email,teams,onedrive,sharepoint}_activity` (`reportRefreshDate`) | **10.6 h** | **108.7 h** | 24 |
| `chatgpt_team.chatgpt_team_chat_activity` (`date`) | 8.6 h | 84.7 h | 3 |
| `chatgpt_team.chatgpt_team_codex_user_daily` (`date`) | 8.6 h | 60.7 h | 4 |
| `claude_team.claude_team_code_metrics` (`metric_date`) | 8.6 h | 60.7 h | 6 |
| `youtrack.youtrack_issue` (incremental, contrast) | 9.4 h | 13.2 h | — |

Right now M365 business data is **4.5 days stale** on this install, but `warn_after: 36h / error_after: 72h` on `_airbyte_extracted_at` reports **green**. The incremental connectors (youtrack/jira) are fine — `ext_age ≈ biz_age` — which is exactly the discriminator.

This matters because the false-green sources are the **AI-usage connectors** that Insight's value story leans on. The connector config is misleading here too — M365 declares `lookback_window: P0D`, but the Graph `getReport` endpoint returns the whole `reportPeriod` window, so the data re-fetches regardless.

**Proposal: adopt the per-source model from the earlier #256 work as the declaration here — it's demonstrably the correct instrument on exactly these connectors.** Concretely, additive to this PR's per-source structure:

1. **Anchor `loaded_at_field` on the business-date column** for the windowed connectors instead of `_airbyte_extracted_at` — `reportRefreshDate` (M365), `date` (ChatGPT / Slack), `metric_date` (Claude). Incremental connectors (git, jira, youtrack) keep `_airbyte_extracted_at` as-is.
2. **Tier the thresholds** so the windowed connectors get a higher `warn_after`/`error_after` that absorbs the legitimate 2–3 day vendor publish lag — otherwise anchoring on the business date would false-*alarm*. (This is why #256 carried tiers; the dogfood numbers above show the lag is real, not a bug.)
3. Per-table opt-outs (`freshness: null` + a written reason) for the slow-moving lookup/catalog streams.

I have the full per-connector `loaded_at_field` + tier mapping from #256 ready to go. Happy to push it straight onto this branch so the declaration half ships correct in one piece, or land it as the immediate follow-up the runtime PR sits on — whichever you prefer for review flow.

SharedQA:
This is the right catch, and it changes my call from earlier in the thread — thank you for dogfooding it on live data.

You're right that `_airbyte_extracted_at` measures sync liveness, not data freshness, and that it goes false-green on exactly the windowed vendor-analytics connectors (M365 Graph, ChatGPT Team, Claude Team) where the API re-emits a fixed reporting window every run. I corroborated the anchors you propose are real columns on those connectors (`reportRefreshDate` on M365, `metric_date` on Claude, `date` on the ChatGPT/Slack streams), so the per-source anchoring is feasible as a drop-in.

So I'm reversing my "keep it flat, declaration-only is fine as-is" position: **the per-source + tiered model belongs in this PR**, not a follow-up. A freshness gate that reads green while M365 business data is 4.5 days stale is worse than no gate — and it's worst on the AI-usage connectors that matter most. My earlier 36h/72h reasoning only held for the true-incremental connectors (`ext_age ≈ biz_age`); it's wrong for the windowed ones.

Please **push the #256 per-connector mapping straight onto this branch** (`claude/dbt-test-coverage`) — you have the live-verified `loaded_at_field` + tier values and I'd only be guessing the per-stream business-date columns and the lag-absorbing thresholds, which is exactly where a wrong guess turns into false-*alarm*. Concretely, to land the declaration correct in one piece:

1. Anchor `loaded_at_field` on the business-date column for the windowed connectors; keep `_airbyte_extracted_at` for the incrementals (git/jira/youtrack) — `ext_age ≈ biz_age` is the clean discriminator.
2. Tiered `warn_after`/`error_after` so the windowed connectors absorb the real 2–3 day vendor publish lag (your numbers show it's genuine).
3. Per-table `freshness: null` + a one-line reason for the slow-moving lookup/catalog streams.

The runtime/alerting CronWorkflow stays the separate follow-up we discussed — this keeps that PR sitting on a declaration that's actually correct. If pushing to the branch is awkward from your side, open it as a PR into `claude/dbt-test-coverage` and I'll merge it in.

dzarlax:
Opened it as a PR into this branch: SharedQA/insight#1 — anchors the windowed connectors (m365 / chatgpt_team / claude_team verified, the rest column-verified or inferred-and-marked) on their business-date column with tiered thresholds; incrementals untouched. The PR body breaks down the three confidence tiers and the one caveat (no end-to-end `dbt source freshness` run on the full set yet). Ready for you to merge in whenever it looks right.

dzarlax:
One more before this lands — reviewing the merged anchors I caught a latent bug and opened SharedQA/insight#2 against this branch:

- **openai**: `bucket_start_time` is Unix seconds (integer), not a date string — the connector's own models use `fromUnixTimestamp(CAST(... AS UInt32))`. The anchor I added in #1 used `parseDateTimeBestEffortOrNull`, which would `runtime error` on every openai usage/cost table once it's ingested. #2 fixes it to the epoch form.
- **confluence/outline**: aligned `wiki_pages` to `parseDateTime64BestEffortOrNull(updated_at, 3)` to match each connector's own model convention (functionally equivalent, just consistent).

openai is empty in the dogfood install so it wouldn't fail today, but it's a latent false-positive. Worth folding #2 in before #1346 merges so the gate ships correct. All other inferred anchors were re-checked against the connectors' silver models and parse fine.

dzarlax (approved):

PR #1348, opened 2026-06-15
test(data): static gate — no dbt model may dedup on a nullable key (#1330)

SharedQA:
## What
`scripts/ci/nullable_key_audit.py` — a static (warehouse-free) gate that fails CI if any dbt model's ReplacingMergeTree dedup key (`order_by`) is projected as `Nullable(...)` / `CAST(NULL …)`.

## Why
A NULL in the sorting key never compares equal to itself, so RMT can't collapse those rows — they accumulate as silent duplicates and the metric drifts as background merges run (the "same period, different number" class, [#1330](https://github.com/constructorfabric/insight/issues/1330)). `allow_nullable_key: 1` is set in ~99 models (it's needed just to *create* a table whose ORDER BY has a nullable column), so ClickHouse permits it — this gate is the backstop.

## How
Paren-depth-aware backward scan isolates the key's own projection, so `MD5(concat(a, b, c))` inner commas and a sibling column's `CAST(NULL …)` don't cause false positives (both were real cases — see the parser self-test). Blocking from day one in `data-integrity.yml`; no service container needed.

## Result today
80 ordered/RMT models scanned, **0** nullable dedup keys — green. Keys are non-null by construction (MD5 / CAST AS String / coalesce); this locks that invariant so a future model can't quietly reintroduce the duplicate-metric bug. Parser self-test: 4 cases incl. the MD5-inner-comma and sibling-CAST(NULL) false-positive guards.

This is the **write/schema** complement to the `not_null` read tests and the enforced-contract work on #1346.

## Summary by CodeRabbit

* **Chores**
  * Added automated data integrity validation in CI to prevent dbt models from using nullable columns for deduplication.

* **Tests**
  * Added comprehensive tests for the data integrity validation system.


.github/workflows/data-integrity.yml
```diff
@@ -0,0 +1,26 @@
...
+
+# Static, warehouse-free guards on the dbt models. Blocking from day one — no
+# service container needed, so there's no reason to let these regress.
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - 'src/ingestion/**'
+      - 'scripts/ci/nullable_key_audit.py'
+  workflow_dispatch:
```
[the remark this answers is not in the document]
SharedQA: Fixed — added `scripts/ci/test_nullable_key_audit.py` and `.github/workflows/data-integrity.yml` to the `paths` filter, so changes to the guard's test or the workflow itself now trigger it.

.github/workflows/data-integrity.yml
```diff
@@ -0,0 +1,26 @@
...
+
+jobs:
+  nullable-key-guard:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
```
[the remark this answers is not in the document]
SharedQA: Fixed — SHA-pinned `actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5` (v4) and `actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065` (v5), and added `persist-credentials: false` to checkout (this job only reads the tree).

scripts/ci/nullable_key_audit.py
```diff
@@ -0,0 +1,119 @@
...
+    sibling columns by a paren-depth-aware backward scan.
+
+    Walking left from `AS <col>`, commas *inside* parentheses (e.g. the args of
+    `MD5(concat(a, b, c))`) sit at depth > 0 and are skipped; the expression
+    starts at the first comma seen at depth 0 (the boundary with the previous
+    column). This is what makes `CAST('' AS String) AS unique_key` read as just
+    that cast and not pick up a neighbour's `CAST(NULL …)`.
+    """
+    out: list[str] = []
+    for m in re.finditer(rf"\bAS\s+{re.escape(col)}\b", text, re.IGNORECASE):
```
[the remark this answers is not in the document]
SharedQA: Skipping — your own scan confirmed zero quoted aliases (`AS "col"` / backtick) anywhere in `src/ingestion`; the models use unquoted identifiers. Leaving the regex as-is to keep it readable; will add the alternation if a model ever adopts quoted aliases.

scripts/ci/nullable_key_audit.py
```diff
@@ -0,0 +1,119 @@
...
+    for col in cols:
+        if col.startswith("_"):
+            continue  # housekeeping cols (e.g. _tracked_at) aren't the identity key
+        for proj in projections_for(text, col):
+            if NULLABLE_RE.search(proj) or CAST_NULL_RE.search(proj):
+                out.append(
+                    ("ERROR", col, "dedup key projection is Nullable / CAST(NULL …)")
+                )
+                break
+    return out
```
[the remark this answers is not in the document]
SharedQA: Adopted, with a guard against false positives. It now fails closed (`unable to resolve dedup key projection`) when a key is neither projected locally nor passed through — but a blanket fail-closed flagged 49 legitimate models, so it treats these as inherited (key audited upstream): `SELECT *`, `union_by_tag` unions, bare/qualified passthroughs (`unique_key,` / `c.unique_key,`), and macro-generated keys (`{{ snapshot(unique_key_col='unique_key') }}`). Added self-tests for the fail-closed path and each inherited pattern. Gate stays green: 80 models, 0 nullable keys.

scripts/ci/nullable_key_audit.py
```diff
@@ -0,0 +1,150 @@
...
+  nullable_key_audit.py --check    # exit 1 if any model dedups on a nullable key
+"""
+from __future__ import annotations
+
+import argparse
+import pathlib
+import re
+import sys
+
+ROOT = pathlib.Path(__file__).resolve().parents[2]
```
cyberantonz, line 32: It is very weird construction. Claude likes it.

scripts/ci/nullable_key_audit.py
```diff
@@ -0,0 +1,150 @@
...
+INGEST = ROOT / "src" / "ingestion"
+
+ORDER_BY_RE = re.compile(r"order_by\s*=\s*\[([^\]]*)\]")
+NULLABLE_RE = re.compile(r"Nullable\s*\(", re.IGNORECASE)
+CAST_NULL_RE = re.compile(r"CAST\s*\(\s*NULL\b", re.IGNORECASE)
+def projections_for(text: str, col: str) -> list[str]:
```
cyberantonz, line 40: You mention linters and formatters in PR template. So, can you use ruff to format this code and mypy to statically check it?

scripts/ci/nullable_key_audit.py
```diff
@@ -0,0 +1,150 @@
...
+                out.append(
+                    ("ERROR", col, "dedup key projection is Nullable / CAST(NULL …)")
+                )
+                break
+    return out
+
+
+def main() -> None:
+    ap = argparse.ArgumentParser()
+    ap.add_argument("--check", action="store_true", help="exit 1 on a nullable dedup key")
```
cyberantonz: Why we need this flag? From my point of view it is useless and always should be set to true. When can we need to have report only?

scripts/ci/nullable_key_audit.py
```diff
@@ -0,0 +1,150 @@
...
+        for _sev, col, note in audit_model(path):
+            rel = path.relative_to(ROOT)
+            errors.append((rel, col, note))
+            print(f"  ERROR {rel} [{col}]: {note}")
+
+    print(
+        f"\nsummary: {ordered} ordered/RMT models scanned · "
+        f"{len(errors)} dedup on a NULLABLE key"
+    )
+    if args.check and errors:
```
cyberantonz:
```python
if not args.check:
    return

if errors:
    print(f"\nFAILED: {len(errors)} model(s) dedup on a nullable key")
    sys.exit(1)

print("\nPASSED")
```

cyberantonz (commented):
Need work

scripts/ci/nullable_key_audit.py
SharedQA: It just resolves the repo root: the file is `scripts/ci/nullable_key_audit.py`, so `parents[2]` walks up two levels (ci → scripts → root). I've added a comment on that line spelling it out. Happy to switch to a named search-upward if you'd prefer it explicit.

scripts/ci/nullable_key_audit.py
SharedQA: Done. Ran `ruff format` on both scripts and made the audit `mypy`-clean. I also added a step to `data-integrity.yml` that runs `ruff check` + `ruff format --check` + `mypy` (pinned versions), so it stays enforced rather than being a one-off.

scripts/ci/nullable_key_audit.py
SharedQA: Agreed — removed it. A CI guard has no use for report-only, so the script now always enforces: it exits 1 on any violation, 0 otherwise. The workflow invokes it with no argument.

scripts/ci/nullable_key_audit.py
SharedQA: Adopted this shape. `main()` now returns an int and exits 1 when there are errors (0 otherwise); the `if not args.check` early-return is gone since the flag was removed. `__main__` calls `sys.exit(main())`.

SharedQA:
Two separate things are bundled in the question — the static guard in this PR, and the data-quality tests that genuinely need a warehouse. They need different answers.

## Why this one is a script, not a data test

`nullable_key_audit.py` is a **static analyzer over the model SQL**, not a data-quality test. The defect it prevents — a `Nullable(...)` column in a ReplacingMergeTree `order_by`, so duplicates never collapse (NULL != NULL) and the metric drifts — is a property of the model's definition. It's visible in the `.sql` with **no data and no warehouse**. So it runs in CI in seconds with zero infrastructure.

And it *is* tested: `test_nullable_key_audit.py` is a pytest suite that proves the parser flags the nullable key, ignores commas inside `MD5(concat(...))`, isn't tainted by a sibling `CAST(NULL …)`, and fails closed on an unresolvable key. So it's "a script **with** tests" — which is the right tool for a structural invariant. Turning it into a dbt test would mean standing up a warehouse to check something that's readable straight off the file.

## What the data-quality checks need (the ones you mean — non-negative, spike, caps, bounds, freshness)

These do need a harness — a real ClickHouse, dbt, and seeded data, exactly as you said. Today they run only **post-deploy**, on the Argo `data-quality` CronWorkflow (daily 06:00 UTC) against the deployed warehouse. To run them on a PR we need warehouse-in-CI. The plan:

1. **Reuse the existing rig, don't build a new one.** The bronze→API e2e harness (`src/ingestion/tests/e2e/`, `./e2e.sh`) already stands up ClickHouse + MariaDB + dbt + analytics-api in docker compose with seeded bronze fixtures. That's most of what's needed.
2. **Add a fixture pair per check** — a known-good dataset that passes and a known-bad dataset that *trips* the check. This is the genuinely missing part: without a violating fixture, a check can silently never fire and nobody would notice. (The static guard already has this in its pytest; the catalog checks don't yet.)
3. **Add a runner step** that, after `dbt build` on the seeded data, runs `dbt test --selector data_quality` and asserts the bad fixtures produce the expected findings and the good ones produce none.
4. **Wire it as a CI job** on `src/ingestion/**`, next to the existing e2e gate.

This maps to the warehouse-in-CI items already on the QA plan (ClickHouse + dbt in CI; seeded test tenant + fixtures). Until it lands, the catalog's coverage is the nightly cron, and the static guards like this one are the part we can already gate on a PR with no env.

Short version: this PR's guard is already fully testable and tested with no harness — that's why it's a script. The harness work is for the data tests, and the fastest route is extending the e2e rig with violating fixtures, not standing up something new.

SharedQA:
Closing as absorbed — the nullable-key dedup guard (`nullable_key_audit.py` + its self-test + the ruff/mypy lint + the blocking job) now lives in the consolidated `data-quality.yml` on #1335, alongside connector-silver-coverage and dbt-coverage. This removes the separate `data-integrity.yml` workflow so an ingestion PR triggers one data-quality workflow instead of three. No logic lost; same gate, same SHA-pinned hardening.

Issue #1323, opened 2026-06-13
data(silver): enforced dbt contracts + key tests on all silver metric models


SharedQA:
**Layer: SILVER.** Independent of bronze/gold.

- Add `contract: { enforced: true }` + per-column `data_type` (+ `on_schema_change: fail` for incrementals) to every `class_*`/`mtr_*`/`fct_*` model and the `silver:*` tag-union feeders.
- Add `not_null`+`unique` on each model's `unique_key`.
- Use the live ClickHouse catalog / `dbt docs generate` for exact types.

Baseline (dbt-coverage gate): **1 / 124 metric-path models contracted, 45/124 key-tested**. One done — `class_collab_document_activity` (#1320). Burn down **per domain**: collab → crm → git → ai → hr → wiki → task → identity.

Done when: 124/124 metric-path models contracted + key-tested; `dbt_coverage.py` shows silver at 100%.

**Not blocked by the gears-rust migration.** This is plain dbt/OpenAPI work; it does not require adopting any gears toolkit/lints. Gears adoption is a separate track.

Gate + first contract: #1320. Related: #1318.

PR #1350, opened 2026-06-15
test(data): silver business-rule data tests — non-negative counts (#1321)

SharedQA:
## What
First **custom singular data test** for the silver layer (the feedback's section 4): `assert_collab_document_counts_non_negative.sql` — activity counts on `class_collab_document_activity` can never be < 0. dbt fails the build if any row violates it; honest NULLs (`NULL < 0` → NULL) are not flagged.

## Why
A negative count = a broken transform or bad source row, and it silently corrupts any metric that sums it. This is the verifiable, no-warehouse-needed slice of the silver-integrity work under [EPIC #1321](https://github.com/constructorfabric/insight/issues/1321) / [#1323](https://github.com/constructorfabric/insight/issues/1323).

## Scope / next
- **Now:** the pattern, on the collab document model (columns confirmed).
- **Rollout:** same non-negative + date-sanity + `accepted_values` tests across the other `class_*` tables.
- **Deferred (needs warehouse-gated CI):** enforced `contract: {enforced: true}` with `data_type` — the union models' exact ClickHouse types can't be validated without a `dbt build`, so they land once `data-checks.yml` runs against a populated ClickHouse. Authoring them blind would risk a red build on a wrong type.

Draft until the dbt CI run validates it (dbt isn't runnable in this environment).

## Summary by CodeRabbit

* **Tests**
  * Added a new data-quality validation for collaboration document activity metrics. The test monitors document viewing, editing, syncing, sharing, and page visit counts to ensure they remain non-negative. When potential data anomalies are detected, detailed failure information is automatically logged for investigation, enhancing overall data integrity and system reliability across collaboration features.


cyberantonz (approved):

ktursunov (approved):

SharedQA:
Two things changed in the last commit, and the reasoning behind each:

**1. Wrapped the test in a `config` block so a failure is operationally useful, not just noise**

- `store_failures=true` — when a negative count turns up, dbt writes the offending rows into an audit table instead of only logging an error. You can then `SELECT` those rows and trace each one back to the feeder mapping, rather than re-running to find them.
- `severity='warn'` — the check warns instead of hard-failing the build, so a single bad row doesn't halt the pipeline and starve downstream dashboards, while still surfacing as a finding for the team.
- `meta` (title / domain / category / tier / remediation) — so when it fires, the alert carries the fix instructions with it; whoever responds doesn't have to guess what the test meant or how to act on it.

This also matches the repo's documented data-quality convention (`tests/README.md`) rather than being a one-off blocking test.

**2. Read the table `FINAL` (ClickHouse specific)**

`class_collab_document_activity` is a `ReplacingMergeTree`. On update, ClickHouse keeps the old row and merges away the duplicate in the background later. Without `FINAL`, a plain read can briefly see both the old (broken) row and the new (fixed) row before that merge runs, which would report a phantom failure. `FINAL` resolves the duplicates at query time, so the test only ever evaluates the latest true state of each row. It's also the convention for reading RMT tables in this repo.

SharedQA:
### Review feedback — disposition

Thanks for the thorough review. Triaged the three points; one applied, two intentionally not:

**1. Add `date` to the `SELECT` — ✅ applied** (a803a97c)
Included `date` in the stored-failure rows so on-call gets the activity day of the offending count directly in the audit schema, no second lookup. Kept as **triage context only, not a filter**.
Note: the cited *performance* rationale (partition pruning) doesn't apply to this table — `class_collab_document_activity` is `ReplacingMergeTree(_version)` ordered by `unique_key` with **no `partition_by`**, so a `unique_key` lookup already hits the sort key. The value here is human triage, not planner efficiency.

**2. Window `FINAL` to the last 7 days — ⛔ intentionally not done**
This is a hard physical-bound invariant (`tier=error`): a negative count is corruption *wherever* it lands, including older partitions rewritten by a late backfill or a historical transform bug. A `date >= today() - 7` window would deliberately blind the check to exactly those cases. Completeness wins over speculative `FINAL` cost on a correctness invariant. If `FINAL` ever proves measurably expensive, the correct bound is **arrival time** (`collected_at` / `_version`), not activity `date` — bounding by activity date reintroduces the late-backfill blind spot (the same one addressed in the #1355 spike-check follow-up by switching to `collected_at`).

**3. "Silent drop" upstream blindspot — ✅ acknowledged, out of scope for this file**
Correct: a dbt test on the silver output can't catch an upstream `GREATEST(0, …)` clamp or a `WHERE count >= 0` drop. That's covered by the broader strategy — pair this with a pytest e2e test using a known-bad bronze seed (part of #1321), not something fixable in this SQL.

mitasovr (approved):

PR #1351, opened 2026-06-16
ci: advisory AI failure-investigation workflow (optional, needs ANTHROPIC_API_KEY)

SharedQA:
Split out of #1317 so the **key/funding decision** doesn't hold up the real gates.

**What it is:** advisory only — on a same-repo CI gate *failure* it reads the logs + co-located specs and posts a root-cause comment on the PR. **Never a merge gate.** Forks can't trigger it (secret-exfiltration guard). Capped at `--max-turns 30` / 15 min.

**What it needs:** an Anthropic **API key** (console.anthropic.com) as the `ANTHROPIC_API_KEY` Actions secret (repo or org-level). Billed as metered API usage to whichever Anthropic org owns the key; small since it only runs on failures.

**Parked as draft** until Constructor decides whether to fund/use it. None of the real quality gates depend on this. Can add a skip-guard so it cleanly no-ops when the secret is unset — say the word.

## Summary by CodeRabbit

* **New Features**
  * Introduced an AI-powered workflow that automatically investigates failed mandatory CI/CD workflows. When failures occur, it collects failure logs and posts results as a single PR comment (or opens a “CI failure investigation” issue if no PR exists), helping teams diagnose issues faster.


.github/workflows/ai-investigate.yml
```diff
@@ -0,0 +1,102 @@
...
+          } > /tmp/failure/meta.txt
+
+      - name: AI investigation (full co-located context)
+        # Pinned to the supported moving major @v1 per Anthropic's published
+        # guidance (the v1 tag is the maintained release channel; a frozen SHA
+        # would miss security fixes). This step runs only on workflow_run after
+        # a gate already failed, is advisory, and creates no release artifact.
+        uses: anthropics/claude-code-action@v1
+        with:
+          anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
```
cyberantonz, line 100: What happens here if the secret is not present? I suggest to add checks here and in `Collect failing job logs` and skip jobs if there is no `secret.ANTHROPIC_API_KEY` is set

cyberantonz (commented):
Need work

.github/workflows/ai-investigate.yml
SharedQA: Done — gated on the secret. Added a job-level `env: HAS_ANTHROPIC_KEY: ${{ secrets.ANTHROPIC_API_KEY != '' }}`, and both "Collect failing job logs" and "AI investigation" now run only `if: env.HAS_ANTHROPIC_KEY == 'true'`. When the key is absent, a note step logs the skip and the job passes — it's advisory, never a gate, so a missing key must not fail CI.

PR #1354, opened 2026-06-16
fix(github-copilot): lookback re-fetch on daily-metrics streams (+ seats Link pagination)

mozhaev-dev:
## Problem
The `copilot_org_metrics` / `copilot_user_metrics` streams advanced the incremental cursor **per-slice unconditionally** (the "Major #5" rule) and started the next run at `cursor+1`. But Copilot daily reports **lag** (often >24–48h; weekends later) and can be **restated**, so a recent day that returned **HTTP 204 because its report wasn't ready yet** still advanced the cursor past it — and was then **never re-fetched** once the data landed. → silent data gaps for late-arriving / adjusted days.

Same class of bug as the jira `#941` cursor-skip and the Zendesk `lookback_window` lesson (`#1316`): never advance the cursor past data that isn't finalized/available.

## Fix
Both report streams now re-query a **trailing lookback window** — `start = max(start_date, cursor − metrics_lookback_days)` instead of `cursor+1` — so the recent tail self-heals every run. New optional config **`metrics_lookback_days`** (default **7**), wired through `source.py` + `spec.json`. RMT dedup on `unique_key` (`{tenant}-{source}-[user]-{day}`) makes the overlap idempotent; genuinely-empty historical days are still not re-fetched forever (only the trailing window is revisited).

Also: `copilot_seats` pagination now follows GitHub's `Link: rel="next"` header (count heuristic kept as fallback) instead of stopping purely on "page < 100".

`descriptor` 2.1.0 → **2.2.0** (minor: new optional config, no Bronze schema change).

## Verified live (real Copilot Business org)
- Logic: cursor `2026-06-13` + lookback 7 → slices `2026-06-06…06-15` (re-includes the active pre-cursor days the old `cursor+1` skipped).
- **Live `read`**: re-fetched `copilot_org_metrics` for all 7 pre-cursor days + `copilot_user_metrics` real per-user rows on those days + `copilot_seats` (5, Link-paginated); **0 errors**; state advances correctly to `06-15`.

## Summary by CodeRabbit

* **New Features**
  * Added configurable `metrics_lookback_days` (default: 7) to re-fetch a trailing window for Copilot daily metrics and avoid permanent skips when reports arrive late.

* **Improvements**
  * Updated daily metrics ingestion to use the trailing lookback window based on the saved cursor, improving resilience to late/restated data (with downstream deduplication).
  * Improved pagination for Copilot seat ingestion by prioritizing GitHub `Link` header `rel="next"` navigation.
  * Expanded dbt incremental reprocessing windows from 3 to 7 days to align with connector lookback behavior.

### DIFF
--- a/src/ingestion/connectors/ai/github-copilot/dbt/copilot__ai_dev_usage.sql
+++ b/src/ingestion/connectors/ai/github-copilot/dbt/copilot__ai_dev_usage.sql
@@ -171,8 +171,13 @@ WHERE s.user_email IS NOT NULL
     OR coalesce(m.used_cli, false)
   )
 {% if is_incremental() %}
+  -- 7-day re-process window: must be >= the connector's metrics_lookback_days
+  -- (default 7, source_github_copilot) so days the connector re-fetches when a
+  -- Copilot report lands late or is restated actually reach Silver. A narrower
+  -- window would let those days sit in Bronze but be dropped here (older than
+  -- max(day) - N), wasting the connector lookback. (Refs #1354.)
   AND toDate(parseDateTimeBestEffortOrNull(m.day)) > (
-      SELECT coalesce(max(day), toDate('1970-01-01')) - INTERVAL 3 DAY
+      SELECT coalesce(max(day), toDate('1970-01-01')) - INTERVAL 7 DAY
       FROM {{ this }}
   )
 {% endif %}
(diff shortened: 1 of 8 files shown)


mozhaev-dev:
Addressed the two cosmetic notes from review in `8a747c93` (no functional change):
1. Removed the now-orphaned `_next_day` helper from `org_metrics.py` / `user_metrics.py` — the lookback change removed its only call-sites (confirmed it's used nowhere else in the package).
2. `seats.next_page_token` Link-header branch now returns `{"page": int(...)}` for type-parity with the fallback branch.

Re-verified after the change: slices still re-fetch the trailing window (cursor 2026-06-13 + lookback 7 → 06-06…06-15), imports clean.

mitasovr (approved):

Issue #752, opened 2026-06-04
Silver + connector staging watermarks not partitioned by tenant_id/source_id — silent data loss on multi-tenant


il10241024:
> 🔗 **Mirrored from** [cyberfabric/cyber-insight#290](https://github.com/cyberfabric/cyber-insight/issues/290)
> Originally opened by **aleksdotbar** on 2026-05-05T13:56:43Z

---

## Summary

All connector staging models and every silver `class_*` model use a global incremental watermark:

```sql
WHERE _version > (SELECT max(_version) FROM {{ this }})
```

In a multi-tenant deployment this silently drops rows. If tenant A has rows up to _version = 1000 and tenant B is onboarded with backfill at _version = 500, tenant B's rows are filtered out — 500 > 1000 is false. No errors, no logs.

Affects every late-joining tenant, every paused-then-resumed source, every historical backfill.

## Fix
Replace global subquery with per-tenant/source HWM JOIN. Reference: hubspot__crm_accounts.sql:

```sql
{% if is_incremental() %}
SELECT src.*
FROM src
LEFT JOIN (
  SELECT tenant_id, source_id, max(_version) AS hwm
  FROM {{ this }}
  GROUP BY tenant_id, source_id
) w ON w.tenant_id = src.tenant_id AND w.source_id = src.source_id
WHERE src._version > coalesce(w.hwm, 0)
{% else %}
SELECT * FROM src
{% endif %}
```

---

PR #1360, opened 2026-06-17
fix(ingestion): per-(tenant,source) incremental watermark — stop silent multi-tenant data loss (#752)

SharedQA:
## What
Replaces the **global** incremental watermark with a **per-(tenant, source)** one, via a new macro `incremental_watermark()`. Closes the data-loss half of #752.

## Why
Silver/connector incremental models filter new rows with:
```sql
WHERE _version > (SELECT max(_version) FROM {{ this }})
```
In a multi-tenant table this **silently drops** rows for any `(tenant, source)` whose own `max(_version)` is below the table-wide max — every late-joining tenant, paused-then-resumed source, and historical backfill. No error, no log.

## How
New macro `dbt/macros/incremental_watermark.sql` filters each row against the high-watermark of **its own** `(tenant, source)` using a LEFT JOIN (ClickHouse has no correlated subqueries). Column names are passed per-model because they vary across silver (`tenant_id` vs `insight_tenant_id`, `source_id` vs `insight_source_id`).

## Verified (live Kind instance, seeded data)
- macro compiles to valid ClickHouse SQL with correct per-model columns
- all 4 converted models build (`--full-refresh`) and the incremental JOIN executes — **0 errors**
- **two-tenant proof:** existing target tenant A (`_version` 5000–5010); incoming batch = A new row (5011) + tenant B backfill (3000–3004). Old predicate keeps **B=0** (dropped); new keeps **B=5** (preserved); A unaffected in both.

## Scope of this PR (4 of ~42 models)
Converted + end-to-end verified (had seeded data):
- [x] `silver/collaboration/class_collab_chat_activity` (tenant_id, insight_source_id)
- [x] `silver/collaboration/class_collab_email_activity` (tenant_id, insight_source_id)
- [x] `silver/collaboration/class_collab_meeting_activity` (tenant_id, insight_source_id)
- [x] `silver/ai/class_ai_dev_usage` (insight_tenant_id, source_id)

**Remaining ~38 models** use the same global-watermark pattern and should adopt the macro — but each needs its `tenant_col`/`source_col` confirmed (the naming is inconsistent) and ideally a build check, so they're intentionally **not** swept blindly here. Full list: `grep -rl "max(_version) FROM {{ this }}" src/ingestion`. Happy to follow up per-domain.

## Side note (separate issue worth filing)
The silver layer mixes `tenant_id`/`insight_tenant_id` and `source_id`/`insight_source_id` across models — a data-contract inconsistency that made this fix per-model rather than blanket.

## Release Notes

* **New Features**
  * Introduced enhanced incremental data filtering scoped per tenant and data source, providing more granular control over data synchronization.

* **Refactor**
  * Updated multiple incremental data pipeline models to leverage optimized filtering mechanism, improving efficiency and consistency across data ingestion.


SharedQA:
Thanks @coderabbitai — valid catch, and exactly the data-loss class this PR targets. Fixed in 6cccd04c.

One nuance: the suggested `_wm_hwm IS NULL OR …` guard doesn't actually fire in ClickHouse. Unmatched LEFT JOIN columns are filled with type **defaults (0), not NULL**, unless `join_use_nulls=1` — I verified a new `(tenant, source)` pair with a `version=0` row was *still* dropped with the `IS NULL` form. So I used an explicit `_wm_seen` flag (`1 AS _wm_seen` in the aggregate; unmatched → 0) as the "never loaded" signal instead.

Verified on a live instance — new pair keeps all rows including `version=0`, existing pair keeps only rows above its own high-watermark, and all 4 models still build with 0 errors.

ktursunov:
Should be fixed in scope of https://github.com/constructorfabric/insight/issues/1561

PR #1362, opened 2026-06-17
test(e2e): bronze→silver transform test for collab document activity

SharedQA:
## What

Adds an e2e test that exercises the **bronze→silver transform** for `class_collab_document_activity`, closing the blind spot in the data-quality silver-seed PoC (#1348): that PoC seeds silver directly and proves "bad silver → check fires", but never proves the transform itself is correct.

This test seeds **bronze** M365 sharepoint/onedrive rows, runs `dbt build --select +class_collab_document_activity`, and asserts the silver output — column mapping, `person_key` lower-casing, `reportRefreshDate`→`date`, the sharepoint/onedrive union, `unique_key` MD5, and OneDrive's NULL `visited_page_count`.

**Headline assertion:** a negative `viewedOrEditedFileCount` is *preserved* through the transform into silver — the exact case the silver-seed PoC could only assume.

## Fixes required to make it run (each a real issue)

1. **`create-bronze-placeholders.sh`** — the m365 document-activity bronze placeholders were missing the columns the staging models read. `tenant_id`/`source_id` are connector-injected physical fields (connector.yaml adds_fields); also `reportRefreshDate`, `reportPeriod`, `viewed/synced/visited` counts. `visitedPageCount` typed `Int64` to match the OneDrive staging cast — a `Float64` makes the silver UNION fail `NO_COMMON_TYPE`.
2. **`dbt_runner.py`** — profile host was hardcoded `127.0.0.1`; the dockerized runner reaches CH at the compose service name. No existing fixture runs `dbt build` (both omit `dbt_selector`), so this was latent/dead. Now uses `cfg.ch_host`.
3. **`pytest.ini`** — collect the new `transform/` dir.

## Verification

`./e2e.sh test -k document_activity_transform -v` → **`1 passed`**, run twice from a clean stack.

## Note for reviewers

Touches shared rig infra (`dbt_runner.py` host, `pytest.ini`, placeholders), so blast radius is slightly wider than a lone test file.


ktursunov:
Covered with https://github.com/constructorfabric/insight/issues/1413

PR #1358, opened 2026-06-17
test(e2e): PoC — run a data_quality check against seeded silver (#1348)

SharedQA:
Proof-of-concept for the question on #1348 — *how do we actually test the `data_quality` catalog checks, which need a warehouse and data?* Concrete answer: reuse the existing bronze→API e2e rig and add the two missing pieces.

**What's here**

- **`DbtRunner.run_test(<check>)`** — runs `dbt test --select <check>` and returns `(status, failures)` from `run_results.json`. The catalog checks are `severity='warn'`, so dbt exits 0 even on a violation; reading the `failures` count is exactly how the deployed emitter detects findings. `--defer --state` resolves the model `ref()` to the seeded silver relation without rebuilding (and wiping) it.
- **`dataquality/test_collab_document_counts_non_negative.py`** — the good/bad fixture pair for the #1321 non-negative check (PR #1350):
  - seed clean rows → **0 violations, status `pass`**
  - seed one negative count among good rows → **exactly 1 flagged, status `warn`** (proving warn-severity surfaces a finding without failing the build)

**Two drifts this PoC surfaced** (the genuinely useful part for the discussion):

1. The **silver `class_collab_document_activity` placeholder** had drifted from the model — missing `unique_key`, `synced_count`, `visited_page_count`, `product`. Aligned here so the check's projection resolves.
2. The **bronze `sharepoint_activity` placeholder** lacks the document-activity source columns, so the full bronze→silver build of this model isn't rig-testable yet. This PoC seeds **silver** directly to sidestep that; extending the bronze placeholder is the natural follow-up that would let the whole path run.

**Status: draft.** Validated against the rig's conventions and APIs but **not run locally** (no warehouse here) — the e2e CI job on this PR is the first real execution. Stacked on #1350 (the check under test lives there), so the diff includes that file until #1350 merges.

Opening this as a worked example to anchor the #1348 plan, not as a merge candidate yet.

## Summary by CodeRabbit

* **New Features**
  * Introduced end-to-end data quality testing framework with controlled fixture seeding to validate catalog checks
  * Added capability to run individual dbt data-quality checks with result extraction

* **Tests**
  * Added data quality validation tests for check accuracy
  * Enhanced test configuration validation

* **Chores**
  * Extended placeholder table schema with additional required columns for type-checking

Issue #736, opened 2026-06-04
M365 silver: unlicensed users (isLicensed=false) leak into class_collab_meeting_activity


il10241024:
> 🔗 **Mirrored from** [cyberfabric/cyber-insight#264](https://github.com/cyberfabric/cyber-insight/issues/264)
> Originally opened by **dzarlax** on 2026-05-04T09:45:45Z

---

## Symptom
`silver.class_collab_meeting_activity` (M365 feeder) does not filter on `isLicensed`, so unlicensed users — guests, ex-employees, service accounts — leak into team collaboration metrics.

## Evidence (local CH dump, 2026-05-04)
- `bronze_m365.teams_activity`: 25 rows with `isLicensed=false` out of 5711 (~0.4%).
- The same gap likely affects email/onedrive/sharepoint silver feeders (need verification on each).

## Impact
- Team-level meeting / call counters are slightly inflated (low magnitude, but conceptually wrong).
- Unlicensed `userPrincipalName` values flow into IC-level views without a matching `insight.people` row, producing orphan rows in identity-resolved gold views.

## Proposed fix
Add `AND coalesce(isLicensed, false) = true` to the WHERE clause in:
- `src/ingestion/connectors/collaboration/m365/dbt/m365__collab_meeting_activity.sql` (line 42-43).
- `src/ingestion/connectors/collaboration/m365/dbt/m365__collab_email_activity.sql` (verify column presence first; bronze schema would need `isLicensed` propagated if missing).
- `src/ingestion/connectors/collaboration/m365/dbt/m365__collab_document_activity_onedrive.sql` (same).
- `src/ingestion/connectors/collaboration/m365/dbt/m365__collab_document_activity_sharepoint.sql` (same).

If `isLicensed` is not present in some bronze tables, mirror the field via the connector mapping in `connector.yaml`.

---

il10241024:
**dzarlax** commented on 2026-05-04T10:07:28Z:

---

Revising the proposed fix here in light of the same class of bug found in Slack (#737).

The original "filter `isLicensed=false` out at silver" approach loses the signal permanently — we then can't ever look at unlicensed-user behavior (e.g. guest engagement in Teams calls, recently-offboarded activity) for retroactive analysis.

[...]

PR #1363, opened 2026-06-17
test(data): mapping-coverage drift/floor check + e2e fixture (#736)

SharedQA:
A referential **mapping-coverage** guard for #736 (Entra ≫ BambooHR; orphan unlicensed M365 users). It flags the latest complete day per tenant when the share of active users that map to a BambooHR employee (`person_key = lower(email)`) either drops **>5 points** below its 7-day trailing average **or** falls under a **75% floor**. `data_quality` catalog: `severity=warn` + `store_failures`, reads `FINAL`.

**This is the corrected form of a proposed design.** The original draft SQL wouldn't run here: it joined a non-existent `class_people.person_key` (the roster has `email`), used Postgres idioms (`::float`, `date_trunc`, `current_timestamp`, `nullif`), and was a generic test wired via `schema.yml` rather than this repo's singular `data_quality` convention. Fixed: email-equality join, ClickHouse functions + LEFT-JOIN default-not-null semantics, singular tagged test.

**Decisions baked in:** BambooHR is the denominator (Decision 1 — Entra-only = expected external); relative drift **and** absolute floor (Decision 3).

**Placeholder alignment** (the rig couldn't test this otherwise — both were drifted from the real models): `class_people` placeholder was missing `tenant_id` + `source`; the activity placeholder used `insight_tenant_id` while the real model emits `tenant_id`. Added additively.

**e2e fixture** (`dataquality/test_collab_document_mapping_coverage.py`): a healthy multi-day series passes; 7 healthy days + an all-orphan latest day trips both floor and drift (exactly one tenant-day, warn).

**Honest follow-ups, not closed here:**
- `person_key` is email, so a renamed user (marriage/legal change) orphans until the identity service lands `person_id` in silver — that's the deeper fix (upstream), not this check.
- Slack `user_id` resolution and a **by-domain orphan rollup** for steward triage are separate.

Stacks on #1358 (shares `DbtRunner.run_test` + the `dataquality/` harness). Draft — the e2e CI job is the validator.


ktursunov:
Already wired

PR #1380, opened 2026-06-19
test(e2e): declarative YAML metric-test rig (replaces the CSV rig)

### LINKED ISSUE #1422: test(e2e): declarative YAML metric-test rig (replaces the CSV rig)
## Goal

Replace the per-folder **CSV** e2e rig with a single declarative **`<name>.test.yaml`** format (plus specs and an authoring skill), so each metric's full path — `bronze → dbt silver → gold view → analytics-api` — is seeded and asserted from one fixture file. The path itself is unchanged; only the authoring format and the assertion engine change.

Implemented by #1380.

## Proposed approach (what #1380 delivers)

**Format** — `src/ingestion/tests/e2e/fixtures/`
- `schemas/<db>.<table>.yaml` — one JSON schema per bronze table (all real Airbyte columns; resolved by table name).
- `templates/*.yaml` — reusable records; a record is a field map with optional `$ref: "<file>#/<json-pointer>"` to inherit + sibling-key override (closest wins). Base records carry every column incl. `_airbyte_*` (transforms depend on `_airbyte_extracted_at`).
- `bronze:` keyed by table → records (`$ref` + the fields under test); padded to the full schema and validated at load; duplicate rows allowed (dedup must hold).
- `cases:` → `request` to the batch endpoint `POST /v1/metrics/queries`; `expect` is a list of rules: `in` (pick result by id) / mongo-style `find` / `equal` (subset) or CEL `assert`.

**Runner** — `e2e_lib/`
- `ref_resolver` (compose), `schema_validator` (pad + validate), `expect_engine` (find + equal + CEL via `cel-python`).
- `fixture_loader` rewritten; `ch_seeder` seeds resolved records; `analytics_api.call_request` does the batch call; `dbt_runner.derive_selectors` derives dbt models from the manifest.
- `test_fixtures` runs truncate → seed → 2-pass dbt → recreate gold views → refresh MV → batch → evaluate.
- Retires `csv_asserter` / `spec_schema` and the CSV-era meta tests; adds `meta/test_ref_resolver.py` (12 invariants) + `meta/test_expect_engine.py`.

**Specs & skill**
- `docs/domain/bronze-to-api-e2e/specs/feature-yaml-rig/FEATURE.md` (+ DESIGN v1.1, artifacts.toml).
- Skill `/metric-e2e-test` (guide + scaffold).

## Acceptance criteria

- [ ] Declarative `<name>.test.yaml` format (schemas + `$ref` templates + `bronze:` + `cases:`) is the authoring surface; the CSV rig and its meta tests are removed.
- [ ] `$ref` composition and the expect engine (`find` / `equal` / CEL) are covered by pure unit tests (`meta/test_ref_resolver.py`, `meta/test_expect_engine.py`).
- [ ] Bronze records are padded to the full table schema and validated at load; duplicate rows are allowed and must dedup.
- [ ] A reference fixture (`collab_emails_sent.test.yaml`) exercises the full path end to end — incl. a re-sync duplicate that dedups (value 40, median 20, range 10–40).
- [ ] `./e2e.sh test` is green on a fresh run.
- [ ] FEATURE/DESIGN specs and the `/metric-e2e-test` authoring skill are in-repo.

## Out of scope / follow-ups

- **m365 per-metric checks — follow-up, tracked in #1413.** Defining m365's bronze fixtures + per-metric golden values across all `class_collab_*` metrics builds *on top of* this rig (the rig ships one reference m365 metric; #1413 covers the rest).
- bronze→API gate endpoint-coverage extension — #1412.
- Parent: #1404 (Quality vector: Correctness), alongside siblings #1412 / #1413; part of the broader data-correctness effort (EPIC #1321).

mitasovr:
Closes #1422

## What

Replaces the per-folder CSV e2e rig with a single declarative **`<name>.test.yaml`** format, plus the specs and an authoring skill. The full path (`bronze → dbt silver → gold view → analytics-api`) is unchanged; only the authoring format and the assertion engine change.

### Format (`src/ingestion/tests/e2e/fixtures/`)
- `schemas/<db>.<table>.yaml` — one JSON schema per bronze table (all real Airbyte columns; resolved by table name).
- `templates/*.yaml` — reusable records; a record is a field map with optional `$ref: "<file>#/<json-pointer>"` to inherit + sibling keys override (closest wins). Base records carry every column incl. `_airbyte_*` (transforms depend on `_airbyte_extracted_at`).
- `bronze:` keyed by table → records (`$ref` + the fields under test); padded to the full schema and validated at load; duplicate rows allowed (dedup must hold).
- `cases:` → `request` to the batch endpoint `POST /v1/metrics/queries`; `expect` is a list of rules: `in` (pick result by id) / mongo-style `find` / `equal` (subset) or CEL `assert`.

### Runner (`e2e_lib/`)
`ref_resolver` (compose), `schema_validator` (pad + validate), `expect_engine` (find + equal + CEL via `cel-python`). `fixture_loader` rewritten; `ch_seeder` seeds resolved records; `analytics_api.call_request` does the batch call; `dbt_runner.derive_selectors` derives the dbt models from the manifest; `test_fixtures` runs truncate → seed → 2-pass dbt → recreate gold views → refresh MV → batch → evaluate. Retires `csv_asserter`/`spec_schema` and the CSV-era meta tests; adds `meta/test_ref_resolver.py` (12 invariants) + `meta/test_expect_engine.py`. Bronze placeholders for `bamboohr.employees` / `m365.email_activity` aligned to the real schema.

### Specs & skill
`docs/domain/bronze-to-api-e2e/specs/feature-yaml-rig/FEATURE.md` (+ DESIGN v1.1, artifacts.toml). Skill `/metric-e2e-test` (guide + scaffold).

## Status — green
Fresh `./e2e.sh test` = **37 passed**, incl. the reference `collab_emails_sent.test.yaml` (IC bullet `…0012` team median over `m365_emails_sent`, with a re-sync duplicate that must dedup → value 40, median 20, range 10–40). Resolver + expect-engine covered by 20 pure unit tests.

## Release Notes

* **New Features**
  * Added a declarative YAML end-to-end test rig for analytics metrics, including schema validation and `/v1/metrics/queries` batch testing.
  * Introduced rule-based expectations with CEL boolean assertions and result/row selection.

* **Refactor**
  * Migrated E2E fixtures from CSV folder structures to single `*.test.yaml` files with `$ref`-driven bronze record composition and dedup scenarios.

* **Documentation**
  * Replaced prior YAML-rig docs with a complete authoring/execution guide and examples.

* **Bug Fixes / Chores**
  * Improved warm-run determinism by clearing stale seeded tables between re-runs; expanded placeholder schemas for broader test compatibility.


src/ingestion/tests/e2e/e2e_lib/expect_engine.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def _floatify(obj: Any) -> Any:
+    """Coerce JSON numbers to float so CEL comparisons on metric values work
+    regardless of whether the API serialized e.g. `40` or `40.0` (CEL is strictly
+    typed and will not compare IntType to DoubleType). `bool` is left intact."""
+    if isinstance(obj, bool):
+        return obj
+    if isinstance(obj, int):
+        return float(obj)
```
cyberantonz: Is it ok about precision lost on big numbers?

```
1000000000000000.0
>>> a = int(1_123_123_123_123_123_123)
>>> b = float(a)
>>> b
1.1231231231231231e+18
>>> c = int(b)
>>> c
1123123123123123072
```

.claude/skills/metric-e2e-test/SKILL.md
```diff
@@ -0,0 +1,185 @@
...
+user-invocable: true
+allowed-tools: Bash, Read, Write, Edit, Glob, Grep
+---
+
+# Author a metric e2e test (declarative YAML)
+
+This skill writes and validates `*.test.yaml` fixtures that drive the full
+`bronze → dbt silver → gold view → analytics-api` path and assert the result.
+
+## Source of truth (read THIS turn before authoring)
```
ktursunov: Is that required for this skill? We could reference those as sources, but does it required to load into context every time?

.claude/skills/metric-e2e-test/SKILL.md
```diff
@@ -0,0 +1,185 @@
...
+  import yaml
+  s=set(yaml.safe_load(open("schemas/<db>.<table>.yaml"))["schemas"]["<db>.<table>"]["properties"])
+  t=set(yaml.safe_load(open("templates/<group>.yaml"))["templates"]["<base>"]); t.discard("$ref")
+  print("missing", sorted(s-t), "extra", sorted(t-s))
+  PY
+  ```
+- Each `expect` rule has `find`+(`equal`|`assert`) or a bare `assert`; `in` matches a
+  declared query `id`; CEL expressions parse.
+
+## Running
```
ktursunov, line 183: Suggest to add commands examples for all tests run, list existing tests along with each test execution

.claude/skills/metric-e2e-test/SKILL.md
```diff
@@ -0,0 +1,185 @@
...
+| `status` | the batch HTTP status code (int) | always |
+
+Numbers under `it`/`items`/`result`/`results` are float-coerced (CEL won't compare
+`int` to `double`) → compare metric values with float literals (`it.value > 39.5`);
+`status` and `size(...)` are ints. For exact / `null`, use `equal` (Python `==`), not
+`assert`. CEL macros available: `size()`, `has()`, `.exists()`, `.all()`, `.map()`, `.filter()`.
+
+## Scaffolding a new test
+
+1. **Resolve the metric_id and its shape.** Find it in the seed catalog
```
ktursunov, line 142: Let's create script that will query all existing metrics and their final schemas? Probably not in this pr

.claude/skills/metric-e2e-test/SKILL.md
```diff
@@ -0,0 +1,185 @@
...
+| Binding | Value | Present when |
+|---|---|---|
+| `it` | the single row matched by `find` | only with `find` (else `null` → `it.x` errors) |
+| `items` | the selected result's `items` array | a result is selected (`in` or sole query) |
+| `result` | the selected result `{id, status, metric_id, items, page_info}` | a result is selected |
+| `results` | the full `results[]` of the batch | always |
+| `status` | the batch HTTP status code (int) | always |
+
+Numbers under `it`/`items`/`result`/`results` are float-coerced (CEL won't compare
+`int` to `double`) → compare metric values with float literals (`it.value > 39.5`);
```
ktursunov: Which file type comparisons supported? Worth addition additional docs.

.claude/skills/metric-e2e-test/SKILL.md
```diff
@@ -0,0 +1,185 @@
...
+
+1. **Resolve the metric_id and its shape.** Find it in the seed catalog
+   (`grep -rn "<label>" src/backend/services/analytics-api/src/migration/*.rs`) and
+   the live `query_ref` rewrite for that metric. Note whether it returns a bullet
+   (`metric_key`/`value`/`median`/`range_*`) or per-person rows, and whether `median`
+   is company-wide (Team bullet) or team/org_unit (IC bullet).
+2. **Ensure a schema file per table.** If `schemas/<db>.<table>.yaml` is missing,
+   generate it from the REAL table (do not invent columns):
+   ```bash
+   export KUBECONFIG=/Users/roman/alemira/insight/access/dev-vhc/insight-k8s.kubeconfig
```
ktursunov: Hardcode. Let's use local run related.

src/ingestion/tests/e2e/e2e_lib/expect_engine.py
```diff
@@ -0,0 +1,153 @@
...
+            elif op == "$regex" and (actual is None or not re.search(operand, str(actual))):
+                return False
+            elif op == "$exists" and ((actual is not None) != bool(operand)):
+                return False
+        return True
+    return actual == cond
+
+
+def _find(items: list[dict], selector: dict) -> list[dict]:
+    return [it for it in items if all(_match_value(it.get(f), c) for f, c in selector.items())]
```
cyberantonz: Is there any test using this notation? I saw plain `eq` not `$eq`

ktursunov (changes requested):

src/ingestion/tests/e2e/e2e_lib/expect_engine.py
```diff
@@ -0,0 +1,153 @@
...
+class ExpectError(AssertionError):
+    """A failing expect rule. Message names the case, rule and the mismatch."""
+
+
+# ---------------------------------------------------------------------------
+# Mongo-style selector
+# ---------------------------------------------------------------------------
+
+def _match_value(actual: Any, cond: Any) -> bool:
+    if isinstance(cond, dict) and any(k.startswith("$") for k in cond):
```
cyberantonz: You bring ceply, why you need to invent other language?

cyberantonz (commented):
some questions

src/ingestion/tests/e2e/specs/collab_emails_sent.test.yaml
```diff
@@ -0,0 +1,85 @@
...
+  - name: m365_emails_sent — IC bullet (alice value vs team median)
+    request:
+      url: /v1/metrics/queries
+      method: POST
+      body:
+        queries:
+          - id: collaboration
+            metric_id: 00000000-0000-0000-0001-000000000012
+            $top: 50
+            $filter: "person_id eq 'alice@example.com' and metric_date ge '2026-01-01' and metric_date le '2026-01-31'"
```
cyberantonz, line 72: not `$eq`, not `!=`....  same in skill.

mitasovr:
Thanks @coderabbitai — addressed the actionable findings in `1fbacb47`.

**Applied:**
- `expect_engine` — `find` now raises on an unknown `$`-operator (was silently matching). Added regression tests for unknown-op and `$exists` vs `null`. (`$exists` itself was already correct.)
- `ref_resolver` — resolve list elements recursively so a `$ref` nested in a list value (or a top-level list) is no longer skipped. Added tests.
- `ch_seeder` — strict Bool coercion: raise on non-canonical values instead of silently yielding `False`.
- `analytics_api.call_request` — send the JSON body for any method (was dropped for non-POST/PUT).
- Fail-fast guards: `fixture_loader` validates `bronze` is a mapping and each `case` has `request`+`expect`; `schema_validator` wraps YAML/shape errors as `SchemaError`; `migration_applier.reapply_migrations` errors when no migrations are found.
- `DESIGN.md` — escaped the `|` in the batch-endpoint table cell.
- `SKILL.md` — replaced the hardcoded personal `KUBECONFIG` path with a placeholder.

Unit tests: 24 passed; full `./e2e.sh test`: 41 passed.

**Not applied (stale or false positive):**
- *"Remove the duplicated artifact header" (`artifacts.toml`)* — no duplicate exists; the two `[[systems.artifacts]]` blocks (csv-rig, yaml-rig FEATURE) are both intentional and well-formed.
- *"Point the validation example / component path at `fixtures/`" (FEATURE)* — superseded: the directory was intentionally renamed `fixtures/ → specs/` in `484e2206`. Paths now correctly read `specs/`.
- *"Use exact numeric types for the new counters" (`create-bronze-placeholders.sh`)* — already resolved to `Nullable(UInt32)` during the upstream merge.
- *TOC fragment for the reference-test heading (FEATURE)* — the TOC is generated by `cpt toc` and passes `cpt validate-toc` (our gate).
- *"Rename the truncate-step algorithm reference" (FEATURE)* — cosmetic; the execute-test algorithm deliberately reuses the csv-rig truncate algorithm ID.
- *floatify precision on large ints (`expect_engine`)* — intentional: CEL is strictly typed and won't compare `int` to `double`; metric values are small. `status`/`size()` stay `int`.

mitasovr:
Thanks @cyberantonz / @ktursunov — addressed in `7fb42a35` + `577cb120`.

**@cyberantonz — expect engine**
- *"you bring celpy, why invent another language?"* / *"I saw plain `eq` not `$eq`"* / *"not `$eq`, not `!=`…"* — agreed. Dropped the Mongo-style `find` operators entirely (`$eq/$ne/$gt/$gte/$lt/$lte/$in/$regex/$exists`). `find` is now **exact field-equality only**; anything richer is a CEL `assert`. No second selector language. Specs/docs/skill updated to match.
- *precision lost on big numbers (float-coercion)* — agreed, removed the blanket float-coercion of bindings (it silently truncated large ints, exactly your `1.123e18` example). Bindings now pass through unchanged; since CEL is strictly typed, fractional comparisons cast explicitly: `double(it.value) > 39.5`. Docs/tests updated.

**@ktursunov — skill**
- *"is it required to load the specs into context every time?"* — no; reworded "Source of truth" to reference-only (consult when you need the algorithm/DoD or on a disagreement). 
- *"add commands for running all tests / listing tests"* — added `ls specs/*.test.yaml`, `./e2e.sh test`, `-k <name>`, `-v`, `down`.
- *"hardcode KUBECONFIG, use local-run related"* — the personal path is gone (now a `<path to your dev cluster kubeconfig>` placeholder). Note: the local rig CH only has placeholder schemas, so generating a real per-table schema is a one-off `DESCRIBE` against the dev cluster — hence a kubeconfig, not the local stack.
- *"which type comparisons are supported — worth docs"* — the CH-type → JSON-schema mapping is documented in the "Scaffolding a new test" step (Nullable(String)→`[string,"null"]`, Decimal/Float/Int→`[number,"null"]`, Bool→`[boolean,"null"]`, DateTime*→date-time, JSON→`[object,"null"]`).
- *"script that queries all metrics + their final schemas"* — agree it's useful but out of scope here (as you noted); good follow-up PR.

`./e2e.sh test` = 39 passed after the changes.

.claude/skills/metric-e2e-test/SKILL.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Running
+
+```bash
+cd src/ingestion/tests/e2e
+ls specs/*.test.yaml                       # list existing tests
+./e2e.sh test                              # run all tests (specs/ + meta/)
+./e2e.sh test -k <name>                    # run one test by name
+./e2e.sh test -k <name> -v                 # verbose (per-step log)
+./e2e.sh down                              # reset the stack; do this before a warm re-run
```
ktursunov: Why this script reset the stack? Deploy is for the ./dev-compose.sh

src/ingestion/tests/e2e/e2e_lib/analytics_api.py
```diff
@@ -277,22 +287,27 @@ def client(self) -> httpx.Client:
...
-            response = c.request(method, endpoint, **kwargs)
+            if body is not None:
+                kwargs["json"] = body
+            LOG.info("→ %s %s", method, url)
+            response = c.request(method, url, **kwargs)
             LOG.info("← %d  (%d bytes)", response.status_code, len(response.content))
-            return ApiResponse.from_httpx(response)
+            try:
+                payload = response.json()
+            except Exception:  # noqa: BLE001
```
ktursunov: Too broad exception catch. Suggest using exact json decode error exception catch

src/ingestion/tests/e2e/specs/test_fixtures.py
```diff
@@ -0,0 +1,69 @@
...
+from e2e_lib.expect_engine import evaluate_case
+from e2e_lib.fixture_loader import TestYaml
+from e2e_lib.migration_applier import refresh_intermediates, reapply_migrations
+from e2e_lib.worker import WorkerContext
+
+pytestmark = pytest.mark.fixture
+LOG = logging.getLogger("e2e.runner")
+
+
+def test_fixture(
```
ktursunov: Please rename to test_e2e_metric_smoke - in report test_fixture looks weird

src/ingestion/tests/e2e/meta/test_expect_engine.py
```diff
@@ -0,0 +1,92 @@
+"""Unit tests for expect_engine (DoD cpt-bronze-to-api-e2e-dod-yaml-expect-engine).
```
ktursunov, line 1: What's that test is about? test on test harness?

src/ingestion/tests/e2e/e2e_lib/dbt_runner.py
```diff
@@ -99,6 +99,32 @@ def build(
                 f"stderr tail:\n{result.stderr[-1000:]}"
             )
 
+    def derive_selectors(self, tables: set[tuple[str, str]]) -> tuple[list[str], list[str]]:
```
ktursunov, line 102: Silver layer is not needed for e2e tests now, suggest skipping it completely now

src/ingestion/tests/e2e/meta/test_ref_resolver.py
```diff
@@ -0,0 +1,150 @@
+"""Unit tests for ref_resolver — the 12 invariants of
```
ktursunov, line 1: Is this test required for test harness?

ktursunov (changes requested):

src/ingestion/tests/e2e/e2e_lib/dbt_runner.py
mitasovr: This rig tests the whole path from raw data to what the API returns — `bronze (seeded) → dbt staging → dbt silver class_* → insight.collab_bullet_rows (gold view) → /v1/metrics/queries`. The gold view reads `silver.class_collab_*`, so without the silver dbt transformations the view reads an empty placeholder and the metric returns zero rows — there's nothing left to assert.

Note we don't *seed* silver at all anymore; it's built by dbt from the seeded bronze (`derive_selectors` returns the staging models to build first, then their `silver:<class>` targets). The only way to "skip silver dbt" would be to go back to seeding silver tables directly — the approach we intentionally dropped, since it stops exercising the bronze→silver dedup/transform. So the silver build stays. Did you have a specific simplification in mind?

src/ingestion/tests/e2e/meta/test_expect_engine.py
mitasovr: Yes — these (and `test_ref_resolver.py`) are unit tests of the test harness's own logic, not of any product metric. `meta/` is where the rig tests itself.

This one covers `expect_engine`: `in` result selection, exact-equality `find`, `equal` subset matching, and CEL `assert` evaluation (incl. the `result.status`/`size(items)`/`double()` cases). They're pure and fast — no ClickHouse/dbt — so they catch regressions in the expect engine in milliseconds, independently of the full e2e run. Cheap insurance for the matcher that decides whether every metric test passes; I'd keep them.

src/ingestion/tests/e2e/meta/test_ref_resolver.py
mitasovr: Same as the `test_expect_engine.py` thread: this is a harness self-test under `meta/`, not a product test. It pins the `$ref`/sibling-override resolver — the 12 invariants from the FEATURE DoD (local & cross-file refs, override precedence, nested-ref-in-its-own-file, deep merge, cycle detection, missing file/pointer errors, list elements). Pure and fast (no ClickHouse/dbt). Since every fixture's bronze rows are built by this resolver, a regression here would silently corrupt seed data — so the unit coverage is worth keeping.

ktursunov (approved):

PR #1464, opened 2026-06-23
ci(helm): helm-validate workflow + CI render values (split 3/4 of #1317)

SharedQA:
Split out of #1317 (too large to review as one). This is the **helm** slice.

## What
- `.github/workflows/helm-validate.yml` — blocking chart gate:
  - Chart.yaml `version` (X.Y.Z) + `appVersion` (build-tag) format guard (the `---` defect class).
  - Forbid `:latest` image tags in values.
  - `helm lint --strict` (umbrella + service subcharts).
  - Full `helm template` render with CI values + `kubeconform` on the output.
- `charts/insight/ci/ci-values.yaml` — render-only placeholders (the chart intentionally leaves infra hosts/tags UNSET and fail-fasts; these let the render/kubeconform gate run).

## Notes
- Self-standing: helm/kubeconform run directly (no Makefile), no docs/security/release files.
- Verified locally: `helm template` renders 21 objects, `helm lint --strict` clean.
- Part of splitting #1317; #1317 keeps the release-train + coverage slice.


ktursunov:
https://github.com/constructorfabric/insight/pull/1559
Done here

Issue #1440, opened 2026-06-23
Metric smoke: AI Adoption metrics


ktursunov:
## AI Adoption — Metric smoke

Gold view `insight.ai_bullet_rows` · flows …0006 (Team) / …0013 (IC) · grain member · team. Source connectors: cursor, claude_code/claude_team, chatgpt_team/codex.

> **Happy-path *smoke* only.** Seed one clean, deterministic bronze dataset → assert the served values match the golden CSV. Negative / auth / empty / malformed / edge cases are **out of scope** (separate negative-test track). Fixture: `ai_smoke`. Runs in the rig (#1422).

## Target metrics (22)

- **Active members** — `active_ai_members` · ✅ real
- **Claude Code — active members** — `cc_active` · ✅ real
- **Claude Code Cost** — `cc_cost` · ✅ real
- **Claude Code Accepted Lines** — `cc_lines` · ✅ real
- **Claude Code Sessions** — `cc_sessions` · ✅ real
- **Claude Code Tool Acceptance** — `cc_tool_accept` · ✅ real
- **ChatGPT Activity** — `chatgpt` · ✅ real
- **ChatGPT — active members** — `chatgpt_active` · ✅ real
- **Codex — active members** — `codex_active` · ✅ real
- **Codex Accepted Lines** — `codex_lines` · ✅ real
- **Codex Sessions** — `codex_sessions` · ✅ real
- **Cursor — active members** — `cursor_active` · ✅ real
- **Cursor Agent Interactions** — `cursor_agents` · ✅ real
- **Cursor Completions** — `cursor_completions` · ✅ real
- **Cursor Accepted Lines** — `cursor_lines` · ✅ real
- **Team AI Accepted Lines** — `team_ai_loc` · ✅ real
- **AI Code Acceptance** — `ai_loc_share2` · 🔁 reconstructed
- **Claude Code Tool Acceptance** — `cc_tool_acceptance` · 🔁 reconstructed
- **Cursor Acceptance Rate** — `cursor_acceptance` · 🔁 reconstructed
- **Claude.ai Activity** — `claude_web` · ⛔ locked stub
- **Total PRs (CC window)** — `prs_total` · ⛔ locked stub
- **PRs with Claude Code** — `prs_with_cc` · ⛔ locked stub

*Legend — ✅ real (assert the gold value) · 🔁 reconstructed in `query_ref` (Σnum/Σden — assert the API value) · ⛔ locked stub / ComingSoon (assert NULL).*

[...]

PR #1468, opened 2026-06-24
test(e2e): AI metrics per-metric specs + bronze_cursor placeholder & watermark fixes (#1440)

mozhaev-dev:
Part of #1440 (E2E smoke: AI Adoption metrics). Covers the cursor-derivable AI metrics as **one spec per metric** (per review), plus two ingestion fixes the rig exposed.

## Tests — 10 specs, one per metric, over a clean cursor seed
- **Cursor:** `ai_cursor_lines`=40, `ai_cursor_completions`=50, `ai_cursor_agents`=30, `ai_cursor_active`=1-of-3, reconstructed `ai_cursor_acceptance`=62.5.
- **Tool-agnostic:** `ai_active_ai_members`=1-of-3, `ai_team_ai_loc`=40.
- **Locked stubs → NULL:** `ai_claude_web`, `ai_prs_total`, `ai_prs_with_cc`.

All assert the **IC Bullet AI** (…0013) served value/median/range; each cursor spec includes a re-sync duplicate that must dedup. All 10 pass together in one session.

## Fixes
1. **bronze_cursor placeholder** (`create-bronze-placeholders.sh`) — was a stale 14-column subset; the current `cursor__ai_dev_usage` model couldn't build without Airbyte. Aligned to the connector `InlineSchemaLoader`.
2. **cursor watermark underflow** (`cursor__ai_dev_usage`) — `coalesce(max(day),'1970-01-01') - INTERVAL 3 DAY` over an empty `this` wraps to ~2149-06-04 (Date underflow) and filters out every row, breaking per-test isolation. Guarded with `(SELECT count() FROM this) = 0 OR …` (m365 pattern). **NB:** this guard is copy-pasted across all 10 AI staging models — `claude_team` / `chatgpt_team` need the same one-liner in their follow-up PRs.
3. e2e `_SESSION_START_TRUNCATE` reset for `cursor__ai_dev_usage` (warm-rerun safety).

## Verification
`./e2e.sh test -k ai_` → **10 passed**. Branch up to date with main.

Refs #1440


mozhaev-dev:
Thanks @coderabbitai — both addressed in the latest push:
- `unique_key` is now non-null in `schemas/bronze_cursor.cursor_daily_usage.yaml` (matches the placeholder's dedup ordering).
- `_airbyte_extracted_at` is now RFC3339 with `Z` in `templates/cursor_usage.yaml`.

Also consolidated to a single `ai_smoke` fixture (per the #1440 design — one seed per shared gold view) and signed off the commits for DCO.

mozhaev-dev:
Thanks for the thorough review — and for reproducing the warm-rerun flake. 🙏

**Main finding (warm-rerun safety) — fixed.** Added `("staging", "cursor__ai_dev_usage")` to `_SESSION_START_TRUNCATE` in `conftest.py`, mirroring the collab append-staging tables. `class_ai_dev_usage` is `delete+insert` (self-correcting), so per your note it's left out.

Verified the exact scenario — two runs in a row on a **reused** CH volume (no `./e2e.sh down` between):
```
./e2e.sh down && ./e2e.sh test -k ai_smoke   # 1 passed
./e2e.sh test -k ai_smoke                    # 1 passed (was: FAIL unique_cursor__ai_dev_usage_unique_key)
```

**Minor notes (acknowledged, intentional):**
- *Dedup location* — correct, the duplicate collapses first in bronze (RMT `ORDER BY unique_key` + `optimize_on_insert`), with `union_by_tag` as the second guard. The fixture only asserts the contract (sums don't double), which holds deterministically; I'll keep the wording accurate.
- *Nullability vs InlineSchemaLoader* — deliberately relaxed: the fixture doesn't set `day`/`userId` on every row, so they must stay nullable in the placeholder/schema; `additionalProperties: false` is intentionally stricter than the connector. Not a literal mirror — the *column set* matches, the constraints are fixture-friendly.

Metrics correctness confirmation much appreciated. Ready for another look.

src/ingestion/tests/e2e/specs/ai_smoke.test.yaml
ktursunov, line 1: Suggest name file according to metric name as stored in metric catalog

mozhaev-dev:
@ktursunov re: one-spec-per-metric — I dug in, and there's a concrete (and fixable) blocker for the AI metrics specifically, not a fundamental one. Wanted to share before splitting.

**Symptom.** With per-metric specs, the 2nd cursor spec in a session fails: `find {metric_key: ...} matched 0 rows`. Each spec passes in isolation. Confirmed in the rig after a 2-spec run:
```
bronze_cursor.cursor_daily_usage = 3   # seed is there
staging.cursor__ai_dev_usage     = 0   # ← append rebuild inserted nothing
silver.class_ai_dev_usage        = 0
```

**Root cause — a Date underflow in `cursor__ai_dev_usage`'s incremental guard.** The new per-test `TRUNCATE` (staging-in-ledger reset, 3ccb0276) empties `this` between specs. The model filters:
```sql
AND toDate(...) > (SELECT coalesce(max(day), toDate('1970-01-01')) - INTERVAL 3 DAY FROM {{ this }})
```
Over an empty `this`, `max(day)` returns `1970-01-01` (not NULL, so the coalesce doesn't kick in), and `1970-01-01 - INTERVAL 3 DAY` **underflows the Date range and wraps to `2149-06-04`**:
```
SELECT toDate('1970-01-01') - INTERVAL 3 DAY                         -- 2149-06-04
SELECT toDate('2026-01-05') > toDate('1970-01-01') - INTERVAL 3 DAY  -- 0
```
→ the watermark jumps to the future, every bronze row is filtered out, 0 rows inserted. In prod this is masked (the silver table is never empty mid-incremental); the rig's per-test truncate is what exposes it.

**Why collab/jira per-metric specs are unaffected.** `m365__collab_email_activity` short-circuits the empty case: `(SELECT count() FROM {{ this }}) = 0 OR ...`. `cursor__ai_dev_usage` doesn't.

**Proposed fix.** Mirror the m365 guard in `cursor__ai_dev_usage` (and any AI sibling staging with the same `max(day) - N day` pattern): `(SELECT count() FROM {{ this }}) = 0 OR ...`, or avoid the underflow via `greatest(coalesce(max(day),'1970-01-01'), '1970-01-04') - INTERVAL 3 DAY`. That unblocks one-spec-per-metric for AI **and** fixes a latent incremental bug.

Want me to include that model fix in this PR and split `ai_smoke` into per-metric specs, or keep `ai_smoke` here and do the model fix + split separately?

mozhaev-dev:
@ktursunov done — split into one spec per metric as you asked (10 specs over the cursor seed; all green together).

One systemic heads-up beyond my earlier cursor note: that Date-underflow watermark (`coalesce(max(day),'1970-01-01') - INTERVAL 3 DAY`) is **copy-pasted across all 10 AI staging models** — cursor, claude_team, chatgpt_team, claude_admin×2, claude_enterprise×2, copilot×2. I fixed cursor's here (mirrors the m365 `count()=0 OR` guard); `claude_team` / `chatgpt_team` will need the same one-liner in their Batch-B follow-up PRs. Flagging in case you'd rather land one shared guard fix across all AI staging models instead.

Up to you/QA whether to land this per-metric split or prefer a single fixture — happy either way.

ktursunov (approved):

PR #1492, opened 2026-06-26
ci: add K3s deployment smoke test

BlackCelebrant:
## Summary

The workflow creates an ephemeral k3d/K3s cluster in GitHub Actions, installs the required L2 services separately, and deploys the current Insight umbrella chart into the cluster.

## What changed

- Added `Functional K3s Smoke` workflow with manual `workflow_dispatch` trigger.
- Installs K3s via k3d using `rancher/k3s:v1.36.1-k3s1`.
- Installs required L2 infra into `insight-infra`:
  - MariaDB
  - ClickHouse
  - Redis
- Pre-creates required database secrets for the app chart.
- Deploys `charts/insight` into the `insight` namespace.
- Dumps pods, events, logs, descriptions, and Helm status on failure.
- Adds a CI values overlay aligned with the documented Kubernetes seed defaults.

## Notes

This is a deployment smoke test only. It does not run demo-data seeding yet; Kubernetes seeding remains manual per `CONTRIBUTING.md`.

## Summary by CodeRabbit

* **New Features**
  * Added a new functional deployment check for the app on a temporary K3s environment.
  * Expanded the deployment setup with a streamlined configuration for core services and app components.

* **Bug Fixes**
  * Improved validation of full-stack startup and rollout behavior before changes are accepted.
  * Added richer diagnostics to help identify deployment issues faster.

### DIFF
--- a/.github/ci/insight-functional-values.yaml
+++ b/.github/ci/insight-functional-values.yaml
@@ -1,0 +1,66 @@
+##
+## Minimal values for the fork-safe functional K3s workflow.
+##
+## The current umbrella chart ships only first-party app services. The
+## workflow installs ClickHouse, MariaDB, and Redis as separate releases in
+## `insight-infra`, then points this app chart at those service names.
+##
+
+global:
+  storageClass: local-path
+  tenantDefaultId: "11111111-1111-4111-8111-111111111111"
+
+credentials:
+  deploymentMode: helm
+  autoGenerate: true
+
+ingestion:
+  templates:
+    enabled: false
+  dataQuality:
+    enabled: false
+
+clickhouse:
+  host: clickhouse.insight-infra.svc.cluster.local
+  database: insight
+  username: insight
+
+mariadb:
+  host: mariadb.insight-infra.svc.cluster.local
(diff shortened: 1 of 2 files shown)


.github/ci/insight-functional-values.yaml
```diff
@@ -0,0 +1,66 @@
...
+analyticsApi:
+  replicaCount: 1
+  image:
+    tag: "2026.06.23.17.18-5f3df70"
+  resources:
+    requests: { cpu: 50m, memory: 128Mi }
+    limits: { cpu: 250m, memory: 384Mi }
+
+identity:
+  deploy: false
```
cyberantonz, line 57: Why you do not want to check identity?

cyberantonz:
Seems good, but why did you not use way from CONTRIBUTING.md with `make ...` ? You did mostly the same, but currently it become one more way to deploy. And we can't catch problems with `make ` such way and may catch other problems... 

If the `make` way is bad - it is ok, i'm not the devops engineer :) But i think there should be one plus or minus canonical way which not need to be synced to various places.

PR #1546, opened 2026-07-01
analytics-api: run as an idiomatic gears-rust gear

cyberantonz:
Migrate analytics-api from a self-hosted axum::serve service to a gears-rust host on toolkit::bootstrap::run_server:

- main.rs: toolkit bootstrap (AppConfig::load_or_default → run_server / gear::run_migrate); links the api-gateway REST host + auth/authz/tenant system gears via inventory.
- gear.rs (new): AnalyticsApiGear #[toolkit::gear(capabilities=[rest])]. Gear::init reconstructs the full former run_server startup (self-managed MariaDB pool + 45 migrations + CHECK/product-default probes, redis catalog cache, ClickHouse + Identity clients, schema validator, admin-CRUD service), stores AppState in a OnceLock, and tokio::spawns the one-shot validate_all() scan. RestApiCapability::register_rest wires the OperationBuilder routes into the host router + Extension(AppState).
- Auth: host runs auth_disabled (the platform api-gateway is the sole authenticator); it injects a SecurityContext with DEFAULT_TENANT_ID (== the single-tenant default the seeded data uses). The former hand-rolled tenant_middleware becomes a thin, stateless layer that reads toolkit_security::SecurityContext and overrides the tenant from X-Insight-Tenant-Id when present. The local SecurityContext struct is removed; handlers + domain services now use toolkit_security only.
- DB stays self-managed (no toolkit `db` capability; ClickHouse is not a toolkit-db backend anyway).
- config.rs: figment loader removed; GearConfig is deserialized from gears.analytics-api.config via ctx.config(). New config/insight.yaml in the gears layout (auth_disabled: true), modeled on api-gateway's no-auth.yaml.

Build, clippy (pedantic), and tests (325 pass) all green. Helm/compose manifests still reference the old ANALYTICS__* env model and are updated in a follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added a gears-based runtime for the analytics API, including `migrate` and `check` commands, with improved startup/config loading.
  * Introduced default Insight analytics configuration and host-led REST/docs/OpenAPI setup.
* **Bug Fixes**
  * Improved tenant scoping consistency across metrics, thresholds, admin actions, and queries using the request’s effective tenant.
  * Refined `X-Insight-Tenant-Id` override behavior; updated routing so health works with host auth-disabled behavior.
* **Tests**
  * Updated tenant-resolution integration tests to focus on host-injected security context override.
  * Added new end-to-end ignored-by-default HTTP live-route tests.
* **Chores**
  * Updated Helm/Docker/GitOps/secrets to use the `APP__gears__analytics_api__config__*` env-var contract.
  * Added missing `source` and `source_id` placeholder columns for seeded schema.


ktursunov (approved):

cyberantonz:
Thanks @coderabbitai — addressed in c08de153. Per-finding:

**Fixed**
- **`insight.yaml:78` — committed DB credentials.** `database_url`/`clickhouse_url` are now empty in the checked-in config; connection strings are injected per-deploy via `APP__gears__analytics_api__config__*` (umbrella Secret / compose). No credential in the runnable config.
- **`gear.rs:96` — redact redis_url.** Added `redact_url()` (strips `user:pass@`); both boot log sites now log the redacted form. (The migrate-path warn never logged the URL.)
- **`main.rs:104` — `check` was a no-op.** It now runs `gear::check_config()` which proves `gears.analytics-api.config` exists, deserializes, and has non-empty `database_url`/`clickhouse_url` — no backend connection.
- **`secrets.yaml:162` — DSN not percent-encoded.** The analytics `database_url` now `urlquery`-encodes the MariaDB user/password, matching the identity DSN.

**Already fixed earlier in this PR**
- **`compose-app-secrets.sh:131` — backticks in unquoted heredoc.** Removed in 40a1285c.

**Deferred, with reason**
- **`insight.yaml:10` + `auth.rs:55` — no-auth listener reachable in-cluster / fail-closed.** This is the intentional *header-trust* posture: the platform api-gateway is the sole authenticator and proxies to this host (`auth_disabled`), which injects a single-tenant `SecurityContext`; the `X-Insight-Tenant-Id` override is the documented bridge. Under `auth_disabled` the host always injects a context, so the `anonymous()` branch is a defensive fallback, not the live path. The correct mitigation is a network trust boundary (NetworkPolicy restricting ingress to the gateway pod / mTLS), which is a deploy-infra change — tracked as follow-up, happy to add a NetworkPolicy template if wanted.
- **`main.rs:97` — redact `--print-config`.** It's an explicit operator/debug flag with the same behavior as api-gateway's; the checked-in file now has no secrets (see above). Not adding a bespoke redactor here.
- **`compose-app-secrets.sh:138` — url-encode bash DSNs.** Pre-existing and applies equally to the identity DSN in the same script; the right fix is a shared urlencode helper across both, out of scope for this PR.
- **`deployment.yaml:18` — Secret-change reload.** In gitops the Secret is externally managed (`compose-app-secrets.sh`, `resource-policy: keep`), so a helm `checksum/secret` can't cover it; the ConfigMap already has `checksum/config`.

Also flagged as follow-up (not CodeRabbit): the gear's ConfigMap renders as `insight-analytics-api-config`, same name as the Secret (different kinds) — will give it a distinct name separately.

ktursunov (approved):

PR #1733, opened 2026-07-10
perf(analytics): cut git observation view query memory

aleksdotbar:
The `git_metric_observations` view runs live on every metric query, and its measure branches execute as concurrent pipelines — so peak memory is per-branch, not per-query. Its `FINAL` reads and cross-table joins pushed the shared ClickHouse memory tracker over its ceiling under load, aborting queries.

## View

- Pre-aggregate file changes to commit × category grain **before** the authorship join. Per-file rows and file-path strings leave the pipeline at aggregation, so they never enter a join side or a measure aggregation.
- Drop `FINAL` from the two reads whose `uniqExact` vote duplicate row versions cannot inflate. `FINAL` stays on the commit and pull-request reads, where it is the cheapest correct dedup.
- Result parity verified; dedup semantics unchanged.

## Query bounds

- Cap `max_threads` and set a per-query `max_memory_usage` in the ClickHouse client config, alongside the existing execution timeout. One enforcement point covers metric queries, validator probes, and the legacy query endpoints.
- A pathological query now fails alone with a typed error instead of exhausting the shared server tracker and killing every in-flight query on the instance.

Refs #1706.

## Summary by CodeRabbit

* **New Features**
  * Added configurable per-query limits for processing threads and memory usage.
  * Analytics queries now use defined resource limits for more predictable execution.

* **Bug Fixes**
  * Improved file-change categorization and aggregation in analytics metrics.
  * Refined pull request author attribution to avoid duplicate records affecting results.

* **Performance**
  * Optimized metric query processing to reduce unnecessary memory usage and improve reliability.

### DIFF
--- a/src/backend/libs/insight-clickhouse/src/lib.rs
+++ b/src/backend/libs/insight-clickhouse/src/lib.rs
@@ -76,6 +76,12 @@ impl Client {
         if let Some(timeout) = self.config.query_timeout {
             q = q.with_option("max_execution_time", timeout.as_secs().to_string());
         }
+        if let Some(max_threads) = self.config.query_max_threads {
+            q = q.with_option("max_threads", max_threads.to_string());
+        }
+        if let Some(max_memory_bytes) = self.config.query_max_memory_bytes {
+            q = q.with_option("max_memory_usage", max_memory_bytes.to_string());
+        }
         q
     }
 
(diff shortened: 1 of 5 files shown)


mitasovr (approved):

src/backend/libs/insight-clickhouse/src/lib.rs
```diff
@@ -159,6 +165,17 @@ mod tests {
...
 
+    #[test]
+    fn query_applies_thread_and_memory_bounds() {
+        // Both `Some` branches in `query` run.
+        let client = Client::new(
+            Config::new("http://localhost:8123", "insight")
+                .with_query_max_threads(4)
+                .with_query_max_memory_bytes(1_610_612_736),
+        );
+        let _q = client.query("SELECT 1");
```
ktursunov, line 176: Assert missing in this test.

Issue #1583, opened 2026-07-03
PLATFORM 1: EPIC: nginx + authorization — replace the API Gateway with nginx auth_request + authenticator gear


cyberantonz:
## What

Replace the planned Rust API Gateway (BFF + Router in one binary) and today's bearer-token gateway with two components:

- **nginx (OpenResty) edge** — all routing and proxying; authentication via `auth_request` subrequests; routes compiled from a reviewable `routes.yaml` by a small configurator. No hand-written proxy code.
- **`authenticator`** — a new idiomatic gear (toolkit: `#[toolkit::gear]`, OperationBuilder, CanonicalError, GTS) owning OIDC (code + PKCE), server-side sessions in Redis, and the signed gateway JWT linked 1:1 to each session.

## Target auth model

- The browser holds **only an opaque `__Host-sid` session cookie**. IdP tokens never leave the authenticator.
- At login the authenticator resolves the request author (`person_id`, tenants, roles) and mints a linked JWT. On every request nginx exchanges the cookie for that JWT and forwards **it and only it** upstream.
- **Zero trust: every downstream service verifies the gateway JWT itself — mandatory, fail-closed, no production disable knob.** A gateway misconfig can only cause a 401, never a breach. Tenant authority lives only in signed claims.
- Sessions: short TTL with mandatory refresh + absolute cap; IdP tokens are refreshed in the background — if the IdP definitively refuses, all linked sessions die.
- Services authenticate to each other with **service tokens** (public-key assertions against a reviewable registry) — one verification path for user and machine traffic.

## What we achieve

- No tokens in browser storage — kills the XSS token-theft class we have today.
- Instant server-side revocation (logout, revoke-all, back-channel, IdP-side deactivation).
- Every service receives signed, complete request-author context — no trusted headers, no network-position trust.
- We stop writing and maintaining a custom reverse proxy; nginx does the commodity part.
- One auth code path everywhere: prod, dev, e2e (fake IdP; `auth_disabled` mode is deleted).

## Notes

- There are **no production installations** — the old gateway service and its spec tree (`docs/components/backend/api-gateway/`) will be deleted outright, no migration.
- Full analysis (risks, decisions, gaps, phasing) lives in the working doc `NGINX_BFF.md`; it will be distilled into the authenticator gear PRD/DESIGN and a short nginx-edge DESIGN.
- Sub-tasks will be created here after planning. This issue intentionally carries goals, not implementation detail.

## Supersedes

[...]

PR #1735, opened 2026-07-10
nginx+auth step 05: route configurator (routegen) + OpenResty gateway

### LINKED ISSUE #1588: nginx+auth 05 — routegen configurator + OpenResty gateway (chart, Lua module, e2e)
Part of EPIC #1583 (**nginx + authorization**). Step 05 of the 11-step implementation plan.

- `routegen` CLI: routes.yaml → validated nginx.conf with the full hygiene block per generated location; golden-file tests; nginx -t in CI
- Lua module per the step-02 decision: shared-dict exchange cache (fixed-size LRU, Cache-Control-driven, 401 never cached), UUIDv7 correlation ids, RFC 9457 error shaping
- Helm chart `insight-gateway` + compose `gateway`; NetworkPolicy for the authenticator main port
- e2e: 401-everywhere, poisoned-header snapshot, cache/revocation window, 503 shaping, SPA passthrough. Depends on: 02, 04.

**Executable prompt** (self-contained instructions for a clean session): workspace file `cf/NGINX_BFF_05.md`.

Dev estimate: 1d

cyberantonz:
## nginx + authorization — Step 05 of 11: route configurator + OpenResty gateway

Implements EPIC #1583 step 05 (closes #1588). Per ADR-0001, the exchange is
**Option A — pure `access_by_lua` + `lua-resty-http`** (not `auth_request`).

### What's here
- **`routegen`** (`src/backend/tools/routegen`) — `routes.yaml → nginx.conf` with semantic validation (unique `/api/` prefixes, upstream URL sanity, `timeout_ms 0` only with `websocket`, reserved-header rejection). Golden snapshots + a rejection case per rule (16 tests).
- **Lua edge** (`src/backend/services/gateway/lua`) — access-phase cookie→JWT exchange with a `lua_shared_dict` cache + `lua-resty-http` cosocket, per-request UUIDv7 `X-Correlation-Id`, RFC 9457 problem-details error shaping.
- **Helm chart** `insight-gateway` — OpenResty Deployment, generated-config ConfigMap (resolver overridable per cluster), checksum-annotation reload, **local-only readiness** (DESIGN 3.15), non-root + read-only rootfs.
- **NetworkPolicy** in the authenticator chart — G4 layer 2 (ingress from gateway pods only; cross-chart dep noted, disabled by default).
- **compose** `gateway` service (profile `gateway`) + committed compose config.
- **e2e** (`src/backend/services/gateway/tests`) — real authenticator + fakeidp behind the gateway; all five §D scenarios green.
- **CI** `gateway.yml` — routegen fmt/clippy/tests, committed-config up-to-date assertion, `nginx -t` on the real image, `helm lint`, full e2e. `routegen` registered in the component map; service Dockerfiles updated for the new workspace member.

### Verification
- `cargo test -p routegen` → 16 passing (golden + every validation rejection).
- `nginx -t` passes on the real `openresty` image with `lua-resty-http`.
- Full e2e run green: no-cookie 401 on every route; login → verifiable ES256 JWT injected, `__Host-sid` stripped, forged `Authorization` replaced, unique UUIDv7; cache serves hits while the authenticator is down + cold cookie fails closed (503 + `Retry-After`); logout revocation within the cache window; dead upstream → 502; SPA passthrough on `/`; `/internal/*` → 404.

### Notes / deviations
- **Readiness is local-only** and there is no `/healthz/authenticator` endpoint: the authenticator is a separate Deployment with its own k8s probes, and gating the edge's readiness on a dependency would turn a blip into a fleet-wide outage (DESIGN 3.15). This diverges from the step-05 brief's "readiness proxies authenticator `/ready`" wording.
- Resolver defaults to `10.96.0.10` (kubeadm kube-dns), overridable via the chart's `resolver` value (k3s: `10.43.0.10`).

### Out of scope (later steps)
Service tokens (06), downstream JWT-verification flips (07), SPA cutover (08), deleting the legacy gateway (09).

## Summary by CodeRabbit

* **New Features**
  * Added an OpenResty gateway service (Helm chart and Docker image) that compiles a shipped `routes.yaml` into runtime config and routes analytics/identity traffic.
  * Introduced session cookie → JWT exchange and improved canonical error shaping.
* **Bug Fixes**
  * Standardized 401/503 responses for consistent fail-closed behavior.
* **New Features (CI)**
  * Added a Gateway routegen/config/e2e GitHub Actions workflow with fmt/clippy/test, lint, and end-to-end scenarios.
* **Tests**
  * Added gateway end-to-end scenarios and a new `routegen` CLI test suite (unit/golden/CLI checks).
* **Documentation**
  * Updated gateway runtime/design documentation to match the new startup compilation and edge surface.


.github/workflows/gateway.yml
```diff
@@ -0,0 +1,115 @@
+name: Gateway — routegen, config, e2e
+
+# The gateway is generated, not hand-written (DD-GW-02): routes.yaml in git, the
+# nginx.conf compiled from it by routegen and committed next to it. This workflow
+# is the guard: routegen's own tests, an assertion that every committed config is
+# in sync with its routes.yaml, `nginx -t` on the real OpenResty image, and the
+# full §D e2e. Runs only when the gateway, the generator, or the compose config
```
cyberantonz: What is §D here?

.github/workflows/gateway.yml
```diff
@@ -0,0 +1,115 @@
...
+          cargo test -p routegen
+
+      - name: Assert committed nginx.conf is in sync with routes.yaml
+        working-directory: src/backend
+        run: |
+          # k8s config (default settings)
+          cargo run -q -p routegen -- \
+            --routes services/gateway/routes.yaml \
+            --check services/gateway/helm/files/nginx.conf
+          # compose config (compose network settings)
```
cyberantonz: Docker compose is for local development only, why you build it here? The dev-compose.up should regenerate compose configuration and never commit.

deploy/compose/gateway/nginx.conf
```diff
@@ -0,0 +1,150 @@
+# Generated by routegen -- DO NOT EDIT.
+# Source of truth: routes.yaml. Regenerate with the routegen CLI; a route
+# change is a PR that re-runs it (gateway DESIGN 3.13). CI asserts this
```
cyberantonz: Put the full path to DESIGN from the sources root. Now nobody can understand what is this comment about

deploy/compose/gateway/nginx.conf
```diff
@@ -0,0 +1,150 @@
...
+            '"upstream_time":"$upstream_response_time",'
+            '"upstream_addr":"$upstream_addr",'
+            '"correlation_id":"$correlation_id"'
+        '}';
+    access_log /dev/stdout gateway_json;
+
+    # --- client-IP truth: trust only the ingress hops (G8) ---
+    real_ip_header X-Forwarded-For;
+    real_ip_recursive on;
+    set_real_ip_from 10.0.0.0/8;
```
cyberantonz: Are you sure that addresses always be in this range?

deploy/compose/gateway/nginx.conf
```diff
@@ -0,0 +1,150 @@
...
+
+    # --- OpenResty Lua runtime (DESIGN 3.11; ADR-0001 Option A) ---
+    lua_package_path "/etc/nginx/lua/?.lua;;";
+    lua_shared_dict jwt_cache 64m;
+    # cosockets resolve DNS themselves -- a resolver is mandatory (ADR-0001).
+    resolver 127.0.0.11 ipv6=off;
+
+    init_by_lua_block {
+        require("gateway").init({
+            authz_url = "http://authenticator:8083/internal/authz",
```
cyberantonz: I do not like that you bake endpoints here. Why not to use variables? You already have lua here, so example for you:
```
set_by_lua server_name 'return os.getenv("SERVER_NAME")';
```

You run this in k8s or docker, so your endpoints and settings are provided via environment as for other pods. You can send them via configmap/.env 

Or you can bake routegen inside and generate nginx.conf at startup, not in github workflow. It is pretty fast with parameters from routes.yaml.

And routes.yaml may be passed as configmap, i.e. it can be a part of chart itself and can be changed in gitops.

deploy/compose/gateway/nginx.conf
```diff
@@ -0,0 +1,150 @@
...
+    proxy_http_version 1.1;
+    # writable temp paths -> non-root + read-only rootfs
+    client_body_temp_path /tmp/client_body;
+    proxy_temp_path /tmp/proxy;
+    fastcgi_temp_path /tmp/fastcgi;
+    uwsgi_temp_path /tmp/uwsgi;
+    scgi_temp_path /tmp/scgi;
+
+    # --- upstreams (keepalive-pooled) ---
+    upstream authenticator { server authenticator:8083; keepalive 32; }
```
cyberantonz: And here we go again with baked settings.

deploy/compose/gateway/nginx.conf
```diff
@@ -0,0 +1,150 @@
...
+        # --- auth API: plain proxy, no exchange (it IS the auth) ---
+        location /auth/ {
+            limit_req zone=auth_per_ip burst=120 nodelay;
+            proxy_pass http://authenticator;
+            proxy_set_header Connection "";
+            proxy_set_header Host $host;
+            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+            proxy_set_header X-Forwarded-Proto $scheme;
+        }
+        location = /.well-known/jwks.json {
```
cyberantonz: Who from outer world will ask for this endpoint?

scripts/ci/components.py
```diff
@@ -66,6 +66,15 @@
     {"name": "fakeidp", "lang": "rust", "root": "src/backend",
      "package": "fakeidp",
      "paths": ["src/backend/services/fakeidp"]},
+    # routegen is the build-time gateway config compiler (gateway DESIGN
+    # DD-GW-02); fmt + clippy + its golden/rejection tests run here. cover=False:
```
cyberantonz: cover must be true, why you excluded it from coverage?

src/backend/services/gateway/helm/files/nginx.conf
```diff
@@ -0,0 +1,150 @@
+# Generated by routegen -- DO NOT EDIT.
+# Source of truth: routes.yaml. Regenerate with the routegen CLI; a route
```
cyberantonz: Copy of existing nginx file with baked parameters... Looks like you forgot about DRY and single source of truth.

src/backend/services/gateway/lua/errors.lua
```diff
@@ -0,0 +1,49 @@
...
+    if extra then
+        for k, v in pairs(extra) do
+            body[k] = v
+        end
+    end
+    ngx.say(cjson.encode(body))
+    return ngx.exit(status)
+end
+
+--- 401: no session / authenticator refused. Never cached upstream of here; the
```
cyberantonz: AFAIR we talked about proper canonical error format. I do not see it here.

src/backend/services/gateway/lua/uuid7.lua
```diff
@@ -0,0 +1,49 @@
...
+local resty_random = require("resty.random")
+
+local _M = {}
+
+local byte = string.byte
+local format = string.format
+local floor = math.floor
+
+--- Return a lowercase 8-4-4-4-12 UUIDv7 string.
+function _M.generate()
```
cyberantonz: `luarocks install lua-resty-uuid` and you will have uuidv7, why reinvent the wheel?

src/backend/services/gateway/lua/gateway.lua
```diff
@@ -0,0 +1,148 @@
...
+
+function _M.init(opts)
+    cfg.authz_url = opts.authz_url
+    cfg.authz_connect_timeout_ms = opts.authz_connect_timeout_ms or cfg.authz_connect_timeout_ms
+    cfg.authz_read_timeout_ms = opts.authz_read_timeout_ms or cfg.authz_read_timeout_ms
+end
+
+-- Extract the __Host-sid value from the Cookie header. The stock $cookie_*
+-- variables cannot address the dash in the name, hence the manual match.
+local function read_sid(cookie)
```
cyberantonz, line 35: A cookie name is defined as a 'token' within the parlance of the defining [RFC6265](http://www.faqs.org/rfcs/rfc6265.html). A token is defined in [RFC2616](http://www.faqs.org/rfcs/rfc2616.html)

So, __Host-sid is not good name. Rename it to __HostSid and drop this crutch

cyberantonz: The hyphen is actually fine here, and the name can't be changed without losing a security property:

- **`-` is a valid `token` char.** RFC 2616 `token` excludes only CTLs and `separators` (`()<>@,;:\"/[]?={}` + space/tab). `-` is not in that set, so `__Host-sid` is a syntactically valid cookie name. (Confirmed by the browsers accepting it and by RFC 6265bis examples.)

- **`__Host-` is a required, browser-enforced prefix, not styling.** Per the cookie-name-prefixes spec (RFC 6265bis §4.1.3), a cookie whose name starts with `__Host-` is only accepted by the browser if it's `Secure`, `Path=/`, and has **no `Domain`** attribute. That's exactly the hardening we want for the session cookie. Renaming to `__HostSid` drops the `__Host-` prefix and with it those guarantees — the browser would no longer enforce them.

- **The gateway doesn't own this name.** The cookie is minted by the authenticator (step 04); the gateway only reads it. The manual `string.match` in Lua isn't a workaround for the hyphen — with the `access_by_lua` design (ADR-0001) we parse the `Cookie` header in Lua regardless of the name, because we need the value for the shared-dict cache key and to strip it upstream. `$cookie___HostSid` wouldn't remove that.

So I'd keep `__Host-sid`. If we still want to revisit the session-cookie name, it's an authenticator (step 04) decision, not something the gateway can change unilaterally — happy to raise it there.

src/backend/tools/routegen/src/schema.rs
```diff
@@ -0,0 +1,109 @@
...
+        Self {
+            timeout_ms: default_timeout_ms(),
+            strip_prefix: false,
+            websocket: false,
+            strip_request_headers: Vec::new(),
+        }
+    }
+}
+
+fn default_timeout_ms() -> u64 {
```
cyberantonz, line 55: Why is this function and not a constant?

src/backend/tools/routegen/src/validate.rs
```diff
@@ -0,0 +1,160 @@
...
+}
+
+/// RFC 7230 `token`: one or more `tchar`s.
+fn is_valid_header_name(name: &str) -> bool {
+    !name.is_empty()
+        && name.bytes().all(|b| {
+            b.is_ascii_alphanumeric()
+                || matches!(
+                    b,
+                    b'!' | b'#'
```
cyberantonz: Why not to create set of the bytes and check is the `b` in this set? Much more clear than match

src/backend/tools/routegen/tests/fixtures/sample.nginx.conf
```diff
@@ -0,0 +1,166 @@
+# Generated by routegen -- DO NOT EDIT.
```
cyberantonz: Why you called it sample? It is fixture for tests.

src/backend/tools/routegen/tests/fixtures/full.routes.yaml
```diff
@@ -0,0 +1,24 @@
+version: 1
```
cyberantonz, line 1: And again - it is not the sample, it is fixture

src/backend/services/gateway/routes.yaml
```diff
@@ -0,0 +1,31 @@
...
+# full nginx.conf (helm/files/nginx.conf); a route change is a PR that re-runs
+# the configurator, and CI asserts the generated file stays in sync. Humans
+# never write a `location` block -- every /api route gets the full auth +
+# hygiene block emitted for it (DESIGN 3.9).
+#
+# Regenerate after editing:
+#   cargo run -p routegen -- \
+#     --routes src/backend/services/gateway/routes.yaml \
+#     -o src/backend/services/gateway/helm/files/nginx.conf
+version: 1
```
cyberantonz: Why we need versioning? It is in our repo, when and why those routes can drift from this repo stored? Is it overengineering?

docker-compose.yml
```diff
@@ -373,6 +373,38 @@ services:
     ports:
       - "${FAKEIDP_PORT:-8084}:8084"
 
+  # ── Gateway (OpenResty edge: routing + cookie-to-JWT exchange) ─────────
+  #
+  # The nginx+auth gateway (step 05). Routes /api/v1/analytics -> analytics,
```
cyberantonz: Really useless comment.

docker-compose.yml
```diff
@@ -373,6 +373,38 @@ services:
...
+  # /api/v1/identity -> identity, /auth/* + JWKS -> authenticator, / ->
+  # insight-front. The Lua module is baked into the image; the config is the
+  # committed deploy/compose/gateway/nginx.conf (generated by routegen from
+  # deploy/compose/gateway/routes.yaml with compose names + docker resolver).
+  #
+  # Profile-gated (`gateway`) so it does not clash with the legacy api-gateway
```
cyberantonz: Add todo to remove such comments and add this step to task 9

.github/workflows/gateway.yml
cyberantonz: Removed the `§D` shorthand — the workflow header now refers to "the five NGINX_BFF step-05 scenarios in run-e2e.sh". (dd547238)

.github/workflows/gateway.yml
cyberantonz: Agreed — nothing compose-related is committed anymore. The generated `nginx.conf` (k8s and compose both) is deleted; `routes.yaml` is the single source of truth and the container compiles it at startup. The e2e job here builds a throwaway *test* stack (real authenticator + fakeidp behind the gateway) purely to prove the five scenarios in CI — it's a test harness, not shipped config. (dd547238)

deploy/compose/gateway/nginx.conf
cyberantonz: Moot now — `deploy/compose/gateway/nginx.conf` is deleted (no committed config). The emitter's remaining source comment uses the full path `docs/components/backend/gateway/DESIGN.md`. (dd547238)

deploy/compose/gateway/nginx.conf
cyberantonz: Right — no baked network anymore. `set_real_ip_from` comes from `GATEWAY_SET_REAL_IP_FROM` (env), and the `real_ip_*` block is emitted only when trusted CIDRs are actually configured. (dd547238)

deploy/compose/gateway/nginx.conf
cyberantonz: Done exactly this. routegen is baked into the gateway image and the entrypoint generates `nginx.conf` at startup from the mounted `routes.yaml` + env settings (`GATEWAY_AUTHENTICATOR_URL`, `GATEWAY_FRONT_URL`, `GATEWAY_RESOLVER`, `GATEWAY_SET_REAL_IP_FROM`, …). `routes.yaml` ships as a ConfigMap, editable in gitops. (dd547238)

deploy/compose/gateway/nginx.conf
cyberantonz: Resolved by the startup-generation rework — no baked settings; all env-driven. (dd547238)

deploy/compose/gateway/nginx.conf
cyberantonz: You were right. JWKS is no longer fronted by the gateway — it's public and read-only, served directly by the authenticator (the key issuer); downstream fetch it there. The location is removed and the e2e now verifies JWKS from the authenticator directly. (dd547238)

scripts/ci/components.py
cyberantonz: Flipped to `cover: true`. Added `tests/cli.rs` that drives the built binary end to end (output mode, the `--check` up-to-date path, and the failure paths), so `main.rs` is genuinely covered — routegen is at ~93% now. (dd547238)

src/backend/services/gateway/helm/files/nginx.conf
cyberantonz: Fixed — both committed `nginx.conf` files are deleted. Single source of truth is `routes.yaml` (chart ConfigMap + compose mount); the container compiles it to `nginx.conf` at startup. (dd547238)

src/backend/services/gateway/lua/errors.lua
cyberantonz: Done. `errors.lua` now emits the toolkit `Problem` shape — `type` = `gts://gts.cf.core.errors.err.v1~cf.core.err.{unauthenticated,service_unavailable}.v1~`, plus `title`/`status`/`detail`/`context`. The e2e asserts the exact URNs and the `context` (`retry_after_seconds`). (dd547238)

src/backend/services/gateway/lua/uuid7.lua
cyberantonz: I checked, and there is no opm/luarocks library that produces **UUIDv7** (RFC 9562, 2024):
- `thibaultcha/lua-resty-jit-uuid` → v3/v4/v5 only
- `bungle/lua-resty-uuid` → libuuid FFI (v1/v4)

v4 would forfeit the time-ordering the DESIGN wants for correlation ids, so I kept the ~40-line v7 generator and documented that rationale in the file header. If you know a maintained v7 lib I'll happily swap to it. (kept in dd547238)

src/backend/tools/routegen/src/schema.rs
cyberantonz: serde's `#[serde(default = "...")]` takes a function *path*, not a const, so it can't point straight at a constant. I introduced `pub const DEFAULT_TIMEOUT_MS: u64 = 30_000;` as the real source and left the fn as a one-line wrapper returning it (used by both the serde attribute and the `Default` impl). (d9f50f0c)

src/backend/tools/routegen/src/validate.rs
cyberantonz: Done — replaced the `matches!` with a `TCHAR_SYMBOLS` byte-set and a membership check:
```rust
const TCHAR_SYMBOLS: &[u8] = b"!#$%&'*+-.^_`|~";
... b.is_ascii_alphanumeric() || TCHAR_SYMBOLS.contains(&b)
```
(d9f50f0c)

src/backend/tools/routegen/tests/fixtures/sample.nginx.conf
cyberantonz: Renamed `sample.nginx.conf` → `full.nginx.conf` (and the routes file too). (d9f50f0c)

src/backend/tools/routegen/tests/fixtures/full.routes.yaml
cyberantonz: Renamed `sample.routes.yaml` → `full.routes.yaml` — agreed, "sample" was redundant under `tests/fixtures/`. (d9f50f0c)

src/backend/services/gateway/routes.yaml
cyberantonz: Fair question. `version` is the salvaged Router-schema field (DESIGN 3.8). The reason to keep it: `routes.yaml` ships as a ConfigMap edited in gitops **independently** of the routegen binary, so the file and the tool can drift (someone edits the ConfigMap, or an image is rolled back to an older schema). The version lets routegen fail fast on a schema it doesn't understand instead of silently mis-parsing — it's one field + one check.

That said, if you consider it YAGNI for v1 I'll drop it (field + validation rule + the DESIGN 3.8 line). Your call — leaving it in for now.

docker-compose.yml
cyberantonz: Trimmed to just what the service does. (d9f50f0c)

docker-compose.yml
cyberantonz: Added: `TODO(#1583 step 09 — Destroy): once the legacy api-gateway is deleted this takes 8080 unconditionally; drop the `gateway` profile and this note.` The step-09 (Destroy) task in the EPIC already owns deleting the legacy gateway. (d9f50f0c)

src/backend/services/gateway/routes.yaml
cyberantonz: It was before you move this to configmap. Ok, plausible.

docs/components/backend/gateway/DESIGN.md
```diff
@@ -204,7 +197,7 @@ Does not run in the pod or watch anything at runtime. Does not decide routing po
 The commodity edge: routing, proxying, streaming, WebSocket upgrades, timeouts, rate limiting -- the code nobody should hand-write in Rust.
 
 ##### Responsibility scope
-Longest-prefix `location` matching; `auth_request` on every `/api/` location; plain proxy for `/auth/` and JWKS (with a coarse per-IP `limit_req` flood guard on `/auth/`); SPA at `/`; `proxy_buffering off` where streaming matters; WebSocket upgrade pass-through (auth runs once at upgrade, JWT frozen for the socket's life); `/internal/` returns 404; HSTS on every response.
+Longest-prefix `location` matching; an access-phase Lua exchange on every `/api/` location; plain proxy for `/auth/` (with a coarse per-IP `limit_req` flood guard); SPA at `/`; `proxy_buffering off` where streaming matters; WebSocket upgrade pass-through (auth runs once at upgrade, JWT frozen for the socket's life); `/internal/` returns 404; HSTS on every response.
```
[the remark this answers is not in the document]
cyberantonz: Fixed in 5d1c7429 — the paragraph now describes the access-phase Lua exchange (allow on `200` with JWT from `X-Gateway-Jwt`, `401` deny/never-cached, anything else → shaped `503`; cosocket sends no body). No more `auth_request`/`auth_request_set` wording.

src/backend/services/gateway/docker-entrypoint.sh
```diff
@@ -0,0 +1,51 @@
...
+    exit 1
+fi
+
+set -- --routes "$ROUTES" -o "$OUT"
+[ -n "${GATEWAY_AUTHENTICATOR_URL:-}" ] && set -- "$@" --authenticator-url "$GATEWAY_AUTHENTICATOR_URL"
+[ -n "${GATEWAY_FRONT_URL:-}" ]         && set -- "$@" --front-url "$GATEWAY_FRONT_URL"
+[ -n "${GATEWAY_RESOLVER:-}" ]          && set -- "$@" --resolver "$GATEWAY_RESOLVER"
+[ -n "${GATEWAY_LISTEN:-}" ]            && set -- "$@" --listen "$GATEWAY_LISTEN"
+[ -n "${GATEWAY_JWT_CACHE_SIZE:-}" ]    && set -- "$@" --jwt-cache-size "$GATEWAY_JWT_CACHE_SIZE"
+[ -n "${GATEWAY_AUTHZ_PATH:-}" ]        && set -- "$@" --authz-path "$GATEWAY_AUTHZ_PATH"
```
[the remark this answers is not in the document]
cyberantonz: Good catch — wired `GATEWAY_AUTHZ_CONNECT_TIMEOUT_MS` and `GATEWAY_AUTHZ_READ_TIMEOUT_MS` through to `--authz-connect-timeout-ms` / `--authz-read-timeout-ms` in the entrypoint (5d1c7429). Tunable per deployment now.

src/backend/services/gateway/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+# routegen binary + the startup entrypoint.
+COPY services/gateway/lua/ /etc/nginx/lua/
+COPY --from=routegen /build/target/release/routegen /usr/local/bin/routegen
+COPY services/gateway/docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
+RUN chmod +x /usr/local/bin/docker-entrypoint.sh
+
+EXPOSE 8080
+
+STOPSIGNAL SIGQUIT
+ENTRYPOINT ["/usr/local/bin/docker-entrypoint.sh"]
```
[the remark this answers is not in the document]
cyberantonz: Added `USER 1000` (5d1c7429) — matches the pod securityContext so bare `docker run` / compose isn't root either. Verified the container still runs (uid=1000): the entrypoint only writes `/tmp` (config, pid, temp paths) and reads the read-only routes mount, and the e2e suite passes as non-root.

src/backend/tools/routegen/src/main.rs
```diff
@@ -0,0 +1,132 @@
...
+    output: Option<PathBuf>,
+
+    /// Assert the file at this path already matches the generated config, then
+    /// exit non-zero if it drifted. Used by CI to keep the committed config in
+    /// sync with `routes.yaml`.
+    #[arg(long, conflicts_with = "output")]
+    check: Option<PathBuf>,
+
+    /// Base URL of the authenticator (host:port used for the /auth + JWKS
+    /// upstream and the Lua exchange subrequest).
```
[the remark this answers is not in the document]
cyberantonz: Fixed — the `--authenticator-url` help no longer mentions JWKS (5d1c7429).

docs/components/backend/gateway/DESIGN.md
```diff
@@ -96,7 +96,7 @@ The gateway is a **separate service, not bound to the ingress layer at all**. Th
 graph LR
     B[Browser] -->|HTTPS| ING["LB / ingress controller (any)<br/>TLS termination + host routing only"]
     ING --> GW["nginx gateway (Deployment)<br/>auth, paths, headers, HSTS"]
-    GW -->|/auth/*, JWKS, subrequests| AUTH[authenticator]
+    GW -->|/auth/*, exchange subrequest| AUTH[authenticator]
```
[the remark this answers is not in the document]
cyberantonz: Fixed — the 1.3 diagram edge now reads `access_by_lua exchange` instead of "subrequest" (5d1c7429).

src/backend/services/authenticator/helm/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,39 @@
...
+  ingress:
+    - from:
+        - podSelector:
+            matchLabels:
+              {{- toYaml .Values.networkPolicy.gatewayPodSelector | nindent 14 }}
+          {{- with .Values.networkPolicy.gatewayNamespaceSelector }}
+          namespaceSelector:
+            matchLabels:
+              {{- toYaml . | nindent 14 }}
+          {{- end }}
```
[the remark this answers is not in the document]
cyberantonz: Declining this one — it rests on a NetworkPolicy-semantics misread. A `from` peer that has a `podSelector` and **no** `namespaceSelector` selects matching pods **in the policy's own namespace only**, not all namespaces (Kubernetes NetworkPolicyPeer semantics). So omitting `gatewayNamespaceSelector` is the *tighter* default (same-namespace gateway pods only), which is exactly the intent — adding a `fail` gate would instead break that correct default. The optional `gatewayNamespaceSelector` only exists to *widen* the rule to a labelled namespace when a deployment genuinely runs the gateway cross-namespace. The values.yaml comment already states "Empty = same-namespace pods only". No change.

src/backend/services/gateway/lua/uuid7.lua
cyberantonz: You are wrong.
```
lua-resty-uuid

LuaJIT FFI bindings for [libuuid](https://man7.org/linux/man-pages/man3/libuuid.3.html) — a DCE-compatible Universally Unique Identifier (UUID/GUID) library. Generates v1 (time), v4 (random), v6 (reordered time) and v7 (Unix-epoch) UUIDs, plus parses and inspects existing UUID strings.
Status

Production-ready. Used in OpenResty / LuaJIT environments. Pure FFI bindings — no C build step required, just libuuid on the system.
```

Function `uuid.generate_time_v7()`

cyberantonz: You're right, and thanks for the pointer — I'd only checked `bungle/lua-resty-uuid` at the API level and wrongly concluded it was v1/v4 only. It does expose `generate_time_v7()`, backed by libuuid's `uuid_generate_time_v7` (alpine's libuuid is 2.41, which has it).

Swapped in 7b78f93e:
- Dockerfile: `opm get bungle/lua-resty-uuid` + `apk add libuuid`, and a `libuuid.so -> libuuid.so.1` symlink (the FFI loader looks for the unversioned name; alpine ships only `.so.1`).
- `gateway.lua`: `require("resty.uuid").generate_time_v7()`.
- Deleted the hand-rolled `lua/uuid7.lua`.
- e2e now also asserts the correlation id matches the UUIDv7 shape (`...-7xxx-[89ab]xxx-...`), which passes against the real values.

src/backend/tools/routegen/src/schema.rs
cyberantonz: omg...

.github/workflows/gateway.yml
```diff
@@ -0,0 +1,101 @@
...
+          # makes the container generate, validate, and exit without serving.
+          docker run --rm \
+            --add-host analytics:127.0.0.1 \
+            --add-host identity:127.0.0.1 \
+            --add-host authenticator:127.0.0.1 \
+            --add-host insight-front:127.0.0.1 \
+            -e GATEWAY_VALIDATE_ONLY=1 \
+            -e GATEWAY_AUTHENTICATOR_URL=http://authenticator:8083 \
+            -e GATEWAY_FRONT_URL=http://insight-front:8080 \
+            -e GATEWAY_RESOLVER=10.96.0.10 \
```
cyberantonz: Is this address exists in docker you run?

Anyway, the idea of this resolver looks a bit weird. 
You can have the next lines in nginx.conf
```
http {
    resolver local=/etc/resolv.conf; 
}
```

And the problem is solved. Isn't it?

src/backend/services/gateway/helm/values.yaml
```diff
@@ -0,0 +1,61 @@
...
+  port: 8080
+
+# Deployment-specific settings, passed to the container as env vars; the
+# entrypoint feeds them to routegen at startup (the route table itself is the
+# mounted ConfigMap). Nothing is baked into the image.
+gateway:
+  # Authenticator base URL for the /auth proxy and the Lua exchange cosocket.
+  # Use the FQDN: the cosocket's DNS (lua-resty-http) does NOT apply the k8s
+  # search domain, unlike nginx's own upstream resolution.
+  authenticatorUrl: http://authenticator.insight.svc.cluster.local:8083
```
cyberantonz: Isn't those addresses be generated ?

frontUrl, authenticatorUrl

src/backend/services/gateway/helm/values.yaml
```diff
@@ -0,0 +1,61 @@
...
+  # insight-front SPA server.
+  frontUrl: http://insight-front.insight.svc.cluster.local:8080
+  # kube-dns / CoreDNS cluster IP -- mandatory for the Lua cosocket (ADR-0001).
+  # kubeadm defaults to 10.96.0.10; k3s uses 10.43.0.10.
+  resolver: 10.96.0.10
```
cyberantonz: So, if you have various resolvers here for various clusters, how you can be sure what to use?

src/backend/services/gateway/tests/assert.py
```diff
@@ -0,0 +1,195 @@
...
+def phase_upstreamdown():
+    # Scenario 4: exchange succeeds (authenticator up) but the upstream is dead.
+    sid = login()
+    status, _, _ = req(f"{GW}/api/v1/analytics/x", headers={"Cookie": f"__Host-sid={sid}"})
+    check("dead upstream -> 502", status == 502, f"got {status}")
+
+
+PHASES = {"core": phase_core, "authdown": phase_authdown, "revoked": phase_revoked, "upstreamdown": phase_upstreamdown}
+
+if __name__ == "__main__":
```
cyberantonz: why tests not use pytest?

.github/workflows/gateway.yml
cyberantonz: You're right on both counts. Switched to `resolver local=on ipv6=off;` — OpenResty reads the nameservers from `/etc/resolv.conf`, so the exact same config resolves in-cluster (kube-dns) and in compose (docker DNS). It's now a hardcoded constant in the emitted config — no flag, no env, no per-cluster IP. Verified a cosocket resolves a service by name under `local=on`. `GATEWAY_RESOLVER` is gone from the entrypoint, compose, and this CI step. (c4e9c184)

src/backend/services/gateway/helm/values.yaml
cyberantonz: Done — `deployment.yaml` now assembles the URLs from structured values + the release namespace instead of taking a full URL:
```
value: {{ printf "http://%s.%s.svc.%s:%v" .Values.gateway.authenticator.service .Release.Namespace .Values.gateway.clusterDomain .Values.gateway.authenticator.port | quote }}
```
values just hold `service` / `port` / `clusterDomain`. A default install sets nothing; override a single piece (e.g. `gateway.authenticator.port`) if needed. (c4e9c184)

src/backend/services/gateway/helm/values.yaml
cyberantonz: Resolved by the `resolver local=on` switch above — there's no per-cluster resolver value anymore. `local=on` reads `/etc/resolv.conf`, which already points at the right DNS in every environment, so nothing to pick. (c4e9c184)

src/backend/services/gateway/tests/assert.py
cyberantonz: Deliberate, but happy to change it. The driver runs **inside** a vanilla `python:3.12-slim` container with no `pip install` (offline, fast, zero deps), and `run-e2e.sh` orchestrates the phases by stopping/starting containers between them (authenticator down, upstream down) with the session shared via `/work/sid.txt` — that container-lifecycle-between-phases flow isn't a natural pytest fixture model, and it's driven from the shell, not from Python.

If you'd prefer pytest for consistency with the bronze-to-api suite, I can add it to the driver image and restructure the phases as tests (with the lifecycle in fixtures). Want me to?

src/backend/services/gateway/helm/files/routes.yaml
```diff
@@ -0,0 +1,30 @@
...
+  # Operator-extensible deny-list. The gateway-reserved set (Authorization,
+  # X-Correlation-Id, X-Forwarded-*, gateway cookies) is always stripped in
+  # addition; X-Tenant-ID is NOT reserved and passes through as the tenant
+  # selector the downstream middleware validates against the signed tenants[].
+  strip_request_headers:
+    - X-Real-IP
+    - Forwarded
+routes:
+  - prefix: /api/v1/analytics
+    upstream: http://analytics:8081
```
cyberantonz: And again hardcoded values. 
Why not to use helm templating like it was before???

```
    routes:
      - prefix: "/analytics"
        # Templated — `tpl` evaluates this at install time so `.Release.Name`
        # resolves to the umbrella's release name. Override with a literal
        # URL (no template syntax) to point at an external service.
        upstream: 'http://{{ .Release.Name }}-analytics:8081'
        public: false
      - prefix: "/identity"
        upstream: 'http://{{ .Release.Name }}-identity:8082'
        public: false
```

You build yaml for configmap, so you can fill this from values with proper FULL name of service and port.

src/backend/services/gateway/tests/assert.py
cyberantonz: Converted to pytest (525e2795) — you're right, no reason to drift.

- `conftest.py` owns the compose lifecycle and the fail-closed fixtures (`session_sid`, `authenticator_down`, `echo_down`); `test_gateway.py` holds the five scenarios as plain `test_*` functions/asserts; `pytest.ini` for config. Same "conftest owns the stack" shape as the bronze-to-api suite.
- `assert.py` + the phase-orchestration shell are gone. `run-e2e.sh` is now just `python -m pytest`, and the e2e CI job drops the Rust toolchain (images build routegen inside Docker) and only installs pytest.
- pytest runs on the host; the OIDC redirect chain's in-network hostnames (`gateway:8080`, `fakeidp:8084`) are rewritten to the published localhost ports in the client helper.

All 11 tests green locally. One fix fell out of it: added `proxy_connect_timeout 5s` to the generated `/api` locations so a dead upstream fails fast (502/504) instead of hanging on nginx's default 60s connect timeout.

src/backend/services/gateway/helm/files/routes.yaml
cyberantonz: Good catch — and it was actually a correctness bug, not just style: k8s services are `<release>-<name>` (e.g. `insight-analytics`), so the hardcoded short `http://analytics:8081` wouldn't resolve in-cluster. Switched to the umbrella's existing convention (4ce25f3c):

- The route table lives in `values.yaml` (`gateway.routes`) with templated upstreams, and `configmap.yaml` renders them via `tpl`, so `{{ .Release.Name }}` resolves. Rendered ConfigMap:
  ```yaml
  routes:
    - prefix: /api/v1/analytics
      upstream: "http://insight-analytics:8081"
    - prefix: /api/v1/identity
      upstream: "http://insight-identity:8082"
  ```
- Same for the exchange/SPA URLs in `deployment.yaml`: `http://{{ .Release.Name }}-authenticator.{{ .Release.Namespace }}.svc.{{ .Values.gateway.clusterDomain }}:8083` (FQDN — the Lua cosocket doesn't apply the search domain) and `http://{{ .Release.Name }}-frontend:80`.
- Removed the static `helm/files/routes.yaml`. Compose keeps its own small `deploy/compose/gateway/routes.yaml` (docker service names) since compose can't render Helm templates.
- CI now `nginx -t`s both the compose routes and the chart's *rendered* `routes.yaml` (via `helm template`) through the container's generate-and-validate path.

ktursunov (approved):

PR #1746, opened 2026-07-13
fix(zoom): derive start date from sync time; make manifest strictly valid

### LINKED ISSUE #1738: data: Zoom meeting metrics silently flatline — connector's daily sync has failed 10 days straight
## Summary
Zoom meeting metrics (`Zoom Meeting Hours`, `Zoom Meetings Attended`) have been stuck at zero for 10 days. The Zoom connector's daily sync has failed every day since 2026-07-01.

## Impact
Anyone whose meetings are mostly on Zoom now shows as not meeting at all. Blended `Meeting Hours` / `Meetings Attended` (Teams+Zoom) are also understated for these people. No error is shown anywhere — the dashboard just looks flat.

## Steps to Reproduce
1. Check Zoom bronze freshness vs. a healthy connector:
   ```bash
   scripts/ch.sh query --target dev "SELECT max(_airbyte_extracted_at) FROM bronze_zoom.meetings"
   # → 2026-06-30 (stale)
   scripts/ch.sh query --target dev "SELECT max(date) FROM silver.class_collab_meeting_activity WHERE data_source='insight_m365'"
   # → 2026-07-06 (healthy, same table)
   ```
2. Check the Zoom connection's Airbyte job history → last 10 daily syncs (jobs 381–492, 2026-07-01 → 2026-07-10) all fail at the `check` step with:
   ```
   'Stream meetings is not available: Bad request. Please check your request parameters.'
   ```

**Expected:** Zoom rows keep arriving daily, same as every other connector.
**Actual:** No new Zoom data since 2026-06-30.

## Root Cause
The Zoom connector's `check` step calls Zoom's `GET /v2/metrics/meetings` ([connector.yaml#L5-L8](https://github.com/constructorfabric/insight/blob/main/src/ingestion/connectors/collaboration/zoom/connector.yaml#L5-L8), [#L312](https://github.com/constructorfabric/insight/blob/main/src/ingestion/connectors/collaboration/zoom/connector.yaml#L312)), which now returns `400 Bad Request` on every attempt. No connector code changed around this date — this points to something on the Zoom account/API side, not a regression in this repo.

Silver → gold is confirmed fine: `insight.zoom_person_daily` and `insight.collab_bullet_rows` just have no new rows to aggregate, so they correctly show zero. The bug is entirely upstream, in the connector sync.

## Notes
No alerting exists for a connector failing its daily sync — related to #1319.

mitasovr:
## Problem

Zoom syncs on dev-vhc fail at the pre-sync connection check (e.g. Airbyte job 529):

```
'GET' https://api.zoom.us/v2/metrics/meetings?type=past&from=2026-01-01&to=2026-01-30 → 400
{'code': 300, 'message': 'The request can only be queried for a month that falls within the last six months.'}
```

The Zoom Dashboard API (`/v2/metrics/*`) only serves the last six months. The `meetings` stream built its first slice from the static `zoom_start_date` config (2026-01-01), and the Airbyte `check` command receives no state — so once the configured date fell out of the window, every sync died at check, even though the connection's saved cursor (`end_time: 2026-06-29`) was perfectly valid.

## Fix

- `meetings.start_datetime` is now computed: `now_utc() - P150D` (~5 months — a safety margin inside Zoom's window, whose month-granularity semantics are ambiguous at the boundary). Same pattern as m365.
- `zoom_start_date` removed from the connector spec, descriptor `required_fields`, secret example, and README. Existing K8s secrets that still carry the field stay valid (`additionalProperties: true`), so no secret rotation is needed.
- Descriptor version bumped to 1.1.0 so reconcile republishes the manifest (equal versions → noop).

## Also: Builder-UI strict validation

`validate-strict` failed on main with 3 errors (and more masked behind them — the validator surfaces one leaf error per stream). The strict validator performs no $ref resolution, so any typed-object slot holding a $ref fails. Fixed by inlining:

- `request_body` in both `login_requester`s,
- `error_handler` / `paginator` in `users`,
- `authenticator` in `participants`,
- the whole-object parent `$ref: "#/streams/1"` on `participants` replaced with an inline `_meetings` parent stream (youtrack pattern).

Note: `collaboration/m365` currently fails `validate-strict` with the same `login_requester.request_body` $ref pattern (4 streams) — left out of scope here.

## Verification

- `source.sh validate-strict collaboration/zoom` — strictly valid
- `source.sh validate collaboration/zoom` — manifest valid
- `source.sh check` with real dev-vhc credentials — Check succeeded (first slice now inside the window)
- Full `source.sh read` with real credentials — **meetings 24 525, users 375, participants 77 085** records; the inline `_meetings` parent routes partitions correctly. (One transient 400 mid-pagination of a 30-day slice — Zoom's `next_page_token` 15-min expiry during the 150-day cold read; pagination behavior unchanged from main.)

Existing connection state is untouched; syncs continue from their saved cursor.

## Summary by CodeRabbit

- **Updates**
  - Simplified Zoom connector setup by removing the `zoom_start_date` configuration requirement, and updated the connector version, documentation, and credential secret example accordingly.
  - Zoom meetings now automatically backfill a recent window before switching to incremental sync from saved progress.

- **Improvements**
  - Enhanced Zoom API sync reliability with improved retry/backoff and explicit pagination handling.
  - Refined Zoom bronze placeholder behavior to better preserve row uniqueness during merges.

- **Tests**
  - Added/expanded end-to-end metric tests for Zoom and Teams+Zoom cross-source behavior, plus updated related fixture schemas/templates.

- **Chores**
  - Updated e2e session startup truncation and metric coverage skip logic.


ktursunov (changes requested):
Zoom metrics are not covered with tests

src/ingestion/connectors/collaboration/zoom/connector.yaml
```diff
@@ -105,18 +105,62 @@ streams:
...
+        page_token_option:
+          type: RequestOption
+          inject_into: request_parameter
+          field_name: next_page_token
+        pagination_strategy:
+          type: CursorPagination
+          cursor_value: "{{ response.get('next_page_token') }}"
+        stop_condition:
+          type: StopCondition
+          condition: "{{ not response.get('next_page_token') }}"
```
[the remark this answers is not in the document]
mitasovr: Applied in d0ad879d — the `users` requester now sends `request_parameters: page_size`.

One correction to the finding: this was not a regression from the inlining. On main, only `meetings` and `participants` referenced the shared `request_parameters` $ref; the `users` requester never had `request_parameters` at all, so `page_size` was never applied to this stream. Added it now as an improvement (375 users previously paged with Zoom's default page size of 30). Verified with a live read: 374 users.

src/ingestion/connectors/collaboration/zoom/connector.yaml
```diff
@@ -522,13 +569,155 @@ streams:
...
+                  type: RequestOption
+                  field_name: from
+                  inject_into: request_parameter
+                end_time_option:
+                  type: RequestOption
+                  field_name: to
+                  inject_into: request_parameter
+                lookback_window: P7D
+                step: P30D
+                cursor_granularity: P1D
```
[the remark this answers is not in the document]
mitasovr: Applied in d0ad879d — `incremental_dependency: true` set on the `participants` ParentStreamConfig.

Note this was pre-existing behavior, not a regression from this PR: the old whole-object `$ref: "#/streams/1"` parent had no `incremental_dependency` either, so the substream has always re-read the full parent window to enumerate partitions. With the flag, SubstreamPartitionRouter persists the `_meetings` cursor into the substream state and restores it on the next sync (verified the parent-state set/get/migrate mechanics in the CDK 6.60.9 sources), so partition enumeration advances incrementally instead of re-reading 150 days each sync. Semantics caveat: participants for meetings older than the saved parent cursor (minus the P7D lookback) are no longer re-fetched — acceptable, since participant lists of long-closed meetings don't change. Verified with a live full read: 77093 participants.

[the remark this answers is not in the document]
mitasovr: Correction to my previous reply: the new mock-server suite for this connector (12b057a8) disproved it. `incremental_dependency: true` is a **silent no-op on a full-refresh substream** — the parent cursor only piggybacks on an *incremental* child's state, and `participants` emits only the `__ab_no_cursor_state_message` sentinel (no `parent_state`); the earlier live read shows the same. The flag is removed again with an explanatory comment in the manifest, and `tests/test_participants.py::test_full_refresh_substream_emits_no_cursor_state` pins the actual contract. Making the participants fan-out truly incremental would require giving the child stream its own cursor — a semantics change out of scope for this PR.

src/ingestion/connectors/collaboration/zoom/tests/test_meetings.py
```diff
@@ -0,0 +1,177 @@
...
+at now-150d (inside the window) and no request predates it: with no network
+fallthrough, a connector regressing to an out-of-window start date fails this
+test with an unmatched-request error.
+"""
+
+from __future__ import annotations
+
+import json
+
+import freezegun
```
cyberantonz, line 29: Do you have pre commit installed? The headers are not sorted as i see.

ktursunov (approved):

src/ingestion/connectors/collaboration/zoom/tests/conftest.py
```diff
@@ -0,0 +1,7 @@
+import sys
+from pathlib import Path
+
+# Local builder modules (config.py) are importable under --import-mode=importlib.
+sys.path.insert(0, str(Path(__file__).parent))
```
cyberantonz, line 5: Why claude likes to modify syspath? I think there is easier import fix like `from . import`...

mitasovr: seems like copypaste
will fix it later

cyberantonz: Now it is later :) Still not fixed

cyberantonz (approved):

PR #1751, opened 2026-07-13
feat(authenticator): nginx+auth step 06 — service tokens + internal listener

cyberantonz:
Step 06 of 11 for the nginx + authorization EPIC (#1583): service tokens (§10 G1 / DD-AUTH-05) on a dedicated internal listener (§10 G4 / §11.8).

## What
- **`POST /internal/token`** (`src/service_token.rs`): verifies an RFC 7523 `private_key_jwt` client assertion against a gitops-reviewable public-key **registry** (two keys/service for rotation), checking `aud`, `exp`/`iat`, `sub == iss == service`, and `exp − iat ≤ cap`. Replay guard via Redis `SET NX EX` at `asm:svc_jti:{service}:{jti}` (the `logout_jti` pattern; `asm:` is the gear's real key prefix). Issues a **normal gateway JWT** — `sub = sid = service:<name>`, `roles` always including `"service"`, `tenants` only when requested **and** allowed — signed with the same ES256 key and published in the same JWKS. Every issuance is audited.
- **Second listener**: bound in `RunnableCapability::start` on `token_bind_addr` (default `8093`), serving only `POST /internal/token` + `/ready`, excluded from the public OpenAPI, graceful shutdown via the toolkit cancellation token.
- **Network scopes**: token port added to the Service + deployment; a second NetworkPolicy (`allow-token-clients`) scopes it to app-namespace service pods; main-port policy unchanged. Compose exposes 8093.
- **SDK** (`authenticator-sdk`): `ServiceTokenClient` — mints assertions, fetches + caches the token, reissues at 4/5 TTL, exposes `async fn bearer()`.
- **Dev/test**: checked-in **test-only** ES256 keypair + `testclient`/`svc-noscope` registry entries; e2e loop proving token→JWKS verify, replay rejected, wrong-key rejected, tenant honored (testclient) / refused (svc-noscope). Also fixes a `run-e2e.sh` workspace-root `cd` bug.

## Docs
DESIGN updated: concrete registry format (§3.9), `sid = service:<name>` choice (§3.8), `asm:svc_jti` key (§3.7). `cfs validate` clean on the authenticator tree.

## Testing
26 unit tests + both e2e loops (login + service-token) green against a live Redis + authenticator stack; `pre-commit` (cargo fmt/clippy/yamlfmt) passes.

## Out of scope
Flipping downstream verification on (step 07); the permissions service (future epic).

Refs #1583.

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated `/internal/token` listener with new token port support in Docker Compose and Helm/Kubernetes, including optional network-policy restrictions.
  * Implemented RFC 7523 service-token minting/exchange for gateway JWTs with stable service identity, role injection, and tenant-scoped claims.
  * Added authenticator SDK support for creating service-token assertions, fetching/exchanging tokens, and caching bearer tokens by tenant scope.
* **Security**
  * Enforced mandatory tenant scoping, strengthened validation rules, and introduced one-time replay protection for assertions.
* **Documentation**
  * Updated authenticator design docs and configuration references for the service-token registry and JWT claim contract.
* **Tests**
  * Added an ignored-by-default end-to-end service-token loop test plus dev/e2e key and port setup updates.


docs/components/backend/authenticator/DESIGN.md
```diff
@@ -837,6 +841,40 @@ Inherited from the deleted BFF spec unchanged: `authenticator.refresh_grace_ms`
...
+  token_ttl_seconds: 300
+  services:
+    <service-name>:
+      public_keys:            # 1..2 SPKI PEM EC public keys (two during rotation)
+        - |
+          -----BEGIN PUBLIC KEY-----
+          ...
+          -----END PUBLIC KEY-----
+      roles: ["service"]      # baked into the token; "service" is always added
+      tenant_scoped_allowed: false   # may this service request tenants=[t]?
```
cyberantonz: the service can not request anything without tenant, where it get tenant - it is the problem of the service. :)

src/backend/libs/authenticator-sdk/src/service_client.rs
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    /// The cross-tenant service bearer, ready for an `Authorization` header
+    /// (`"Bearer <jwt>"`). Served from cache until 4/5 of its TTL has elapsed,
+    /// then re-fetched ahead of expiry. Concurrent callers share one fetch.
+    ///
+    /// # Errors
+    /// As [`fetch`](Self::fetch), when a refresh is needed and fails.
+    pub async fn bearer(&self) -> Result<String, CanonicalError> {
+        let mut cache = self.cache.lock().await;
```
cyberantonz: Looks like you do request at line 231 with lock held. From my point of view - you are an idiot. Locks must be released right after you get information from cache. 

And when you want to write to the cache - get lock again.

And more - there is perfect thing called read-write-lock, your cache is read mostly storage.

src/backend/libs/authenticator-sdk/Cargo.toml
```diff
@@ -15,3 +15,15 @@ workspace = true
...
+# token, exposes `bearer()`. Consumers (analytics/background jobs, step 07)
+# depend on this crate to obtain a gateway JWT without a user request.
+serde = { workspace = true }
+serde_json = { workspace = true }
+reqwest = { workspace = true }
+tokio = { workspace = true }
+uuid = { workspace = true }
+# ES256 assertion signing. Match the authenticator's crypto provider (aws_lc_rs)
+# — v10 needs exactly one provider process-wide under feature unification.
+jsonwebtoken = { version = "10", default-features = false, features = ["aws_lc_rs", "use_pem"] }
```
cyberantonz: btw fakeidp use version 9 instead of proper string you put here. It is worth to fix.

src/backend/services/authenticator/config/insight.yaml
```diff
@@ -92,3 +92,37 @@ gears:
...
+      # checked-in keypair in dev/service-tokens/ (clearly marked test-only).
+      # Real services land their public key here via a gitops PR; prod's chart
+      # ConfigMap carries its own registry, never these test entries.
+      service_tokens:
+        token_bind_addr: "0.0.0.0:8093"
+        audience: "http://localhost:8093/internal/token"
+        assertion_max_lifetime_seconds: 60
+        token_ttl_seconds: 300
+        services:
+          testclient:
```
cyberantonz, line 113: Do we really have those services in k8s?

src/backend/services/authenticator/dev/service-tokens/README.md
```diff
@@ -0,0 +1,21 @@
+# Dev service-token keypair — TEST ONLY, DO NOT USE IN PRODUCTION
+
+These are a throwaway ES256 (EC P-256) keypair for local development and e2e of
```
cyberantonz: And you again fit the same pattern.... if the keys are throwaway - YOU CAN GENERATE THEM. AND YOU ALREADY DO THIS FOR DOCKER COMPOSE.

Maybe you can just use the same keys? It is development only part.

src/backend/services/authenticator/helm/values.yaml
```diff
@@ -73,3 +90,17 @@ networkPolicy:
...
   # same-namespace pods only (the default NetworkPolicy scope).
   gatewayNamespaceSelector: {}
+
+  # G4 layer 2 for the TOKEN port: /internal/token must be reachable by the
+  # app-namespace service pods that need service tokens (analytics, seeds, the
+  # future permissions service) — a wider scope than the main port, which is
+  # fine because the endpoint demands a valid RFC 7523 assertion (layer 3).
+  # This policy is defense-in-depth, not the authentication.
+  tokenPort:
+    enabled: false
```
cyberantonz, line 103: The comment is misleading. Why not to write
```
The access to token port is disabled by default, enable it in your gitops config
```
And no so much 'professionally looking' comments.

src/backend/services/authenticator/src/config.rs
```diff
@@ -55,6 +57,65 @@ impl Default for IdpConfig {
...
+pub struct ServiceRegistryEntry {
+    /// SPKI PEM public key(s) the service signs its RFC 7523 assertions with.
+    /// Two keys are allowed at once so a rotation overlaps `previous`+`next`.
+    pub public_keys: Vec<String>,
+    /// Roles baked into the issued gateway JWT. `"service"` is always added by
+    /// the issuer, so an entry may leave this empty for a plain service token.
+    pub roles: Vec<String>,
+    /// Whether this service may request a tenant-scoped token (`tenants=[t]`).
+    /// A tenant request from a service without this flag is refused (403).
+    pub tenant_scoped_allowed: bool,
```
cyberantonz: I suggest drop this and use true and only true. We need tenant isolation.

src/backend/services/authenticator/src/service_token.rs
```diff
@@ -0,0 +1,629 @@
...
+        target: "audit",
+        event = "service_token_issued",
+        service = %verified.service,
+        jti = %verified.jti,
+        roles = ?claims.roles,
+        tenants = ?claims.tenants,
+        "service token issued"
+    );
+
+    Json(serde_json::json!({
```
cyberantonz: Hm, looks like this should use GTS. Why you did not use it for requests and responses?

src/backend/services/authenticator/src/service_token.rs
```diff
@@ -0,0 +1,629 @@
...
+
+/// Readiness of the token listener (liveness of the socket; the gear only
+/// binds it once Redis + keys are up, so a reachable `/ready` implies both).
+async fn ready_handler() -> Response {
+    (StatusCode::OK, "ready").into_response()
+}
+
+/// The token listener's router: exactly the two endpoints it owns.
+fn router(state: Arc<AppState>) -> Router {
+    Router::new()
```
cyberantonz, line 410: Hm, OperationBuilder should be here to build routes.

src/backend/services/authenticator/src/service_token.rs
```diff
@@ -0,0 +1,629 @@
...
+        "token_type": "Bearer",
+        "expires_in": st.token_ttl_seconds,
+    }))
+    .into_response()
+}
+
+/// Readiness of the token listener (liveness of the socket; the gear only
+/// binds it once Redis + keys are up, so a reachable `/ready` implies both).
+async fn ready_handler() -> Response {
+    (StatusCode::OK, "ready").into_response()
```
cyberantonz: Isn't cf toolkit use other endpoints to check?

docker-compose.yml
```diff
@@ -324,6 +324,11 @@ services:
       APP__gears__authenticator__config__idp__client_id: "${OIDC_CLIENT_ID:-insight-authenticator}"
       APP__gears__authenticator__config__idp__client_secret: "${OIDC_CLIENT_SECRET:-}"
       APP__gears__authenticator__config__redirect_uri: "${AUTHENTICATOR_REDIRECT_URI:-http://localhost:8083/auth/callback}"
+      # Service tokens (step 06). The token endpoint runs on its own listener
```
cyberantonz: Why you need this comment?

docker-compose.yml
```diff
@@ -336,6 +341,9 @@ services:
       - "run"
     ports:
       - "${AUTHENTICATOR_PORT:-8083}:8083"
+      # Service-token listener (§10 G4 second listener). Exposed for local
```
cyberantonz: Why you need this comment?

cyberantonz:
Thanks for the review — pushed the fixes. Also fixed the CI break (the gateway e2e was 502ing, cause below).

### Your comments

- **Lock held across the fetch (`service_client.rs`)** — you're right. Cache is now a read-mostly `RwLock`: `bearer()` takes a short read lock, releases it, does the fetch with **no lock held**, then a brief write lock to store. (Trade-off: no single-flight — concurrent refreshes may each fetch; acceptable for a service client, and it removes the head-of-line blocking.) Added connect/request timeouts and serve-stale-but-valid on a refresh failure.
- **`tenant_scoped_allowed` → drop it, tenant isolation** — done. The flag is gone; **service tokens are always tenant-scoped**. A `/internal/token` request naming no tenant is rejected `400 tenant_required` (before the assertion's `jti` is consumed). This supersedes NGINX_BFF §10 G1's "no tenants = cross-tenant service work"; DD-AUTH-05 + the sequence/claim tables updated. Where the service gets its tenant is its own concern. 🙂
- **Don't check in throwaway keys — generate them** — done. Deleted the committed keypair; added `public_key_paths` + `public_key_dir` to the registry (inline `public_keys` stays for prod/gitops). `run-e2e.sh` and `dev-compose.sh` now generate a throwaway service keypair like the gateway signing key. Also mounted the dev-keys dir into the compose authenticator (that mount was missing).
- **jsonwebtoken v9 vs v10 skew** — unified `jsonwebtoken` as a workspace dep (v10, aws_lc_rs); bumped fakeidp 9→10 so the tree carries one version + one crypto provider. Promoted `serde_yaml` too.
- **"Do we really have those services in k8s?"** — no. `config/insight.yaml` is the dev/compose config; `testclient` is dev-only (now with no committed key). The chart ships an empty registry (`serviceTokens.services: {}`); real entries land via gitops.
- **`values.yaml` comment / compose comments** — rewritten to your phrasing / removed.
- **OperationBuilder + GTS for the token endpoint** — kept as a separate axum server **by design** (§11.8 / DD-AUTH-06): the token surface needs its own port + NetworkPolicy scope (G4), and the toolkit REST host binds a single address, so OperationBuilder (which registers into the host router + OpenAPI) can't drive it; it's deliberately off the public OpenAPI. I did type the response (`TokenResponse`) and renamed the listener's readiness endpoint `/ready` → `/health` to match the host/k8s convention. If you'd rather collapse it onto the main port (giving up the per-port NetworkPolicy split), say so and I'll change it.

### CodeRabbit

- ✅ reqwest timeout, serve-stale, `token_bind_addr != bind_addr` invariant, trimmed `iat`/`jti` from `required_spec_claims`.
- ❌ **`cf-studio-path` in Cargo.toml** — not applicable; that's a `cfs` config key (lives in `CLAUDE.md`), not a Cargo manifest key.
- ❌ **serde_yaml → yaml_serde** — kept `serde_yaml` (unified as a workspace dep) for consistency with the rest of the tree rather than introducing a second YAML crate.

### CI fix
The `e2e (5 scenarios)` job was failing (502 on `/auth/login`). The authenticator image bakes `config/insight.yaml`, which now carries the dev `testclient` registry entry, so the authenticator wouldn't boot without that public key present. The gateway e2e's `conftest.py` now generates it alongside `current.pem` and the e2e compose sets `public_key_dir=/keys`. Verified locally: **11 passed**.

All unit tests + both authenticator e2e loops (login, service-token) green.

src/backend/libs/authenticator-sdk/src/service_client.rs
```diff
@@ -0,0 +1,249 @@
...
+        service: impl Into<String>,
+        private_key_pem: &str,
+        endpoint: impl Into<String>,
+    ) -> Result<Self, CanonicalError> {
+        let encoding = EncodingKey::from_ec_pem(private_key_pem.as_bytes()).map_err(|e| {
+            CanonicalError::internal(format!("invalid service private key PEM: {e}")).create()
+        })?;
+        let http = reqwest::Client::builder().build().map_err(|e| {
+            CanonicalError::internal(format!("build service-token HTTP client: {e}")).create()
+        })?;
```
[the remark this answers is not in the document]
cyberantonz: Done — added `connect_timeout(5s)` + `timeout(15s)`. And reworked `bearer()` to release the cache lock **before** the fetch (see @cyberantonz's note below), so a slow authenticator no longer blocks concurrent callers on this client.

src/backend/services/authenticator/Cargo.toml
```diff
@@ -74,3 +74,6 @@ rand = "0.8"
 
 [dev-dependencies]
 tower = { version = "0.5", features = ["util"] }
+# Parse the checked-in dev config in a test so a YAML/PEM mistake in the
+# service-token registry fails fast, not only at container boot.
+serde_yaml = "0.9"
```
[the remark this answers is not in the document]
cyberantonz: Not applicable — `cf-studio-path` is a cf-studio (`cfs`) config key set in `insight/CLAUDE.md`, not a Cargo manifest key (cargo would reject the unknown key). No change.

src/backend/services/authenticator/src/config.rs
```diff
@@ -189,6 +254,80 @@ impl AuthenticatorConfig {
...
+                st.assertion_max_lifetime_seconds > 0,
+                "service_tokens.assertion_max_lifetime_seconds must be > 0"
+            );
+            for (name, entry) in &st.services {
+                anyhow::ensure!(
+                    !entry.public_keys.is_empty(),
+                    "service_tokens.services.{name} has no public_keys"
+                );
+            }
+        }
```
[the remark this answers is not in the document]
cyberantonz: Done — added the `token_bind_addr != bind_addr` invariant in `AuthenticatorConfig::validate()`, so a same-address misconfig fails at boot with a clear message.

src/backend/libs/authenticator-sdk/src/service_client.rs
```diff
@@ -0,0 +1,304 @@
...
+                let cache = self.cache.read().await;
+                if let Some(c) = cache.get(&key)
+                    && now < c.expires_at
+                {
+                    return Ok(c.bearer.clone());
+                }
+                Err(e)
+            }
+        }
+    }
```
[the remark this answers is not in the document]
cyberantonz: Done — `Cached` now tracks `expires_at`; on a refresh failure `bearer()` serves the stale-but-valid token if it hasn't truly expired, and only errors otherwise.

docs/components/backend/authenticator/DESIGN.md
cyberantonz: Agreed. Went with mandatory tenant isolation: a service token **must** name a tenant, else `400 tenant_required` — no cross-tenant service token. DD-AUTH-05 + the claim/sequence tables now say so; it explicitly supersedes §10 G1's "no tenants = cross-tenant service work".

src/backend/libs/authenticator-sdk/src/service_client.rs
cyberantonz: Fixed 😅 — cache is now a read-mostly `RwLock`: take a read lock to check, drop it, do the fetch with **no lock held**, then a brief write lock to store. Per-tenant keyed.

src/backend/libs/authenticator-sdk/Cargo.toml
cyberantonz: Fixed — promoted `jsonwebtoken` to a workspace dep (v10, aws_lc_rs) and bumped fakeidp 9→10, so the whole tree carries one version + one crypto provider. Did the same for `serde_yaml`.

src/backend/services/authenticator/config/insight.yaml
cyberantonz: No — this is the dev/compose config; `testclient` is dev-only (and now carries **no** committed key). The Helm chart ships an empty registry (`serviceTokens.services: {}`); real k8s services land their public key via a gitops PR.

src/backend/services/authenticator/dev/service-tokens/README.md
cyberantonz: Done — deleted the committed keypair (and this README). `run-e2e.sh` and `dev-compose.sh` now generate a throwaway service keypair at bring-up (exactly like the gateway signing key), referenced via `public_key_paths` + `public_key_dir`.

src/backend/services/authenticator/helm/values.yaml
cyberantonz: Rewritten to your phrasing ("access to the token port is disabled by default; enable it in your gitops config").

src/backend/services/authenticator/src/config.rs
cyberantonz: Done — `tenant_scoped_allowed` is gone entirely; service tokens are always tenant-scoped (a request with no tenant → `400`).

src/backend/services/authenticator/src/service_token.rs
cyberantonz: The response is a typed struct now (`TokenResponse`). GTS/types-registry registration is for the toolkit REST host's OpenAPI surface; this token endpoint is a **separate** axum listener on its own port (§11.8 / G4) and deliberately off the public OpenAPI, so it isn't a GTS-registered operation. The SDK (`ServiceTokenClient`) is the typed contract for callers.

src/backend/services/authenticator/src/service_token.rs
cyberantonz: OperationBuilder registers into the host router + generated OpenAPI, which is a single bound address. The token listener is a separate axum server on its own port because G4 needs a distinct NetworkPolicy scope and the host binds one address — so it can't go through OperationBuilder. If you'd rather collapse it onto the main port (dropping the per-port NetworkPolicy split), I'll switch it.

src/backend/services/authenticator/src/service_token.rs
cyberantonz: Good point — renamed the listener's `/ready` → `/health` to match the main host + k8s probe convention.

docker-compose.yml
cyberantonz: Removed.

docker-compose.yml
cyberantonz: Removed.

src/backend/libs/authenticator-sdk/src/service_client.rs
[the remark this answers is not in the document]
cyberantonz: Good catch on the `now()` here — the earlier version captured `now` **before** the `fetch().await` and reused it for `refresh_at`/`expires_at`, so with the new 15s request timeout a slow fetch would anchor the cached lifetime too early (expiring the entry before the real JWT `exp`). Fixed in 130ad893: `now()` is recomputed **after** the token is received (`received_at`), and the fast-path/stale checks re-read `now()` too.

src/backend/services/authenticator/src/service_token.rs
cyberantonz: Added a `TODO(#1583)` on `service_token::spawn` (commit 9a8f3d35) recording the target design: split the public surface (`/auth/*`, `/internal/authz`, JWKS) and the private service-token surface into **two containers (sidecar)**, each running on the toolkit host with OperationBuilder + generated OpenAPI and network-scoped at the pod boundary rather than per-port. Keeping the single-process two-listener shape for now (§11.8) since it isn't worth a second image + shared-state wiring until there's a service-token consumer (step 07).

ktursunov (approved):

PR #1759, opened 2026-07-14
fix(zoom): persist _meetings parent cursor in participants state (Heavy-quota fix)

mitasovr:
## Problem

After #1746 unblocked the zoom syncs on dev-vhc, they started tripping Zoom's Dashboard-API quota instead: job 538 succeeded only on attempt 2, job 541 spent hours in 429 retry loops (2026-07-14).

Root cause: `participants` is a substream whose private `_meetings` parent instance has **no persisted state** — parent state only survives piggybacked on an *incremental* child's state message, and `participants` was full-refresh (a no-op pinned by the mock suite in #1746). So **every sync re-enumerated the full now-150d window**: one `GET /v2/metrics/meetings/{uuid}/participants` per meeting ≈ **25k "Heavy" requests per sync** against the account-wide **60k/day** quota. Two syncs plus retry attempts (plus any manual full read) exhaust it; the CDK's six ~10s retries can't outwait a daily quota, so attempts fail.

## Fix

- `participants` gets a **formal `join_time` cursor**: no request options (the endpoint accepts no time filters) and no client-side filtering — it exists solely to make the stream stateful.
- `incremental_dependency: true` on the ParentStreamConfig now actually works: the `_meetings` cursor is persisted as `parent_state`, and each sync enumerates only meetings **newer than the saved cursor minus the parent's P7D lookback** — ~hundreds of Heavy requests per nightly sync instead of ~25k. Lookback duplicates dedup at read time as usual (bronze RMT + `LIMIT 1 BY`).
- **429 policy**: `ExponentialBackoffStrategy(factor 10)` fallback added after `WaitTimeFromHeader` on all four requesters — Zoom omits `Retry-After` on per-second throttles. Daily-quota 429s still fail the attempt (by design: Airbyte's attempt scheduling handles the wait better than an in-process sleep).
- Mock tests: the old "flag is a no-op" pin is replaced by its inverse — `test_parent_state_persisted_in_child_state` and `test_resume_enumerates_parent_from_saved_cursor` (exact from/to matchers prove the second sync's narrowed enumeration; a record older than the child cursor is still emitted, proving nothing is filtered).
- Descriptor 1.2.0, README note on the quota math.

First sync after deploy still does one full 150-day fan-out (no prior parent_state) — expected, same as today's every-sync cost; every following sync is cheap.

## Verification

- Mock suite: 36 passed / 1 skipped (pre-existing jira schema-drift skip)
- `validate-strict` + `validate`: green
- Existing connection state on dev-vhc is compatible: the participants stream simply gains state on its first post-deploy sync.

## Summary by CodeRabbit

- **Improvements**
  - Zoom participant synchronization now uses incremental backfills, fetching participants only for recently updated meetings.
  - Saved progress enables synchronization to resume efficiently after interruptions.
  - Reduced unnecessary Zoom Dashboard API requests and improved handling of temporary API errors.
  - Participant records now use more complete identifiers for reliable tracking.
- **Documentation**
  - Updated Zoom connector documentation to explain incremental backfill behavior and API quota benefits.
- **Maintenance**
  - Updated the connector version to 1.2.0.

### DIFF
--- a/src/ingestion/connectors/collaboration/zoom/tests/test_participants.py
+++ b/src/ingestion/connectors/collaboration/zoom/tests/test_participants.py
@@ -8,15 +8,34 @@ uuid URL-escaped in the path (`/`→%2F, `+`→%2B, `=`→%3D). AddFields stamps
 tenant_id / source_id / meeting_uuid (from the partition) / unique_key =
 "{tenant}-{source}-{meeting_uuid}-{participant_uuid}-{join_time}".
 
+How the sync works, for readers new to Airbyte substreams: `participants` has
+no meeting list of its own — before it can fetch anything it must learn which
+meetings exist. It does that through a private copy of the meetings stream
+(`_meetings`, defined inline in connector.yaml) that it reads itself; the
+top-level `meetings` stream from the catalog is a separate instance and shares
+nothing with it. Every meeting the private parent returns becomes one
+"partition" = one HTTP request to /metrics/meetings/{uuid}/participants.
+
+Those per-meeting requests are the expensive part: they count against Zoom's
+Dashboard-API "Heavy" quota (60 000 requests per day for the whole account).
+If the parent is re-read over its full 150-day window every sync, one sync
+costs ~25k requests on a busy account and two syncs a day exhaust the quota
+(this happened on dev-vhc, 2026-07-14). The fix asserted by the last two
+tests: remember between syncs how far the parent has been read (its
+`end_time` cursor), and on the next sync ask the API only for meetings newer
+than that. The remembering is awkward in Airbyte because state belongs to
+catalog streams and the private parent is not one — its cursor can only be
+saved inside the participants stream's own state, in a `parent_state` field,
+and the CDK only writes that field when participants itself is incremental.
+Hence the "formal" join_time cursor on participants: it filters nothing and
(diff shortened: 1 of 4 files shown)


src/ingestion/connectors/collaboration/zoom/tests/test_participants.py
```diff
@@ -135,13 +147,12 @@ def test_pagination_multi_page(http_mocker: HttpMocker) -> None:
...
 @freezegun.freeze_time(_NOW)
-def test_full_refresh_substream_emits_no_cursor_state(http_mocker: HttpMocker) -> None:
-    """participants is a full-refresh substream: the CDK emits only the
-    no-cursor sentinel state, and in particular NO parent_state — this is why
-    `incremental_dependency` is deliberately absent from the ParentStreamConfig
-    (it only piggybacks the parent cursor on an incremental child's state; on a
-    full-refresh child it is a silent no-op — this test found that). Freshness
-    is instead bounded by the parent's now-150d slicing, re-read each sync."""
+def test_parent_state_persisted_in_child_state(http_mocker: HttpMocker) -> None:
+    """The child's cursor is a formality; what matters is that its state
```
ktursunov: Comment is not clear. What's checked and what's the expected result

ktursunov (approved):

PR #1800, opened 2026-07-16
fix(ingestion): collab silver placeholders emit tenant_id for gold build

cyberantonz:
## Problem

A clean `make deploy ENV=local` fails its post-install `clickhouse-migrate` hook. The hook (`src/ingestion/scripts/apply-ch-migrations.sh`) builds the `tag:gold` dbt models, and `collab_metric_observations` errors on 1 of the 6 gold observation models:

```
Code: 47. DB::Exception: Unknown expression or function identifier `tenant_id` in scope
chat_source AS (SELECT tenant_id, person_key AS entity_id, ...
FROM silver.class_collab_chat_activity FINAL WHERE ...)  (UNKNOWN_IDENTIFIER)
```

This is a unified-metrics-collab merge regression (unrelated to auth); it reproduces identically on `main`.

## Root cause

Not the dbt models. The deploy hook runs only `dbt run --select tag:gold` against the **silver placeholder tables** created by `create-bronze-placeholders.sh` — the real silver class models are never materialised at deploy time.

- The connector staging models emit `tenant_id` and the class models are `SELECT *`, so the **real** silver tables are correct.
- The gold model `collab_metric_observations` selects `tenant_id` (mirroring the working git path), but the collab **placeholders** only declared `insight_tenant_id`.
- Git passes because the `class_git_*` placeholders already carry **both** `insight_tenant_id` and `tenant_id`.

## Fix (mirrors the `class_git_*` placeholders)

- `create-bronze-placeholders.sh`: add `tenant_id` to the chat/meeting/email/document collab silver placeholders, plus the other columns `collab_metric_observations` reads that were missing:
  - chat `direct_and_group_messages` — the `apply-ch-migrations.sh` heal only ALTERs a *real* table, never a placeholder, so a fresh deploy needs it declared;
  - meeting `meetings_organized` / `adhoc_meetings_attended` / `scheduled_meetings_attended`.
- `deploy/seed/generators/collab.py`: write `tenant_id` alongside `insight_tenant_id` (same dual-column pattern as the git seed) so seeded e2e collab metrics carry the real tenant instead of `''`.

No auth/chart/gitops files touched — purely ingestion.

## Verification

Against a real ClickHouse, running the exact deploy steps (placeholders → migrations → `dbt run --select tag:gold`):

- **Before:** reproduced the identical `UNKNOWN_IDENTIFIER 'tenant_id'` failure on `collab_metric_observations`.
- **After:** `collab_metric_observations`, `ai_metric_observations`, and `git_metric_observations` all build cleanly; seed-shaped rows flow the real `tenant_id` into all emitted `insight.collab_metric_observations` rows (verified non-empty tenant).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved collaboration activity handling by consistently including `tenant_id` across meeting, chat, email, and document activity records.
  * Added missing meeting and chat activity metrics to placeholder datasets to ensure complete reporting.
  * Prevented downstream view mismatches by extending/reconciling existing placeholder schemas with the new expected columns.


ktursunov (approved):

deploy/seed/generators/collab.py
```diff
@@ -72,10 +72,22 @@ def seed_meeting_activity(
...
     cols = [
-        "insight_tenant_id", "email", "person_key", "date", "data_source",
-        "meetings_attended", "calls_count", "participants",
-        "audio_duration_seconds", "video_duration_seconds",
-        "screen_share_duration_seconds", "_version",
+        # `tenant_id` mirrors `insight_tenant_id`: the real dbt model /
+        # connectors emit `tenant_id`, and gold (collab_metric_observations)
+        # reads it. Same dual-column pattern the git seed uses.
+        "insight_tenant_id",
+        "tenant_id",
```
ktursunov, line 79: What's the reason for two tenant ids?

cyberantonz: I do not know, maybe better ask the authors of the table. Next you can see claude findings.

cyberantonz: Two names for the same value, kept for compatibility during the ADR-0007 placeholder transition:

- `insight_tenant_id` is what the compose seed has historically written (and matches the connector *config* key `insight_tenant_id`).
- `tenant_id` is the column the real dbt model and connector staging models actually emit, and it's what the gold model `collab_metric_observations` reads.

The `class_git_*` seed uses the same dual-column pattern (writes `tenant_uuid` into both). Writing both here keeps the seeded silver tables readable by gold (`tenant_id`) while staying consistent with the existing `insight_tenant_id` convention. Once silver is created exclusively by dbt (ADR-0007 Variant A), the `insight_tenant_id` alias can be dropped.

src/ingestion/scripts/create-bronze-placeholders.sh
```diff
@@ -129,6 +129,11 @@ if ! ch_table_exists silver class_collab_email_activity; then
   run_ch <<'SQL'
 CREATE TABLE IF NOT EXISTS silver.class_collab_email_activity (
     insight_tenant_id String,
+    -- `tenant_id` is the column the real dbt model emits (connectors SELECT
+    -- `tenant_id`); the gold model collab_metric_observations reads it. The
+    -- placeholder must carry it too or the deploy-time gold build fails with
+    -- UNKNOWN_IDENTIFIER. Mirrors the class_git_* placeholders (both cols).
+    tenant_id         String,
```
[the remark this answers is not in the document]
cyberantonz: Good catch — valid and fixed in 5066a845.

Confirmed the gap: the collab blocks only had a create branch, so on a warm cluster (re-deploy, or a persistent e2e ClickHouse) `CREATE TABLE IF NOT EXISTS` is skipped and the new columns are never applied — the same `UNKNOWN_IDENTIFIER` would recur.

Added `else` reconciliation branches for all four `class_collab_*_activity` placeholders, mirroring the `class_git_commits` / `class_ai_dev_usage` pattern: `INSIGHT_PLACEHOLDER_v1` marker-guarded (a real dbt-built table is never mutated) and idempotent (`ADD COLUMN IF NOT EXISTS`) — `tenant_id` on all four, plus chat `direct_and_group_messages` and meeting `meetings_organized` / `adhoc_meetings_attended` / `scheduled_meetings_attended`.

Verified against a warm ClickHouse: an old-schema placeholder is reconciled in place and `collab_metric_observations` then builds cleanly.

ktursunov (approved):

PR #1813, opened 2026-07-17
feat(ingestion): bootstrap-db — create bronze/silver/gold empty tables 

### LINKED ISSUE #1807: Seed DB from AirByte connectors and DBT models

mitasovr:
Closes #1807

## What

New `src/ingestion/scripts/bootstrap-db/` toolkit that creates every connector's bronze tables in ClickHouse without a running Airbyte, plus removal of the unused `github` v1 and `bitbucket-server` connectors.

## How it works

For each connector the source image runs `discover` with a generated fake config (stream schemas are static for all connectors except hubspot/salesforce, so no credentials and no network access to the source are needed). The resulting catalog is wrapped into a configured catalog (`append`, per-connector namespace, generation/sync ids) and fed to the same `airbyte/destination-clickhouse` 2.x connector Airbyte itself uses, with only per-stream `STREAM_STATUS: COMPLETE` trace messages on stdin — the Bulk CDK creates the tables in `StreamLoader.start()` before any records, so all stream tables appear empty. Afterwards the per-connector `<name>__bronze_promoted` dbt model promotes them to ReplacingMergeTree.

Unlike `create-bronze-placeholders.sh`, the schemas come from the same source of truth as a real sync (the connector spec in git), so they cannot drift.

## Scripts

| Script | Role |
|---|---|
| `generate-connectors-config.sh [pattern]` | descriptors → YAML config with all required spec fields and fake values (`default` → `examples[0]` → type/format) |
| `seed-connectors.sh <yaml>` | iterates the config, resolves `value:`/`env:` fields, calls the per-connector script; a failure is reported and skipped |
| `create-connector-tables.sh <dir> <config.json>` | one connector: discover → configured catalog → destination write → RMT promotion |
| `bootstrap-db.sh <yaml>` | sources optional `.env`, seeds all connectors, runs all dbt models, applies `apply-ch-migrations.sh` |

Config fields support `env: VAR_NAME` references resolved at run time, so real tokens (needed only for hubspot/salesforce, whose schemas come from live describe APIs) never land in the committed file. `connectors-config.yaml` with fake values for 23 connectors is included.

`DESTINATION_CLICKHOUSE_IMAGE` is pinned via `.env` (currently `2.1.25`); the README documents how to query the running Airbyte for the exact version it uses, since the platform seeds connector versions from the registry at install time.

## Connector removal

`github` v1 is superseded by `github-v2` and is not deployed anywhere; `bitbucket-server` was never provisioned. The dangling `-- depends_on: {{ ref('github__*') }}` hints are dropped from the 8 silver git class models — `union_by_tag` selection is tag-driven, so nothing else changes.

## Testing

Verified end-to-end on a throwaway ClickHouse 24.8 container (README describes the setup): outline (nocode) and bitbucket-cloud (CDK) → all 12 tables created empty and promoted to ReplacingMergeTree; hubspot with fake creds fails discover, is reported and skipped, the run continues; all-models dbt run and ClickHouse migrations complete.

## Summary by CodeRabbit

- **New Features**
  - Added a workflow to bootstrap connector tables and initialize ClickHouse ingestion without running Airbyte.
  - Added automated connector discovery, configuration generation, table creation, dbt execution, and migration steps.
  - Added sample environment settings and a consolidated connector configuration template.

- **Documentation**
  - Added setup instructions, prerequisites, usage guidance, troubleshooting details, and script references for database bootstrapping.

- **Changes**
  - Removed support for the Bitbucket Server connector.


src/ingestion/scripts/bootstrap-db/create-connector-tables.sh
```diff
@@ -0,0 +1,96 @@
...
+
+jq -n '{
+    host: env.CLICKHOUSE_HOST,
+    port: env.CLICKHOUSE_PORT,
+    protocol: env.CLICKHOUSE_PROTOCOL,
+    database: env.CLICKHOUSE_DATABASE,
+    username: env.CLICKHOUSE_USER,
+    password: env.CLICKHOUSE_PASSWORD,
+    enable_json: false
+  }' > "${WORKDIR}/destination_config.json"
```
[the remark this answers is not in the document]
mitasovr: This suggestion applies to the 1.x connector spec. `airbyte/destination-clickhouse` 2.x (Bulk CDK) changed the schema: `port` is now a **string** and `tonumber` would fail validation with the opposite error. This is documented in `src/ingestion/reconcile-connectors/lib/reconcile.sh` ("the 1.x->2.x rewrite changed the schema — port is now a STRING (not int)") and verified against a live `2.1.25` image: the string form passes validation and creates the tables, while the 1.x image rejects it with `$.port: string found, integer expected`. Keeping the string.

src/ingestion/scripts/bootstrap-db/run-dbt.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+: "${CLICKHOUSE_HOST:?CLICKHOUSE_HOST must be set}"
+: "${CLICKHOUSE_PORT:?CLICKHOUSE_PORT must be set}"
+: "${CLICKHOUSE_PROTOCOL:?CLICKHOUSE_PROTOCOL must be set (http or https)}"
+: "${CLICKHOUSE_USER:?CLICKHOUSE_USER must be set}"
+: "${CLICKHOUSE_PASSWORD:?CLICKHOUSE_PASSWORD must be set}"
+
```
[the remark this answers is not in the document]
mitasovr: After aligning the profile with the production dbt-run template (schema `silver`, commit 107059b9), the dbt profile does not use `CLICKHOUSE_DATABASE` at all — every model sets its schema explicitly via `schema=` config and `generate_schema_name` only falls back to `target.schema`. The production profile (charts/insight/templates/ingestion/dbt-run.yaml) does not reference it either; `seed-db.sh` requires the variable separately for the migrations step. Adding the check here would assert a variable this script never reads.

src/ingestion/scripts/bootstrap-db/bootstrap-db.sh
ktursunov, line 1: Suggest renaming to migrate db or bootstrap db. seeding term usually about filling the table with data

ktursunov (changes requested):

src/ingestion/scripts/bootstrap-db/connectors-config.yaml
```diff
@@ -0,0 +1,310 @@
...
+        value: fake
+      insight_tenant_id:
+        value: fake
+      zulip_proxy_api_key:
+        value: fake
+      zulip_proxy_base_url:
+        value: http://10.0.0.1:9999/api
+      zulip_proxy_start_date:
+        value: "2024-01-01"
+        # hubspot:
```
ktursunov, line 133: Please add comment why those are commented

src/ingestion/scripts/bootstrap-db/connectors-config.yaml
```diff
@@ -0,0 +1,316 @@
+connectors:
+  chatgpt-team:
+    path: ai/chatgpt-team
+    config:
+      chatgpt_account_id:
+        value: fake
```
ktursunov, line 6: What are those fake names? Does it mean it would be generated?

mitasovr: just placeholders, it doesn't matter what you will pass here
just to overcome required fields check

ktursunov: Could you please make comment

ktursunov (approved):

Issue #1754, opened 2026-07-14
identity-resolution: write side — persons seed + persons-seed API


mozhaev-dev:
Sub-task of #1602. Write side (iteration 2).

- Persons seed: 3-branch resolution (known account / initial bootstrap / steady-state incl. `pending-iresolution`); idempotent; rebuild `account_person_map` (SCD2).
- `persons-seed` API + background worker (`POST` / `GET /v1/persons-seed*`).
- Hand over `identity` DB schema / migration ownership (DbUp → sea-orm-migration); single owner during transition.

Issue #1753, opened 2026-07-14
identity-resolution: API contract / e2e tests (read endpoints)


mozhaev-dev:
Sub-task of #1602.

Black-box HTTP contract / e2e tests for the identity read endpoints against a seeded DB: response shapes, status codes, canonical error envelopes (400 / 404 / 409 / 415). Wire into the e2e API suite (feeds `api-endpoint-coverage-gate`) and address the per-crate coverage gate for DB-touching paths.

Issue #1752, opened 2026-07-14
identity-resolution: complete read API (/v1/profiles + GET /v1/persons)


mozhaev-dev:
Sub-task of #1602.

Finish the read side of the Rust identity service.

- Complete `ProfileResponse`: `ids[]` + org tree (`supervisor_*`, `parent_*`, `subordinates[]`).
- Port deprecated `GET /v1/persons/{email}` (Phase-1 query) + resolve the `analytics` `IdentityClient` dependency (port it, or migrate analytics to `/v1/profiles`).
- Unit / router tests included.

**Contract decisions (blocker for parity):**
- Error taxonomy — gears `type: gts://…` vs .NET `urn:insight:error:*`; gears has no 422 (ambiguous → 409). Byte-exact parity vs gears-native?
- Caller identity — enforce `X-Insight-Person-Id` (401) in-service, or delegate to api-gateway?

First slice landed in #1745 (`POST /v1/profiles` resolve email/id + attributes).

Issue #1602, opened 2026-07-03
GRAPH 1: Identity resolution — port the C# service to Rust


mitasovr:
## Goal

Port the existing .NET (C#) identity-resolution service to Rust at behavioural
parity, cut all traffic over to it via api-gateway, and decommission the .NET
service. The **matching/mapping logic itself** (how accounts collapse into a
canonical person) is specified in its own epic — see Relations; this epic is the
service reimplementation, not new resolution behaviour.

## Context

The identity store is already live (BambooHR + Entra with a reconcile loop). What
remains is moving the running implementation off .NET onto Rust without changing
observable behaviour, then retiring the old service.

## Hard constraints — no administration surface, and no administrator exists

Two separate facts, both verified against `main` (2026-07-27). Together they
determine what can actually be tested today.

**1. There is no administrative user interface.** Everything below is a *service
capability, not a screen*. The product has no administration area: no screen for
access grants, roles, people/linked-account management, or triggering an identity
refresh. What the UI does show: a person + team view (`/ic/<person>`), a metrics
view, a what's-new page — all read-only consumers of identity data.

**2. There is no administrator in any environment.** The mechanism is fully built
(admin-gated operations reject callers without an active `admin` role), but no role
holder is ever created: seeding the first admin is optional config, is **not set
anywhere** in the deployment configuration, and when unset the seeding step is a
**silent no-op** (it also requires a default company id, itself a placeholder).
**Net effect: nobody is an administrator, so every administrative operation is
refused for everyone.**

### What this means for testing

- "Administrator" is a role attached to a request, not a person with a console.
  Do **not** write criteria as "the administrator opens X and clicks Y".
- Every requirement below is testable **only at service level**.
- **Testable with no setup:** BR-7 (see yourself), BR-8 (own reporting chain, any
  depth), BR-9/BR-15 (refused for anyone outside it). Prioritise these — they are
  what actually governs access today.
- **Blocked without seeding an admin first:** acceptance criteria 4, 5, 6, 11.
  Without that step they cannot pass, and that is **current expected behaviour,
  not a defect**.
- The grant-based half of the access model (BR-10, BR-11) is **not exercisable in
  a default install**.
- This is a known product gap, not a defect of this migration. The administration
  surface is separate work, called out as required in #1873.

## Scope

- Port the read side: persons, roles, resolve (JWT-claim → `person_id`) endpoints.
- Port remaining domains: org-chart (`/v1/subchart*`), roles + person-roles
[...]

PR #1825, opened 2026-07-20
feat(identity-resolution): persons-seed write path (POST /v1/persons-seed) (#1754)

mozhaev-dev:
Rust port of the identity **write** path — the persons-seed job — on gears-rust. Second slice of the identity rewrite (epic #1602), on top of the merged read API (#1745 / #1752).

## What this adds
- `POST /v1/persons-seed` (202 + operation id), `GET /v1/persons-seed/{id}`, `GET /v1/persons-seed` — job-tracked via the `operations` table.
- Background worker (spawned in gear init) drains the queue and runs the seed.
- Pipeline: ClickHouse `identity_inputs` reader → fold to per-account profiles → group by email → 4-branch resolution (known-account / email-link / mint / skip) → transactional MariaDB apply (batched `INSERT IGNORE` into `persons` → `account_person_map` SCD2 rebuild → `org_chart` rebuild).
- Domain logic is pure and behind reader/store traits — 31 unit tests with fakes, no DB.

## Live-verified on dev
Seeded the dev tenant end-to-end: `accounts_read=4736, reused=3966, linked=7, minted=634, observations_inserted=25059`, completes in ~21s (batched INSERT; per-row was 20+ min). Org tree intact after rebuild (Serdar → Tasi → Gürcan, subordinates expand).

## Read-side fix included
The live rebuild surfaced a latent read bug: `current_parents_for_child` decoded `parent_person_id` into a non-nullable `Uuid`, so a single NULL-parent `org_chart` row (Path-B root/membership rows, #344) 500'd the whole profile. Fixed by skipping NULL parents in the query — a parent edge with no parent is not an edge.

## Notes
- `identity-resolution` is not yet a registered CI component, so it is not built/gated here; component registration + integration tests land with #1753.
- Ported 1:1 from the .NET `PersonsSeedService` / `Sql.OrgChart.cs`.

Refs #1602, #1754

## Summary by CodeRabbit

* **New Features**
  * Added a `persons-seed` admin workflow with `POST /v1/persons-seed`, `GET /v1/persons-seed/{id}`, and `GET /v1/persons-seed` to manage queued/running/completed/failed operations.
  * Supports `mode` (defaults to link-by-email), returns `202 Accepted`, and provides operation status; background processing reclaims stale jobs and enforces timeouts.
* **Configuration**
  * Added ClickHouse connection settings for the identity-resolution service.
* **Bug Fixes**
  * Prevents invalid organization-chart rows when parent identifiers are missing.
* **API Changes**
  * Removed the deprecated `GET /v1/persons/{email}` endpoint.


src/backend/services/identity-resolution/src/infra/db/seed_repo.rs
```diff
@@ -0,0 +1,546 @@
...
+    // observation; INSERT IGNORE swallows the duplicate-key error. Batched
+    // (multi-row VALUES) so N observations cost ~N/INSERT_CHUNK round-trips
+    // instead of N — 25k+ single-row inserts over a remote pool take minutes;
+    // batches take seconds. Same semantics as per-row INSERT IGNORE.
+    const INSERT_PREFIX: &str = "INSERT IGNORE INTO persons \
+        (value_type, insight_source_type, insight_source_id, insight_tenant_id, \
+         value_id, value_full_text, value, person_id, author_person_id, reason, \
+         created_at) VALUES ";
+    const ROW_TUPLE: &str = "(?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)";
+    const INSERT_CHUNK: usize = 500; // 500 rows × 11 cols = 5500 binds (< 65535)
```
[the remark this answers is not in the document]
mozhaev-dev: Keeping `INSERT IGNORE` here — declining with rationale:

- The only columns that could truncate (`value_type`, `insight_source_type`, `reason`) come from a **controlled vocabulary**, not user input: `value_type` is a fixed domain set (`email`, `parent_email`, `display_name`, …), `insight_source_type` is a connector id (`bamboohr`, `slack`, …), `reason` is the constant `auto-seed-link`. None approaches `VARCHAR(50/100)`.
- Oversized *values* are already handled upstream: `route_value` drops over-limit `value_id` (≤320) / `value_full_text` (≤512) to all-`None` rather than truncating.
- `IGNORE` is load-bearing for **idempotency**: re-running a seed relies on `uq_person_observation` swallowing the duplicate-key error, so an identical re-emitted observation is a no-op. Dropping `IGNORE` would break re-seed.

So the truncation-then-silent-skip path isn't reachable with the data this path produces. Left as-is.

mozhaev-dev:
Addressed the CodeRabbit review in d71e29a9.

**Fixed (5):**
- **`try_start` strands `queued`** — worker now marks the operation `failed` on a transient `try_start` error, matching the queue-full path.
- **No worker timeout** — each `run_seed` is bounded by a 10-min `tokio::time::timeout`; a hung DB call no longer wedges the serial queue for every tenant.
- **Raw error leak** — persisted `error_message` is now a generic "see server logs"; the real anyhow/driver text stays in the tracing log only.
- **Silent multi-person binding merge** — the resolver now detects, counts (`known_binding_conflicts`, surfaced in the operation summary), and warn-logs a group whose accounts are bound to different persons; added a unit test.
- **`operations.list` over-fetch** — pushed the `operation_type` filter and a `LIMIT 200` cap into SQL instead of scanning + filtering in app code.

**Declined, with rationale:**
- **`INSERT IGNORE` truncation** — replied inline: truncatable columns are a controlled vocabulary, oversized values are already dropped by `route_value`, and `IGNORE` is required for re-seed idempotency.
- **`fetch_all` full-materialization** — fine at current scale (~4.7k accounts, ~21s) and matches the .NET reader; row-streaming adds complexity without a present need.
- **Startup sweep for orphaned `queued` rows** — a real robustness improvement, but it needs a cross-tenant query and the multi-pod worker story isn't finalized (single in-process worker today). Tracked for the hardening pass (#1753) rather than this PR.

src/backend/services/identity-resolution/src/api/mod.rs
```diff
@@ -62,7 +66,7 @@ fn build_operations(router: Router, openapi: &dyn OpenApiRegistry) -> Router {
 
     // Deprecated: successor is POST /v1/profiles. Kept for existing callers
     // (authenticator, analytics) until they migrate; emits RFC 8594 headers.
-    OperationBuilder::get("/v1/persons/{email}")
+    let router = OperationBuilder::get("/v1/persons/{email}")
```
cyberantonz: I think it should be dropped. I removed (i hope everywhere) requests to this endpoint

src/backend/services/identity-resolution/src/api/seed.rs
```diff
@@ -0,0 +1,495 @@
...
+
+/// Default page size / cap for the list endpoint (parity with the .NET
+/// `PageRequest.DefaultLimit` / `MaxLimit`).
+const LIST_DEFAULT_LIMIT: u64 = 50;
+const LIST_MAX_LIMIT: u64 = 500;
+
+/// Header carrying the caller's `person_id`, parity with the .NET
+/// `HeaderCallerContext`. JWT id/email-claim fallbacks are deferred until gears
+/// auth carries a subject (the host runs auth-disabled today — no claims).
+const CALLER_HEADER: &str = "X-Insight-Person-Id";
```
cyberantonz: This header is deprecated and removed. The JWT token is the source of truth

src/backend/services/identity-resolution/src/api/seed.rs
```diff
@@ -0,0 +1,495 @@
...
+            tracing::error!(error = %e, "list operations failed");
+            CanonicalError::internal("failed to list operations").create()
+        })?;
+    let items = ops
+        .into_iter()
+        .map(PersonsSeedOperationResponse::from)
+        .collect();
+    Ok(Json(PersonsSeedListResponse {
+        items,
+        next_cursor: None,
```
cyberantonz: Is this used anywhere? I do not see that this function use cursor at all. And also it returns null. 

Or is it for future?

src/backend/services/identity-resolution/src/domain/seed_service.rs
```diff
@@ -0,0 +1,288 @@
...
+    mint: impl FnMut() -> Uuid,
+) -> anyhow::Result<SeedSummary>
+where
+    R: IdentityInputsReader + ?Sized,
+    S: SeedStore + ?Sized,
+{
+    // 1. Build per-account profiles from the (latest-first) input stream.
+    let rows = reader.stream(tenant_id).await?;
+    tracing::info!(input_rows = rows.len(), "persons-seed: input streamed");
+    let profiles = build_profiles(rows);
```
cyberantonz, line 106: Scary place, all users are selected and enriched. Hope it will not fail sometimes :) But currently it is ok, we have no lot of users.

src/backend/services/identity-resolution/src/api/mod.rs
mozhaev-dev: Done in 45771f62 — removed the `GET /v1/persons/{email}` route + `get_person_by_email` handler (and the now-dead `PersonError` / `resolve_person_id_by_email`). POST /v1/profiles is the sole read path.

src/backend/services/identity-resolution/src/api/seed.rs
mozhaev-dev: Done in 45771f62 — dropped the `X-Insight-Person-Id` header entirely. The admin gate now takes the caller from `SecurityContext::subject_id()` (the gateway-JWT subject verified by the host authn pipeline); nil subject → 401. Removed `resolve_caller` + the header constant + its tests.

src/backend/services/identity-resolution/src/domain/seed_service.rs
mozhaev-dev: Agreed — the reader materializes the whole tenant's filtered inputs into a Vec rather than streaming (the .NET reader is IAsyncEnumerable). Fine at current sizes; the run is bounded by SEED_TIMEOUT (10m) and the ClickHouse read timeout (raised to 5m). Row/chunked streaming is tracked for the hardening pass (#1753).

src/backend/services/identity-resolution/src/infra/db/ops_repo.rs
```diff
@@ -0,0 +1,301 @@
...
+pub async fn enqueue(
+    db: &DatabaseConnection,
+    operation_id: Uuid,
+    operation_type: &str,
+    tenant_id: Uuid,
+    author_person_id: Uuid,
+    request_json: Option<&str>,
+) -> anyhow::Result<()> {
+    const SQL: &str = r"
+        INSERT INTO operations
```
cyberantonz, line 86: Those queries will be easily found on security review :)

Can we use toolkit-db for those? Or it is overkill? If so - the comment is needed that it is intended.

src/backend/services/identity-resolution/src/api/seed.rs
mozhaev-dev: Dropped in 85cfd83c — list responses are now the leaner `{items}` shape (removed `next_cursor` here and, on the stacked domains branch, from the roles/person-roles/visibility lists too for consistency).

src/backend/services/identity-resolution/src/infra/db/persons_repo.rs
```diff
@@ -280,11 +280,17 @@ pub async fn current_parents_for_child(
     child_person_id: Uuid,
 ) -> anyhow::Result<Vec<OrgChartEdge>> {
     const SQL: &str = r"
+        -- `parent_person_id IS NOT NULL` intentionally diverges from the .NET
```
cyberantonz: Why the comment inside SQL query ?? ))

mozhaev-dev: fixed

src/backend/services/identity-resolution/src/infra/db/ops_repo.rs
mozhaev-dev: Issue craated

mitasovr (approved):

cyberantonz (approved):

PR #1826, opened 2026-07-20
fix(ingestion): exclude vendored/generated files from authored git LOC (#1798)

### LINKED ISSUE #1798: Exclude vendored and generated files from git line-count metrics
## Problem

Git line-count metrics (e.g. "Code lines added", "Lines added") sum every
added line reported by the commit diff. Files are bucketed only by
**extension** (code / docs / config), with no consideration of whether the
file is authored or machine-produced. As a result, content that a person did
not write by hand is counted as their contribution:

- vendored dependencies checked into the repo (e.g. `.venv` / `site-packages`,
  `node_modules`, `vendor/`)
- build/generated output (e.g. `dist/`, `build/`, minified bundles, generated
  code or data files)
- lockfiles and other machine-managed artifacts

A single commit that adds such a directory can contribute hundreds of
thousands of lines under a code extension (e.g. vendored `.py`/`.js`), which
are then attributed as authored code.

## Impact

- **Per-person distortion:** individuals whose repos happen to include
  vendored/generated content show wildly inflated line counts that do not
  reflect real output. This correlates with role and repo hygiene rather than
  contribution.
- **Peer-pool distortion:** because peer standing is computed against a
  cohort (e.g. department) median/quartiles, one such outlier skews the
  benchmark for everyone in that cohort, degrading every peer comparison in
  the group.
- Line-count metrics are currently untrustworthy wherever a repo does not
  perfectly gitignore generated/vendored content — which is common.

## Root cause

File classification keys on file extension only. There is no path-based
exclusion for vendored/generated content, so such files pass through the same
as authored source.

## Proposed direction

Add a path-based exclusion applied before line-count aggregation, so
vendored/generated files are not counted toward authored line metrics:

- Maintain an exclusion list of well-known vendored/generated path patterns
  (dependency directories, build output, minified assets, lockfiles). Treat it
  as configuration so it can evolve without code changes.
- Apply the exclusion in the git file-change classification / observation
  layer so it flows consistently to every downstream line-count metric and to
  peer aggregation.
- Consider a separate "generated/vendored" category rather than a hard drop,
  so the signal is auditable and could be surfaced distinctly later.

## Acceptance criteria

- [ ] Vendored/generated files are excluded from authored line-count metrics.
- [ ] Exclusion patterns are configurable, not hard-coded per call site.
- [ ] Exclusion is applied once, upstream, and reflected in every line-count
      metric and in peer/cohort aggregation.
- [ ] Coverage for representative vendored/generated path patterns.

## Scope

Independent of the metric-family migrations; affects the git domain only.

cyberantonz:
Closes #1798.

## Problem
Git line-count metrics classify files by **extension only**, so machine-produced content — vendored deps (`node_modules/`, `site-packages/`, `vendor/`), build output (`dist/`, `target/`), and generated/minified files — is counted as authored code. A single commit adding a vendored tree inflates a person's LOC by hundreds of thousands of lines, and because peer standing is computed against a cohort median/quartiles, one such outlier skews the benchmark for everyone in the cohort.

## Fix
Add a `vendored` category, checked **first** in both classifier call sites:
- the shared `git_file_category` gold macro (`src/ingestion/dbt/macros/git_file_category.sql`), consumed by `git_metric_observations`
- the inline classifier in `fct_git_file_change` (`src/ingestion/silver/git/`), consumed by `mtr_git_person_*`

Authored line-count metrics already filter to `category = 'code'` (`code_lines_added`, `code_loc`, `clean_loc`), so they now exclude vendored/generated files automatically. The total `lines_added` breakdown **retains** them under the `vendored` category, so the signal stays auditable rather than silently dropped (as the issue suggested).

Patterns are defined **once** — the new `git_vendored_path_regex` macro backed by the `git_vendored_path_patterns` dbt var (`dbt_project.yml`) — so the exclusion is applied upstream and evolvable via config without a code change.

## Acceptance criteria
- [x] Vendored/generated files excluded from authored line-count metrics (`code_lines_added` / `code_loc` / `clean_loc`).
- [x] Exclusion patterns configurable (`git_vendored_path_patterns` var), not hard-coded per call site (shared `git_vendored_path_regex` macro).
- [x] Applied once upstream and reflected in every line-count metric + peer/cohort aggregation.
- [x] Coverage for representative vendored/generated path patterns (`assert_git_file_category_vendored.sql`, data-independent so it holds on a fresh cluster).

## Notes
- Escaping: var patterns are RE2 fragments embedded into a ClickHouse SQL string literal, so backslashes are doubled to match the existing hand-written regexes; verified end-to-end (YAML load → ClickHouse literal unescape → RE2) against vendored vs authored fixtures.
- Extended the closed-taxonomy contract test (`assert_git_observations_dimension_values.sql`) with `vendored`.

## Summary by CodeRabbit

* **New Features**
  * Added automatic classification of vendored and generated files in Git metrics.
  * Added configurable patterns covering dependencies, build outputs, generated files, and lock files.
  * Added a distinct “Vendored / Generated” category in file and line breakdowns.

* **Bug Fixes**
  * Vendored content is now excluded from code-location metrics and code-line attribution.

* **Documentation**
  * Updated Git metric definitions and category descriptions to explain vendored-file handling.

### DIFF
--- a/src/ingestion/dbt/dbt_project.yml
+++ b/src/ingestion/dbt/dbt_project.yml
@@ -59,11 +59,22 @@ vars:
     - '\\.generated\\.'
     - '\\.designer\\.cs$'
     - '\\.d\\.ts$'
+    # lockfiles / machine-managed manifests (checked before the `config`
+    # branch, so all lockfiles land in one auditable `vendored` bucket)
     - '(^|/)go\\.sum$'
     - '(^|/)Cargo\\.lock$'
     - '(^|/)composer\\.lock$'
     - '(^|/)Gemfile\\.lock$'
     - '(^|/)pnpm-lock\\.yaml$'
+    - '(^|/)package-lock\\.json$'
+    - '(^|/)npm-shrinkwrap\\.json$'
+    - '(^|/)yarn\\.lock$'
+    - '(^|/)poetry\\.lock$'
+    - '(^|/)Pipfile\\.lock$'
+    - '(^|/)packages\\.lock\\.json$'
+    - '(^|/)gradle\\.lockfile$'
+    - '(^|/)mix\\.lock$'
+    - '(^|/)flake\\.lock$'
 
 # on-run-start hooks:
 #  - create_task_field_history_staging:
--- a/src/ingestion/dbt/dbt_project.yml
+++ b/src/ingestion/dbt/dbt_project.yml
@@ -17,6 +17,54 @@ clean-targets:
   - target
   - dbt_packages
 
+vars:
+  # Path patterns (ClickHouse `match` / RE2 regex fragments, matched
+  # case-insensitively) that mark a git file as vendored/generated rather than
+  # authored. Files matching any pattern are classified `vendored` by the
+  # shared `git_file_category` macro and so are excluded from authored
+  # line-count metrics (code_loc / clean_loc / code_lines_added) while still
+  # being retained — and auditable — under the `vendored` category in the
+  # total lines_added breakdown. Override per deployment to extend coverage.
+  # NOTE: values are RE2 fragments embedded into a ClickHouse SQL *string
+  # literal*, so backslashes are doubled (`\\.` -> ClickHouse unescapes to
+  # `\.` -> RE2 sees a literal dot), matching the convention in the other
+  # hand-written regexes in git_file_category.sql. Single-quoted YAML keeps the
+  # doubled backslashes verbatim.
+  git_vendored_path_patterns:
+    # dependency / vendored trees
+    - '(^|/)node_modules/'
+    - '(^|/)bower_components/'
+    - '(^|/)vendor/'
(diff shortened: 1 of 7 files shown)


cyberantonz:
Addressed CodeRabbit review in 15ea04c5:
- **Major (empty-override):** `git_vendored_path_regex` now emits an RE2-safe never-match `[^\s\S]` when `git_vendored_path_patterns` is empty, instead of `(?i)()` (which would have matched every path). Verified end-to-end.
- **Minor (lockfiles):** added the remaining common lockfiles (package-lock.json, yarn.lock, poetry.lock, Pipfile.lock, gradle.lockfile, mix.lock, flake.lock, npm-shrinkwrap.json, packages.lock.json) so all machine-managed lockfiles classify as `vendored` consistently. Fixture test extended.

ktursunov (approved):

Issue #1727, opened 2026-07-09
Add gateway-auth smoke + provider-invariant tests to catch JWT-verify/feature-unification regressions


cyberantonz:
## Problem

The jsonwebtoken crypto-provider conflict fixed in #1725 (dual `aws_lc_rs` +
`rust_crypto` → panic on every JWT verify) reached `main` and 500'd api-gateway
auth on dev **despite** green unit tests and a passing authenticator e2e. It's a
class of failure our current tests structurally cannot catch:

- **Unit tests are per-crate.** They don't reflect the feature set of a
  *combined* build (`build-rust` / the image builds compile api-gateway +
  analytics + authenticator together, so Cargo feature unification changes what
  `jsonwebtoken` ends up with).
- **The authenticator e2e ran the authenticator standalone**, which happened to
  carry the "right" (well, one) provider, so it passed — while api-gateway,
  built in the same combined invocation, got both providers and panicked.
- **Nothing exercises the api-gateway → `oidc-authn-plugin` →
  `toolkit-auth::JwksKeyProvider` → `jsonwebtoken::verify` path** end-to-end, so
  a provider/JWT-verify regression only shows up as a runtime 500 in prod.

## What we want

Tests/checks that would have turned this into a **red CI check**, not a dev
incident. Concretely:

### 1. Gateway-auth smoke e2e (highest value)
Bring up the real stack the way prod is built (combined image build) —
`api-gateway` + `analytics` + `authenticator` + `fakeidp` + redis — then:
1. log in through fakeidp (the existing authenticator loop),
2. call an **authenticated** downstream route (e.g. `POST /analytics/v1/metrics/queries`)
   with the resulting cookie/JWT,
3. assert the response is **not 500** and the request was actually authenticated
   (i.e. the gateway verified the JWT).

This directly reproduces "verify a JWT in a combined build" and fails on a
provider misconfig. Fits alongside the existing `e2e-bronze-to-api` /
`functional-k3s` workflows.

### 2. Boot-time crypto-provider self-check
Have the toolkit-consuming services do a trivial JWT verify (or provider probe)
at startup and **fail readiness with a clear message** instead of panicking
per-request. Turns "mystery 500s under load" into "pod never goes Ready +
actionable log line." (May belong in gears-rust; file upstream if so.)

### 3. Cheap CI guard on the provider invariant
A lint step asserting **exactly one** jsonwebtoken provider is active, e.g.
[...]

PR #1839, opened 2026-07-22
ci(security): guard the jsonwebtoken crypto-provider invariant (#1727)

cyberantonz:
Addresses **#1727, part 3** (the cheap CI guard). Does not fully close the issue — parts 1 (combined-stack gateway-auth smoke e2e) and 2 (boot-time provider self-check) remain as heavier follow-ups.

## Why
Incident #1725: a dependency enabled jsonwebtoken's `rust_crypto` backend while another kept `aws_lc_rs`. Cargo unifies features across every workspace member that shares a build — and the release images compile `api-gateway` + `analytics` + `authenticator` together — so **both** providers turned on, which makes `jsonwebtoken::verify` **panic on every call**. It reached `main` as a runtime 500 on gateway auth *despite* green per-crate unit tests and a passing standalone authenticator e2e, because none of them reflect the combined build's feature set.

## What
A fast CI guard that asserts **exactly one** jsonwebtoken crypto provider is active in the unified backend build:

- `scripts/ci/check_jwt_provider.sh` — runs `cargo tree --edges features --invert jsonwebtoken` (no compile, just the resolved graph) and fails when the count of `{aws_lc_rs, rust_crypto}` enabled features isn't exactly 1. Usable locally too.
- `.github/workflows/ci.yml` — a new lightweight `jwt-provider-guard` job (checkout + toolchain + run). Ungated so a provider flip from *any* dependency bump is caught, not only a changed backend crate.

## Verification
```
$ ./scripts/ci/check_jwt_provider.sh
OK: exactly one jsonwebtoken crypto provider active: aws_lc_rs
```
Confirmed the detection logic flags the dual-provider case (would exit 1). `ci.yml` is valid YAML and yamlfmt-clean.

## Acceptance criteria touched
- [x] The provider-invariant guard (#3) runs in CI and fails on >1 provider.
- [ ] Gateway-auth smoke e2e (part 1) — follow-up
- [ ] Boot-time provider self-check (part 2) — follow-up

## Summary by CodeRabbit

* **Tests**
  * Added a CI validation check to ensure exactly one cryptographic provider is enabled for JWT functionality.
  * Builds now fail with a diagnostic message when the provider configuration is missing or ambiguous.


cyberantonz:
Reworked per feedback: inlined the check (dropped the script), and it now counts providers as jsonwebtoken's enabled features minus the known non-crypto ones — so an unknown *third* backend also trips it (a hardcoded aws_lc_rs/rust_crypto pair would've missed it). Fails when not exactly one, one-line message.

ktursunov (approved):

mitasovr (approved):

PR #1848, opened 2026-07-22
feat(authenticator): /auth/refresh rotation + session management + CSRF (nginx+auth step 10, PR 1/4)

cyberantonz:
Step 10 (auth surface completion) of the nginx+auth EPIC, **PR 1 of 4**. Delivers items **10.1** (`POST /auth/refresh`), **10.2** (session management), and **10.5** (CSRF) of `NGINX_BFF_10.md`.

This is the base of a 4-PR stack; the others build on it:
1. **this PR** — session surface (refresh + sessions + CSRF)
2. idp-lifecycle (back-channel logout + IdP refresher + janitor)
3. ratelimit + audit
4. docs (key-rotation runbook + PRD/DESIGN updates)

GitHub can't base a cross-fork PR on another fork branch, so all four target `main`; review this one first (the later PRs' diffs include these commits until this merges).

## What's here

- **10.1 `POST /auth/refresh` — rotation with grace (G10).** New CSPRNG token mapping; the superseded mapping's TTL drops to `refresh_grace_ms` (250 ms) — the expiring old mapping *is* the grace window (no swap keys). `expires_at = min(now + session_ttl, absolute cap)` across the record, key TTL, and per-user ZSET score. Stable `session_id` + linked JWT untouched. Stale-in-grace resolves to the same session (no re-rotation); past grace or either cap → 401 + clear cookie. Response `{expires_at, refresh_at}`, `refresh_at = expires_at − 90 s ± 60 s` (big-jitter, G8), re-jittered per call and shared with `/auth/me`.
- **10.2 session management.** `GET /auth/sessions` (created_at, expires_at, user_agent, ip, current — attribution captured at login); `DELETE /auth/sessions/{id}` (no-existence-oracle 404); `DELETE /auth/sessions`; admin `DELETE /auth/admin/users/{person_id}/sessions` — an `.authenticated()` op (host authn pipeline verifies the gateway JWT against the authenticator's own issuer), role-gated by `admin_revoke_roles` (default `session_admin`), delegating to the SDK `AuthenticatorClientV1::revoke_user_sessions`. Every revoke: token mappings + session + linked JWT + indexes in one pipeline.
- **10.5 CSRF.** Middleware on state-changing `/auth/*`: constant-time `X-CSRF-Token` vs the session record, `Origin`-allowlist fallback (`csrf_origins`, empty = fail closed), back-channel exempt. `GET /auth/csrf`; `/auth/me` echoes the token. SPA side is the insight-front PR.

## Review-fix commit

[...]

Issue #1593, opened 2026-07-03
nginx+auth 10 — auth surface completion: rotation, sessions mgmt, back-channel, IdP refresher, CSRF, janitor, audit


cyberantonz:
Part of EPIC #1583 (**nginx + authorization**). Step 10 of the 11-step implementation plan.

- /auth/refresh rotation with grace via token-mapping TTL (no swap keys) + big jitter (margin 90 s, window 120 s)
- /auth/sessions list/revoke + authenticated admin revoke; back-channel logout + jti replay guard
- IdP background refresher: due-ZSET schedule, semaphore, transient-vs-invalid_grant semantics (kill linked sessions), no_refresh_token_policy
- CSRF, Redis rate-limit layer 2, janitor, audit events, key-rotation runbook. Independently mergeable items — several small PRs. Depends on: 09 (can start after 04 for most items).

**Executable prompt** (self-contained instructions for a clean session): workspace file `cf/NGINX_BFF_10.md`.

Dev estimate: 3d

PR #1850, opened 2026-07-22
feat(authenticator): layer-2 rate limiting + audit events (nginx+auth step 10, PR 3/4)

cyberantonz:
Step 10 (auth surface completion) of the nginx+auth EPIC, **PR 3 of 4**. Delivers items **10.6** (layer-2 rate limiting) and **10.8** (audit events to Redpanda).

**Stacked on #1848 → PR 2.** Review from the ratelimit commit onward, or diff against `feat/auth-step10-idp-lifecycle`.

## What's here

- **10.6 rate limiting, layer 2 (§9.3, G8).** Atomic Lua token buckets in Redis keyed by the caller: the stable session on `/auth/refresh` (rotation doesn't reset it) and the OIDC `state` on `/auth/callback`; never IP (corporate NAT). Global live login-state cap on `/auth/login` (`asm:login_state_live` ZSET) → 429 before any state is written. Trips → 429 problem+json with a retry hint; the limiter fails **open** on a Redis error (the gateway's coarse per-IP zone still guards) while auth itself stays fail-closed.
- **10.8 audit events.** Every auth action (login ok/fail, refresh, logout, revoke single/all/admin, back-channel, `invalid_grant` kill, service-token issuance) publishes to `insight.audit.events` (Redpanda) with the platform envelope (schema/event_id/timestamp/correlation_id/tenant/actor/service/action/outcome/resource/details). Non-blocking: a bounded channel feeds an rdkafka `FutureProducer` (Kafka-compatible API only — the backend PRD's migration constraint; librdkafka via cmake-build, cmake added to the image); drops are counted (`auth_audit_dropped_total`), never coupled to auth availability; empty `audit.brokers` = disabled (log only).

## Review-fix commit

Applies **L4** (SHA-256-hash the token-bucket key so an attacker-chosen `state` can't inflate Redis or smuggle control chars) and **L5** (clamp the bucket timestamp forward so a clock step-back can't over-refill) from the review.

## Testing

Unit tests + clippy clean; full 8-loop e2e green (rate-limit loop asserts the refresh bucket trips per-session and the callback bucket trips per-state; audit emitter runs in disabled mode in the harness).

EPIC: constructorfabric/insight#1583 · #1593

### DIFF
--- a/charts/insight/templates/secrets.yaml
+++ b/charts/insight/templates/secrets.yaml
@@ -205,6 +205,9 @@ stringData:
   APP__gears__authenticator__config__redirect_uri:       {{ tpl (required "authenticator.oidc.redirectUri is required" .Values.authenticator.oidc.redirectUri) . | quote }}
   # Service-token endpoint audience callers sign assertions for.
   APP__gears__authenticator__config__service_tokens__audience: {{ printf "http://%s-authenticator.%s.svc.cluster.local:8093/internal/token" .Release.Name .Release.Namespace | quote }}
+  # Audit events to the platform Redpanda topic (step 10.8). Empty = disabled
+  # (structured log only) — the global redpanda.brokers wires it when set.
+  APP__gears__authenticator__config__audit__brokers: {{ .Values.redpanda.brokers | default "" | quote }}
 
 {{- if .Values.identity.deploy }}
 ---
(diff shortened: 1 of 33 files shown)


cyberantonz:
Superseded — the four stacked step-10 PRs are collapsed into a single PR (#1851). Same commits, one review, one merge. Closing this subset.

Issue #1551, opened 2026-07-01
deploy: identity service gets no ClickHouse config — persons-seed cannot connect


ktursunov:
## Problem
On a fresh gitops install the identity service is never given ClickHouse connection settings, so `persons-seed` (which must read `identity.identity_inputs`) fails immediately. The install script wires ClickHouse into the analytics-api config but omits it from the identity config.

## Context
`persons-seed` (identity service) reads `identity.identity_inputs` from ClickHouse and writes the `persons` / `org_chart` MariaDB tables. The gitops install composes a per-service config Secret; the identity one is `insight-identity-config`.

## Environment
- gitops install (`deploy/gitops`, `make apply-app-secrets` / `deploy-app`) with the identity service enabled.

## Steps to reproduce

**Fastest check — 1 command (no persons-seed needed):**
```bash
kubectl -n insight get secret insight-identity-config -o jsonpath='{.data}' \
  | tr ',' '\n' | grep -i clickhouse        # → prints NOTHING
```
The identity config has `IDENTITY__mariadb__url` and a tenant default, but **no `IDENTITY__clickhouse__*` keys at all**.

**Full path:**
1. Do a fresh gitops install so `compose-app-secrets.sh` composes `insight-identity-config`.
2. `POST /v1/persons-seed` as admin.
3. Operation fails with: `System.ArgumentException: clickhouse options must set either 'url' or 'host' (+ port/user/password/database)`.

**Expected:** identity has ClickHouse settings and the seed can read `identity_inputs`.
**Actual:** it throws immediately; `persons-seed` can never connect.

## Evidence
- [`compose-app-secrets.sh`](https://github.com/constructorfabric/insight/blob/main/deploy/gitops/scripts/compose-app-secrets.sh) reads `CH_HOST/PORT/USER/DB` and emits them only into `insight-analytics-api-config` (`ANALYTICS__clickhouse_*`). The [`insight-identity-config` block (#L158)](https://github.com/constructorfabric/insight/blob/main/deploy/gitops/scripts/compose-app-secrets.sh#L158) emits only `IDENTITY__mariadb__url` + `IDENTITY__identity__tenant_default_id` — no `IDENTITY__clickhouse__*`.
[...]

PR #1865, opened 2026-07-23
fix(identity): HOTFIX — persons-seed reads identity_inputs without tenant filter (#1550)

### LINKED ISSUE #1550: identity: persons-seed silently reads 0 rows — identity_inputs tenant is sipHash-derived but read by the raw tenant
## Problem
`persons-seed` — the step that turns ingested identity observations into the `persons` / `org_chart` tables — reads zero rows and completes "successfully" with all-zero counts. The cause is a tenant-representation mismatch: the dbt identity pipeline writes `identity.identity_inputs` with `insight_tenant_id` **hashed** (sipHash128 of the raw tenant), while the identity service reads `identity_inputs` filtered by the **raw** app tenant. The two never match, so nothing is consumed — and it fails silently.

## Context
Connectors land people-data in ClickHouse as `identity.identity_inputs` (one row per observed fact about a person: email, name, title, manager). A separate identity-service call, `POST /v1/persons-seed`, copies that into the MariaDB tables the app reads (`persons`, `org_chart`). "Tenant" is the org/customer ID that scopes all data.

## Environment
- A deployed Insight instance with an identity-contributing connector synced (e.g. bamboohr) so `identity.identity_inputs` is populated.
- `POST /v1/persons-seed` (identity service), called as admin, tenant = the app's configured tenant.

## Steps to reproduce

**Fastest check — 2 ClickHouse queries, no persons-seed needed:**
```sql
-- there IS data:
SELECT count() FROM identity.identity_inputs;                         -- non-zero (e.g. 164)
-- but none under the tenant persons-seed / the app use:
SELECT count() FROM identity.identity_inputs
WHERE insight_tenant_id = '<app tenant, e.g. 00000000-df51-5b42-9538-d2b56b7ee953>';   -- 0
```
The second count is **0** while the first is non-zero — `persons-seed` runs exactly that `WHERE`, so it reads nothing. Why:
```sql
SELECT DISTINCT insight_tenant_id FROM identity.identity_inputs;      -- a hash, e.g. 3accefc1-…
SELECT toString(toUUID(UUIDNumToString(sipHash128('<app tenant>'))));  -- == that same hash
```

**Full path (shows the user-visible impact):**
1. Sync an identity connector so `identity.identity_inputs` has rows.
2. `POST /v1/persons-seed` (mode `link-by-email`) as admin for the app tenant.
3. Operation returns `status: completed` but `accounts_read: 0`.
4. MariaDB `persons` / `org_chart` stay empty; nobody resolves / the org chart is blank — **no error anywhere**.

**Expected:** the seed reads the synced observations and populates `persons` / `org_chart`.
**Actual:** 0 rows read; tables empty; silent success.

## Evidence
- Every `identity.identity_inputs` row carries one hashed tenant (example `3accefc1-c546-8bae-d64e-35be081d139e`), not the app tenant.
- App tenant example `00000000-df51-5b42-9538-d2b56b7ee953`; in ClickHouse `sipHash128(<app-tenant>)` → that exact hashed value — proving `identity_inputs.insight_tenant_id = sipHash128(rawTenant)`.
- `persons-seed` summary: `accounts_read=0, accounts_linked_by_email=0, accounts_minted_new=0, observations_inserted=0`.

## Root cause
- **Producer** hashes the tenant — [`identity_inputs_from_history.sql#L79`](https://github.com/constructorfabric/insight/blob/main/src/ingestion/dbt/macros/identity_inputs_from_history.sql#L79):
  ```
  toUUID(UUIDNumToString(sipHash128(coalesce(tenant_id, '')))) AS insight_tenant_id
  ```
  (documented in that macro as intentional for cross-source join stability; flagged TEMPORARY / "not looked up from a table yet").
- **Consumer** uses the raw tenant — [`ClickHouseIdentityInputsReader.cs#L38`](https://github.com/constructorfabric/insight/blob/main/src/backend/services/identity/src/Insight.Identity.Infrastructure/ClickHouse/ClickHouseIdentityInputsReader.cs#L38): `... FROM identity.identity_inputs WHERE insight_tenant_id = {tenant:String} ...`, with `tenant` bound to the raw app tenant (`tenantId.ToString("D")`, from `AppOptions.tenant_default_id` / JWT).
- Nothing reconciles the two representations, so the filter never matches. The service also *writes* `persons` under the raw tenant, so calling the seed with the hashed tenant would read the rows but file `persons` under a tenant the gateway/JWT never presents — neither single-tenant call is consistent end to end.

## Other notes
The silent success (completed, zero counts, no error) is what makes this hard to catch in the field.

mozhaev-dev:
Fixes #1550.

> ⚠️ **HOTFIX** — pre-release, temporary, identity-scoped. No config /
> ingestion / re-seed change. To be removed once the tenant representation
> is unified end to end (tracked as follow-up).

## Problem

`POST /v1/persons-seed` could silently read **0 rows** and complete
"successfully" with all-zero counts, leaving `persons` / `org_chart` empty.
Root cause: a tenant-representation mismatch on `identity.identity_inputs`.

- **Producer** (dbt) writes `insight_tenant_id` **hashed** —
  `sipHash128(<connector's configured insight_tenant_id string>)` rendered as
  a dashed UUID (`identity_inputs_from_history.sql`, documented there as a
  TEMPORARY cross-source join key).
- **Consumer** (`ClickHouseIdentityInputsReader`) filtered by the caller's
  tenant, which never equals the hashed stored value → 0 rows → silent success.

An earlier revision of this PR matched `IN (tenant, sipHash(tenant))`. Live
verification on the dev stand showed that is still not enough: the connector
config's `insight_tenant_id` is a free-form string (`'default'` on dev) that
need not equal the caller's tenant at all, so **no formula derived from the
caller's tenant can reliably match the stored hash** — a three-way mismatch
(JWT tenant ≠ connector string ≠ its hash) still read 0 rows.

## Fix

Drop the tenant predicate from the `identity_inputs` read entirely:

- Insight deployments are **single-tenant** — every `identity_inputs` row
  belongs to the deployment, whatever tenant string the connector stamped.
- The seed's **write side is unchanged and stays tenant-scoped**:
  `PersonsSeedService` binds the request tenant for every row it writes and
  never uses the row's tenant, so output lands under the caller's tenant and
  tenant-scoped deletes/rebuilds are untouched.
- `StreamAsync(tenantId, …)` keeps its signature so the interface (and the
  Rust port tracking it) stays stable for when the filter comes back.

This is a temporary, identity-scoped shim to restore the seed before the
release without touching configs, ingestion, or seeded data. Restore the
filter once the producer resolves real tenant UUIDs instead of hashing.

## Verified live (dev stand `insight-dev-vhc`, patched image, synthetic admin tenant, all changes reverted after)

| build | accounts_read | result |
|---|---|---|
| current image (old `= tenant` filter) | **0**, status `completed` | reproduces #1550 — silent empty seed |
| this PR | **4758** (all 36,581 rows, 5 sources) | 4629 persons minted, 25,206 observations, org_chart 4225 rows |

- Output was written **only** under the caller's tenant; the pre-existing
  tenant's `persons` (42,698) and `org_chart` (4,719) were byte-identical
  before/after — write isolation holds without the read filter.
- Counts line up with the last known-good seed run (2026-07-20:
  accounts_read=4740, before a connector re-sync re-stamped the data and
  broke tenant matching).

## Follow-ups (not in this PR)

- **Silent success**: a seed completing with `accounts_read = 0` still emits
  no warning — worth a warn/metric.
- **Rust port**: the `identity-resolution` reader (`infra/identity_inputs.rs`)
  carries the same tenant filter and needs the same change before cutover.
- **Chart gap (#1551)**: the helm chart wires no `IDENTITY__clickhouse__*`
  for identity, so persons-seed fails at startup on a fresh chart deploy —
  must be closed for the seed to run at all in the release.
- **Proper fix**: unify the tenant representation end to end (producer
  resolves real tenants instead of hashing), then restore the filter.

### DIFF
--- a/src/backend/services/identity/src/Insight.Identity.Infrastructure/ClickHouse/ClickHouseIdentityInputsReader.cs
+++ b/src/backend/services/identity/src/Insight.Identity.Infrastructure/ClickHouse/ClickHouseIdentityInputsReader.cs
@@ -20,10 +20,27 @@ public sealed class ClickHouseIdentityInputsReader : IIdentityInputsReader
     // DELETE). DELETE rows are streamed too; they flag the account as
     // closed but never produce a persons observation.
     // identity_inputs stores insight_tenant_id / insight_source_id as
-    // Nullable(String) (dashed UUID text), so the tenant filter is a
-    // plain string compare. The `_str` aliases avoid shadowing the
-    // source columns referenced in WHERE (a same-name SELECT alias
-    // would otherwise be substituted into the WHERE clause).
+    // Nullable(String) (dashed UUID text). The `_str` aliases avoid
+    // shadowing the source columns referenced in WHERE (a same-name
+    // SELECT alias would otherwise be substituted into the WHERE clause).
+    //
+    // HOTFIX (#1550) — TEMPORARY, identity-scoped, pre-release. The dbt
+    // producer writes insight_tenant_id *hashed* — sipHash128(rawTenant)
+    // rendered as a dashed UUID string (identity_inputs_from_history.sql,
+    // documented there as a TEMPORARY cross-source join key). So a plain
+    // `= {tenant}` matched nothing when the service is configured with the raw
+    // tenant, and persons-seed silently read 0 rows.
+    //
+    // Match BOTH representations of the caller's tenant — the raw value AND its
+    // sipHash — so the read works whether the deployment is configured with the
+    // raw tenant (the hash term matches the hashed data) OR already with the
+    // hashed value (the raw term matches; some envs were worked around that
+    // way). This is strictly additive to the previous `= {tenant}`, so it
+    // cannot regress a currently-working deployment and needs no config /
+    // ingestion / re-seed change. Both terms derive from the single bound
+    // tenant, so tenant isolation holds (a collision would need a real tenant
+    // UUID to equal a sipHash output). Remove once the tenant representation is
+    // unified end to end.
     private const string StreamSql = """
         SELECT
             toString(insight_tenant_id) AS tenant_id_str,
@@ -35,7 +52,10 @@ public sealed class ClickHouseIdentityInputsReader : IIdentityInputsReader
             _synced_at,
             operation_type
         FROM identity.identity_inputs
-        WHERE insight_tenant_id = {tenant:String}
+        WHERE insight_tenant_id IN (
+                  {tenant:String},
+                  toString(toUUID(UUIDNumToString(sipHash128({tenant:String}))))
+              )
           AND operation_type    IN ('UPSERT', 'DELETE')
           AND value             IS NOT NULL
           AND value             != ''


aleksdotbar (approved):

mitasovr (approved):

src/backend/services/identity/src/Insight.Identity.Infrastructure/ClickHouse/ClickHouseIdentityInputsReader.cs
```diff
@@ -35,7 +52,10 @@ public sealed class ClickHouseIdentityInputsReader : IIdentityInputsReader
             _synced_at,
             operation_type
         FROM identity.identity_inputs
-        WHERE insight_tenant_id = {tenant:String}
+        WHERE insight_tenant_id IN (
```
cyberantonz, line 55: But the reality with this tenant is differs from real tenant_ids. Are the insight_tenant_id in clickhouse contains proper tenant ids?

mozhaev-dev:
Clarification: this PR merged the `IN (raw, sipHash(raw))` shim (`f3deaf35`). The PR description was edited (just before the merge happened) to describe a stronger follow-up that did **not** land here: dropping the tenant filter entirely. That change is now #1871 — the shim in this PR still reads 0 rows on deployments where the connector config's `insight_tenant_id` string differs from the caller's tenant (verified live on dev). Please see #1871 for the actual pre-release fix.

Issue #1687, opened 2026-07-08
authenticator: resolve tenant memberships from Identity (not the id_token)


cyberantonz:
Follow-up from #1686 (nginx+auth step 04).

The authenticator's `PersonResolver` sources a session's `tenants` from the **id_token claim** (which fakeidp supplies) because Identity's `GET /v1/persons/{email}` returns `insight_source_id` (person_id) but no tenant memberships. Real IdPs won't carry a `tenants` claim.

**To close:** add an Identity membership API (email/person → tenant memberships) and make it the `tenants` authority in `PersonResolver::resolve`; drop the id_token-tenants fallback. Code marker: `identity::resolve` + the module doc-comment.

**Not in scope here:** first-admin bootstrap, universe-admin designation, and RBAC/ACL were removed from step 04 and belong to a **separate universe-admin initiative** (local dev seeds the persons table; unknown persons are denied 403). File/track that separately.

Refs #1583.

PR #1874, opened 2026-07-23
fix(authenticator): tenant-scope the person-lookup service token

cyberantonz:
The service JWT the authenticator mints for the login-time `/internal/persons/by-email/{email}` lookup hardcoded `tenant_id = ""` (a cross-tenant service token). But the tenant is already known at that point — `exchange_code_pkce` resolves it from the id_token claim (or `default_tenant_id`) **before** `resolve()` runs — so the lookup token should carry it.

This threads the resolved `tenant_id` into `mint_service_token`, so the token is scoped to the caller's tenant (empty only when the IdP named no tenant and no `default_tenant_id` is set).

**Authenticator-only (producer half).** Identity still resolves by email any-tenant today; having it *scope* the `by-email` query to the token's tenant is the separate #1687 work. This change makes the tenant travel so that lands cleanly, and it correctly attributes the tenant in audit/trace meanwhile.

Verified: clippy clean, authenticator unit tests pass.

Ref: #1583, #1687

### DIFF
--- a/src/backend/services/authenticator/src/api/handlers.rs
+++ b/src/backend/services/authenticator/src/api/handlers.rs
@@ -150,6 +150,25 @@ pub async fn callback(
         }
     };
 
+    // A session without a tenant is unusable — the gateway JWT's `tenant_id` is
+    // the sole tenant authority and downstream fails closed without it. If the
+    // id_token named no tenant and no `default_tenant_id` is configured, deny
+    // the login here rather than minting a dead session (and never issue a
+    // tenant-less JWT — enforced again at the signing chokepoint).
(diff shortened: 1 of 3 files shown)


cyberantonz:
Pushed `680855c`: enforce the **no-tenant-less-JWT** invariant per review — `KeyStore::sign()` now refuses any `GatewayClaims` with an empty `tenant_id` (the single chokepoint every mint path passes through), and `/auth/callback` denies the login up front (403, audited `login_denied_no_tenant`) when the id_token carried no tenant and no `default_tenant_id` is set — rather than minting a dead session. Added a unit test for the sign guard.

mitasovr (approved):

PR #1875, opened 2026-07-23
fix(m365): treat unknown Teams license as unlicensed; adopt semver

### LINKED ISSUE #736: M365 silver: unlicensed users (isLicensed=false) leak into class_collab_meeting_activity

mitasovr:
Follow-up to #736 (shipped in #1869), addressing the CodeRabbit review on that PR.

## 1. Teams license filter — exclude unknown (NULL) too

CodeRabbit flagged that the merged Teams filter kept `isLicensed IS NULL` rows instead of excluding them as the issue proposed. Switched both Teams feeders:

```diff
- AND coalesce(isLicensed, true) = true
+ AND coalesce(isLicensed, false) = true
```

Now an unlicensed **or** unknown-license row is dropped; only an explicit `isLicensed=true` row is kept (`m365__collab_meeting_activity`, `m365__collab_chat_activity`).

- The e2e teams fixture template now defaults `isLicensed: true` so existing fixtures stay counted.
- The regression test [`collab_meeting_unlicensed_excluded.test.yaml`](https://github.com/constructorfabric/insight/blob/main/src/ingestion/tests/e2e/metrics/collab_meeting_unlicensed_excluded.test.yaml) gains an `isLicensed=null` case asserting exclusion (alice `true`→3, frank `false`→null, grace `null`→null).

The **email/onedrive/sharepoint** feeders keep their conservative `assignedProducts` filter (drop only explicitly-empty lists, keep NULL): those Graph reports expose no `isLicensed`, and the JSON-string serialization is inferred rather than verified, so a NULL there means "unknown/unsynced" and must not nuke the whole feeder.

## 2. descriptor.yaml → strict semver (ADR-0015), step 1 of 2

CodeRabbit also noted the incremental models don't retro-delete rows materialized before the #736 filter. The blessed cleanup per [ADR-0015](https://github.com/constructorfabric/insight/blob/main/docs/components/airbyte-toolkit/specs/ADR/0015-semver-and-full-refresh.md) is a **MAJOR-bump full-refresh** over `dbt_select` (`tag:m365+`) — but that only fires when the *current deployed* version is already semver.

m365 is currently `version: "2026.05.04"` (non-semver), so `classify_bump.py` would treat any bump as a `migration` (republish + re-discover, **no** full-refresh). This PR is **step 1**: migrate to `1.0.0` to establish the semver baseline. A **step 2** PR (`1.0.0 → 2.0.0`) will then dispatch the one-shot `dbt --full-refresh` that purges the pre-#736 unlicensed rows.

## Testing

Dockerized e2e rig, affected Teams metrics:

```
4 passed, 140 deselected in 290.30s
```
(new 3-way exclusion test + `collab_meetings` / `collab_messages` / `collab_meeting_free_days`)

Refs #736. Follows #1869.

## Summary by CodeRabbit

- **Bug Fixes**
  - M365 collaboration chat and meeting activity now excludes records where licensing status is unknown or explicitly unlicensed.
  - Standard Teams activity test data now defaults to licensed status, improving consistency in reporting.

- **Tests**
  - Expanded coverage verifies that both unlicensed and unknown-license users are excluded from meeting metrics.

- **Other**
  - Updated the M365 connector version to `1.0.0`.


ktursunov (approved):

src/ingestion/tests/e2e/metrics/templates/m365_teams.yaml
```diff
@@ -19,7 +19,11 @@ templates:
     reportPeriod: "D1"
     reportRefreshDate: null
     lastActivityDate: null
-    isLicensed: null
+    # Default to licensed: the meeting/chat feeders drop rows where
+    # coalesce(isLicensed, false) != true, so an unset flag would otherwise
+    # exclude every fixture row. Tests exercising the unlicensed path set this
+    # explicitly to false (or null). See #736.
+    isLicensed: true
```
ktursunov, line 26: let's check all cases - true/false/null

Issue #1885, opened 2026-07-23
Connectors never sync on a fresh install when Airbyte runs in a separate namespace


ktursunov:
## Summary
On a fresh install where Airbyte runs in a different namespace than the app, connectors never sync. The reconcile and airbyte-sync jobs read Airbyte's credentials from a Secret named `airbyte-auth-secrets` in the app namespace. That Secret only exists in Airbyte's own namespace. Nothing copies it into the app namespace, so the jobs cannot authenticate to Airbyte and no data is ingested.

## Impact
No connector syncs on a split-namespace install. Bronze never fills, so the whole dashboard stays empty even after everything else is healthy.

## Steps to Reproduce
1. Install with Airbyte as a separate release in its own namespace (app in `insight`, Airbyte in `insight-infra`).
2. `kubectl -n insight get secret airbyte-auth-secrets` → `NotFound`.
3. `kubectl -n insight-infra get secret airbyte-auth-secrets` → present.
4. The reconcile / airbyte-sync jobs reference `airbyte-auth-secrets` via `secretKeyRef` in the `insight` namespace, so they cannot read the credentials.

**Expected:** the sync jobs authenticate to Airbyte and reconcile connectors.
**Actual:** the auth Secret is absent in the app namespace, so authentication fails.

The credentials themselves are valid — this is placement, not validity. Reading `instance-admin-client-id` / `instance-admin-client-secret` from the Secret in Airbyte's namespace and calling the Airbyte token endpoint returns `HTTP 200` with an `access_token`.

## Root Cause
The chart reads the Airbyte auth Secret with a `secretKeyRef` in the app namespace ([airbyte-sync.yaml#L145](https://github.com/constructorfabric/insight/blob/main/charts/insight/templates/ingestion/airbyte-sync.yaml#L145)). A pod's `secretKeyRef` can only read a Secret in its own namespace. The Secret is created by the Airbyte chart in Airbyte's namespace, and the chart assumes Airbyte is in the same namespace as the app ([values.yaml#L146](https://github.com/constructorfabric/insight/blob/main/charts/insight/values.yaml#L146) — "created by the Airbyte chart itself"). Nothing replicates it into the app namespace.

The fix is deploy-side. The install must place `airbyte-auth-secrets` (or at least its `instance-admin-*` and `jwt-signature-secret` keys) in the app namespace. The chart cannot do this — Kubernetes has no cross-namespace Secret references.

## Notes
related to #844

PR #1913, opened 2026-07-24
fix(ingestion): read airbyte-auth-secrets from the Airbyte namespace

### LINKED ISSUE #1906: Reconcile token helper reads the Airbyte auth Secret from the app namespace instead of the Airbyte namespace
## Summary

The reconcile token helper (`ab_get_token` in `lib/airbyte.sh`) issues a `kubectl get secret airbyte-auth-secrets` against the app namespace (`$INSIGHT_NAMESPACE`). When Airbyte runs in a separate namespace, that Secret does not exist in the app namespace, the token fetch fails, and the entire reconcile run aborts before any connector is created or synced.

## Impact

- Connector reconcile is completely broken on any cluster where Airbyte runs in a namespace different from the app — no Airbyte sources or connections are ever created, so ingestion never starts.
- The run aborts with a message pointing at the app namespace (`failed reading secret/airbyte-auth-secrets in ns <app>`), which misdirects diagnosis toward RBAC in the wrong namespace.

## Steps to Reproduce

1. Deploy with Airbyte as a separate release in its own namespace (app and Airbyte in different namespaces).
2. Enable ingestion templates and let the reconcile workflow run (or trigger it manually).
3. Observe the reconcile pod exit non-zero with `ab_get_token: kubectl failed reading secret/airbyte-auth-secrets in ns <app>`.

## Root Cause

`src/ingestion/reconcile-connectors/lib/airbyte.sh` — `ab_get_token()` reads `airbyte-auth-secrets` from `$INSIGHT_NAMESPACE`, assuming Airbyte shares the app namespace. When Airbyte is deployed to its own namespace the Secret is not present in the app namespace, so the `kubectl get` fails. The helper needs to read the Secret from the Airbyte namespace (a separately configurable namespace, defaulting to the app namespace so single-namespace installs are unaffected), and the reconcile ServiceAccount needs `get` permission on that Secret in the Airbyte namespace.

cyberdima:
## What

`ab_get_token` read `airbyte-auth-secrets` from `$INSIGHT_NAMESPACE`, assuming Airbyte shares the app namespace. When Airbyte runs in its own namespace the Secret is absent there, the token fetch fails, and the whole reconcile run aborts before any connector is created or synced.

## Change

- **airbyte.sh**: new `AIRBYTE_NAMESPACE` env (defaults to `INSIGHT_NAMESPACE`, so single-namespace installs are unaffected); `ab_get_token` reads the Secret from it.
- **chart**: new `airbyte.namespace` value (`""` = `.Release.Namespace`), `insight.airbyte.namespace` helper, `AIRBYTE_NAMESPACE` rendered on the reconcile CronWorkflow, and a conditional Role/RoleBinding granting the reconcile ServiceAccount `get` on the Secret in the Airbyte namespace — rendered only when it differs from the release namespace (single-namespace installs render nothing extra).

## Scope

`bash -n` clean; `helm lint` passes.

Refs #1906

### DIFF
--- a/charts/insight/templates/_helpers.tpl
+++ b/charts/insight/templates/_helpers.tpl
@@ -122,6 +122,10 @@ http://{{ .Values.airbyte.releaseName }}-airbyte-server-svc.{{ include "insight.
 {{- end -}}
 {{- end -}}
 
+{{- define "insight.airbyte.namespace" -}}
+{{- .Values.airbyte.namespace | default .Release.Namespace -}}
+{{- end -}}
+
 {{/*
 ==============================================================================
  APP SERVICE HOSTS
--- a/charts/insight/templates/_helpers.tpl
+++ b/charts/insight/templates/_helpers.tpl
@@ -122,10 +122,6 @@ http://{{ .Values.airbyte.releaseName }}-airbyte-server-svc.{{ include "insight.
 {{- end -}}
 {{- end -}}
 
-{{- define "insight.airbyte.namespace" -}}
-{{- .Values.airbyte.namespace | default .Release.Namespace -}}
-{{- end -}}
-
 {{/*
 ==============================================================================
  APP SERVICE HOSTS
--- a/charts/insight/templates/ingestion/reconcile-cron.yaml
+++ b/charts/insight/templates/ingestion/reconcile-cron.yaml
@@ -67,18 +67,11 @@ spec:
               value: {{ include "insight.airbyte.url" . | quote }}
             - name: INSIGHT_NAMESPACE
               value: {{ .Release.Namespace | quote }}
-            # Airbyte's auth Secret lives in Airbyte's namespace, which may
-            # differ from the app's (issue #1885). lib/airbyte.sh reads it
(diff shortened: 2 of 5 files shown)


cyberdima:
Closing as superseded by #1935, which already merged the same fix (cross-namespace `airbyte-auth-secrets` read) to `main` under issue #1885, via a more complete implementation (`AIRBYTE_AUTH_SECRET_NAME`/`AIRBYTE_AUTH_SECRET_NAMESPACE`/`AIRBYTE_CLIENT_ID_KEY`/`AIRBYTE_CLIENT_SECRET_KEY`, dedicated `airbyte-auth-rbac.yaml`, and binds the `argo-workflow` fallback SA in addition to the reconcile SA).

Rebased this branch onto `main` to confirm: after resolving conflicts in favor of `main`'s version, the diff across all 5 touched files (`_helpers.tpl`, `reconcile-cron.yaml`, `reconcile-rbac.yaml`, `values.yaml`, `airbyte.sh`) is zero. Nothing left to merge.

PR #1914, opened 2026-07-24
fix(gitops): scope the Argo Workflows controller to the app namespace too

### LINKED ISSUE #1907: Argo Workflows controller is installed watching only the infra namespace, so the reconcile CronWorkflow in the app namespace never runs
## Summary

The gitops target that installs the Argo Workflows controller scopes it to a single watched namespace (the infra namespace). The reconcile CronWorkflow is rendered into the app namespace. A controller that watches only the infra namespace cannot see or run a CronWorkflow in the app namespace, so reconcile never fires — silently, with no error.

## Impact

- Ingestion reconcile never runs on any cluster that separates the app and infra namespaces. The CronWorkflow object exists but is never scheduled (`lastScheduled: never`) and nothing logs an error.
- The workflow-executor RBAC is applied only in the infra namespace, so even a widened controller would fail to run workflow pods in the app namespace for lack of permissions.

## Steps to Reproduce

1. On a cluster that separates the app namespace from the infra namespace, install the Argo Workflows controller via the gitops target.
2. Deploy the app chart with ingestion templates enabled (this renders the reconcile CronWorkflow into the app namespace).
3. Observe the controller never picks up the CronWorkflow: its `.status` stays empty, no controller log mentions it, and no reconcile runs.

## Root Cause

`deploy/gitops/Makefile` — the `system-argo` target sets `controller.workflowNamespaces` to the infra namespace only, and applies the workflow-executor Role/RoleBinding only in the infra namespace. The environment inventories declare a two-namespace split (services vs. infra) as the standard topology, and the reconcile CronWorkflow lives in the services (app) namespace. The controller must watch both namespaces and hold executor RBAC in both. The RBAC template also conflates the ServiceAccount's home namespace with the namespace the Role must be created in, so it cannot express "SA in namespace A, Role in namespace B."

cyberdima:
## What

`system-argo` installed the Argo Workflows controller watching only `NS_INFRA` and applied the workflow-executor RBAC only there. The reconcile CronWorkflow renders into `NS_APP` (the standard services-vs-infra split declared in every `environments/*/inventory.yaml`), so a controller scoped to `NS_INFRA` could never see or run it — reconcile silently never fired.

## Change

- **Makefile**: set `controller.workflowNamespaces[1]=$(NS_APP)` alongside `[0]=$(NS_INFRA)`, and apply the executor RBAC in both namespaces.
- **argo-rbac.yaml.template**: new `${SA_NAMESPACE}` placeholder decoupled from `${NAMESPACE}`, so the RoleBinding subject points at the ServiceAccount's home namespace (always `NS_INFRA`) while the Role/RoleBinding itself is created in each watched namespace. Header comment corrected (it previously claimed all components share one namespace).

## Scope

Two files: `deploy/gitops/Makefile`, `deploy/gitops/bootstrap/argo-rbac.yaml.template`.

Refs #1907


cyberantonz (approved):

cyberantonz:
https://github.com/constructorfabric/insight/issues/1907

PR #1947, opened 2026-07-27
fix(ci): publish-chart missing the workflow_dispatch full-rebuild fallback

mozhaev-dev:
## Summary
- Follow-up to #1939/#1943. Every per-service build job (e.g. `backend-identity-resolution`) triggers on `needs.changes.outputs.<service> == 'true' || (github.event_name == 'workflow_dispatch' && inputs.frontend_tag == '')` — a manual full-rebuild dispatch (empty `frontend_tag`) rebuilds every image regardless of what `paths-filter` detected changed.
- `publish-chart`'s own OR-list only checked the `changes.outputs.*` flags plus `inputs.frontend_tag != ''` — it had no equivalent full-rebuild fallback. So a manual `workflow_dispatch` would rebuild every backend image but never actually publish the umbrella chart bundling them.
- Found while trying to verify #1943 (the `identity_resolution` flag missing from this same OR-list) without waiting for an unrelated merge to naturally touch identity-resolution/analytics/etc — a manual full-rebuild dispatch seemed like the obvious way to trigger it now, until this gap turned up too.

## Test plan
- [ ] YAML validated locally (`yaml.safe_load`)
- [ ] A manual `workflow_dispatch` (empty `frontend_tag`) should now actually run `publish-chart` and publish a fresh umbrella chart


mitasovr (approved):

ktursunov (approved):

.github/workflows/build-images.yml
```diff
@@ -1211,6 +1211,7 @@ jobs:
         || needs.changes.outputs.toolbox      == 'true'
         || needs.changes.outputs.umbrella     == 'true'
         || inputs.frontend_tag != ''
+        || (github.event_name == 'workflow_dispatch' && inputs.frontend_tag == '')
```
cyberantonz, line 1214: It is wrong approach. If the image was built, the lines 1196-1203 will work. 
So, please wait, looks like you do something wrong.

mozhaev-dev: How to do better?

cyberantonz: Let's talk about it :) Maybe i do not understand clearly what are you trying to solve.

cyberantonz:
Let's drop this PR

Issue #1960, opened 2026-07-28
Identity resolution: replace internal by-email login lookup with (provider, external_id)


mozhaev-dev:
## Goal

Replace the internal login-bootstrap resolution endpoint so it resolves a
person by **IdP provider + the provider's internal user id** instead of by
email, and remove the email-based endpoint and all its dependents.

## Background

`identity-resolution` currently exposes a service-only endpoint used solely
by `authenticator` at login bootstrap:

```
GET /internal/persons/by-email/{email}
```

- Handler: `src/backend/services/identity-resolution/src/api/handlers.rs:156` (`internal_person_by_email`)
- Repo query: `persons_repo::resolve_person_id_by_email_any_tenant` (`persons_repo.rs:80`)
- Only caller: `authenticator/src/identity.rs` (`IdentityPersonResolver::lookup_person_id`, line ~127)
- Also referenced by: `authenticator/tests/run-e2e.sh` + `authenticator/tests/identity-stub.py` (e2e stub), and
  `src/ingestion/tests/e2e/identity/test_internal.py` (contract tests)
- Legacy .NET twin (being phased out): `src/backend/services/identity/src/Insight.Identity.Api/Endpoints/PersonsEndpoints.cs:173`

We don't need email-based resolution here — we need to resolve by
`(idp_provider, idp_user_id)`, e.g. `(ms-entra, <Entra oid>)`.

## What already exists (from investigation)

- The public `POST /v1/profiles` resolver (`handlers.rs:189`, `resolve_person_ids`) already supports
  `value_type: "id"` scoped by `insight_source_type` + `insight_source_id` — this is exactly the
  `(provider, external_id)` shape we need; we're extending the *internal* any-tenant endpoint to match it,
  not inventing a new data model.
- For **Entra**, the data already flows end-to-end: the `ms-entra` connector syncs the Entra Object ID
  (`oid`, the JWT `oid` claim) as `id` — via `identity_inputs_from_history` this lands in ClickHouse
  `identity.identity_inputs` as `value_type='id', insight_source_type='ms-entra', value=<oid>`, and the
  persons-seed job (`domain/seed_service.rs::run_seed`) writes it into MariaDB `persons`. So an Entra `oid`
  can already be resolved through the generic `"id"` path today, given a fresh seed run.
- **Okta has no connector at all** — out of scope here, tracked separately.

## Scope

1. Add a new internal any-tenant resolver:
[...]

Issue #1941, opened 2026-07-27
Admin 'view as another user' override (`?__override=<email>`) no longer works


dzarlax:
## Summary

The admin/dev "view the dashboard as another user" override — used as a URL
query param, e.g. `https://insight-dev.constr.dev/?__override=ao%40constructor.tech`
— no longer works. Setting `?__override=<email>` used to switch the viewer
identity so an operator could see the portal exactly as that person sees it
(their scope, their metrics). Today it has no effect.

## Repro

1. Open the portal with `?__override=<some-other-user-email>`.
2. Expected: the dashboard renders as that user (their person context / scope).
3. Actual: the override is ignored — the viewer stays the authenticated user.

## Likely root cause (to confirm)

`__override` is not present anywhere in the current code (grep across both
`insight` and `insight-front` returns nothing), so the mechanism appears to have
been dropped rather than intentionally re-implemented. Two changes plausibly
removed it — confirm which:

1. **Gateway authz hardening (#1769).** The fix there makes viewer identity come
   **only** from the verified, gateway-authored identity and **strips inbound
   client `X-Insight-*` headers**. The old override almost certainly rode on a
   client-set `X-Insight-Person-Id` (or equivalent), which is now stripped/ignored
   by design. So the legitimate "view-as" feature was closed as collateral of
   closing the spoofing hole.
2. **Frontend rewrite.** The HAI3 frontend was removed wholesale during the
   TanStack/React-Query rebrand; if the `?__override=` query-param handling lived
   in that FE code, it was deleted and never re-added.

## Ask

Restore a **sanctioned** "view as another user" override that works with the
current gateway-authored identity model and **does not reintroduce #1769**:

- Authorization decided **server-side** from the verified principal (admin role
  read from the validated JWT / gateway-authored headers), **not** from a
  client-supplied identity header.
- An authorized admin may then act-as another `person_id`; the acting-as target
  is applied server-side (gateway/identity), so a non-admin cannot spoof it by
  setting a query param or header.
- Scope of the impersonated view = the target's own scope. Consider surfacing an
  explicit "viewing as X" banner and auditing each act-as (who viewed as whom).

## Relations

- Security constraint / root-cause candidate: #1769 (derive authz from verified
  identity, gateway strips inbound `X-Insight-*`).
- Viewer/identity surface: identity-resolution service; #1873 (identity mapping
  engine), #1881 (person attributes / slicing).

## Notes

[...]

PR #1989, opened 2026-07-28
fix(identity): login-bootstrap resolves by IdP source_type + external id, not email

mozhaev-dev:
## Summary

- Replaces the internal login-bootstrap lookup with two SEPARATE, non-overlapping service-only contracts (no shared `value_type` dispatch, so a login that lacks its external id can never silently fall through to email):
  - `GET /internal/persons/by-external-id?source_type=...&external_id=...` — login bootstrap ONLY, scoped to the configured IdP's `source_type` (e.g. `ms-entra`) + its source-native external user id (Entra: the `oid` claim).
  - `GET /internal/persons/by-email-override?email=...` — the authenticator's admin `__override` (view-as, #1941) lookup ONLY.
- `authenticator`'s `IdpIdentity` now carries an explicit `ResolveTarget::ExternalId(String) | Email(String)` enum instead of inferring the resolution mode from field emptiness, and **fails closed** (login refused) if the configured `idp.external_id_claim` is absent from the id_token.
- The authenticator is now unconditionally wired to `identity-resolution` (Rust) for this lookup — the `.NET identity` twin was intentionally NOT given an equivalent endpoint (out of scope by explicit direction; **`.NET identity` service is untouched by this PR**). Wired through docker-compose, the Helm chart (new `insight.validate` check + `identityResolution.deploy` default flipped to `true`), and the gitops `compose-app-secrets.sh` script (separate `AUTHENTICATOR_IDENTITY_URL` + required `idp.source_type`/`idp.external_id_claim`).
- Fixes Keycloak-mode dev/demo seeding: the whole 25-person roster can log in now (not just the dev lead), since `gen-realm.py` pins every realm user's `sub` to their own roster uuid.
- `deploy/seed/identity.py`'s login-id seed is idempotent via an explicit existence check (not `INSERT IGNORE` — migration 004 put `created_at` in `persons`' unique key, so re-runs no longer dedupe on their own).

Part of #1960 (see the issue's "Update (post-implementation)" section for the final-scope writeup, including why the email lookup is intentionally kept for `__override`).

## Test plan

- [x] `cargo fmt --all -- --check` (backend workspace) — clean
- [x] `cargo test -p authenticator -p identity-resolution` — 54 + 58 unit tests pass
- [x] `cargo clippy -p authenticator -p identity-resolution --all-targets -- -D warnings` — clean
- [x] Full `src/backend/services/authenticator/tests/run-e2e.sh` (real fakeidp + authenticator + Redis stack) — all scenarios pass, incl. login and `__override`
- [x] `python3 -m unittest deploy/seed/test_identity.py -v` — 2/2 pass (idempotency + Keycloak whole-roster seeding)
- [x] `helm template` of the umbrella chart (functional-ci values + local overlay values.yaml.template) — renders cleanly, `identity_url`/`idp.source_type`/`idp.external_id_claim` verified in output
- [x] `docker compose config -q` — valid
- [ ] Python e2e contract tests (`test_internal.py`, `test_persons_seed.py`) — not run locally (needs the identity-resolution + MariaDB + ClickHouse compose stack)
- [ ] Real cluster `helm install`/ArgoCD sync — only `helm template` was run, no live cluster available

## Rollout notes (not code changes, just flagging for reviewers)

- This is a breaking chart upgrade: without `identityResolution.deploy: true` + `authenticator.oidc.sourceType` set, the render/gitops-compose fails closed with a clear error; an authenticator upgraded to the new image but still on an old secret (missing `idp.source_type`) will crash-loop at boot validation.
- Even with correct config, logins 403 until `persons` has `value_type='id'` rows under the right `source_type` (for Entra: a fresh persons-seed run after the `ms-entra` connector sync). Rollout order: seed → deploy.

## Summary by CodeRabbit

* **New Features**
  * Added source-scoped external identity resolution for OIDC login bootstrap.
  * Added configurable OIDC source type and external ID claim, defaulting to `sub`.
  * Added identity seeding for roster users with safe, repeatable behavior.
  * Separated login identity lookup from administrator email-based overrides.

* **Bug Fixes**
  * Improved authenticator identity resolution across supported identity providers.
  * Added validation to prevent deployments with incomplete identity-resolution configuration.

* **Documentation**
  * Updated Helm deployment guidance, configuration examples, and identity-resolution endpoint documentation.


charts/insight/templates/secrets.yaml
```diff
@@ -192,7 +192,14 @@ metadata:
 type: Opaque
 stringData:
   APP__gears__authenticator__config__redis_url:     {{ $redisUrl | quote }}
-  APP__gears__authenticator__config__identity_url:  {{ printf "http://%s:8082" (include "insight.identity.host" .) | quote }}
+  # The login-bootstrap resolve (GET /internal/persons/by-external-id /
```
cyberantonz, line 196: The comments about .NET are useless. After a while there will be no .NET service but comments remain.

src/backend/services/identity-resolution/src/api/handlers.rs
```diff
@@ -139,42 +139,128 @@ async fn visible_person_ids(
...
+/// to if it still existed — but it doesn't: this route is override-only by
+/// contract, never called from the login path.
+pub async fn internal_person_by_email_override(
+    Extension(state): Extension<Arc<AppState>>,
+    Extension(ctx): Extension<SecurityContext>,
+    Query(query): Query<InternalByEmailOverrideQuery>,
+) -> Result<impl IntoResponse, CanonicalError> {
+    require_service(&ctx)?;
+
+    let email = query.email.trim();
```
cyberantonz, line 242: Maybe also need to convert to lowercase?

charts/insight/templates/secrets.yaml
mozhaev-dev: Fair point — trimmed this and the similar comments in `_helpers.tpl`, `compose-app-secrets.sh`, `docker-compose.yml`, and `config.rs` to state the current requirement + issue link instead of narrating the .NET rationale at length. Should age better.

src/backend/services/identity-resolution/src/api/handlers.rs
mozhaev-dev: No lowercasing needed here — `persons.value_id` collation was switched to `utf8mb4_unicode_ci` in ADR-0011 (superseding ADR-0004's app-level lowercase-on-lookup approach), so the SQL comparison in `resolve_person_id_by_email_any_tenant` is already case-insensitive at the storage layer. The pre-existing `resolve_person_ids_by_email` (used by the public `/v1/profiles` path) follows the same convention — trims but doesn't lowercase — so this stays consistent with it.

deploy/HELM_DEPLOY.md
```diff
@@ -245,17 +245,33 @@ authenticator:
...
 
+identityResolution:
+  deploy: true                       # REQUIRED — the authenticator's login-bootstrap resolve
+                                      # only exists here (constructorfabric/insight#1960)
+  replicaCount: 1
+  databaseName: "identity"           # SAME database as `identity` above during the cutover
+  resources:
+    requests: { cpu: 100m, memory: 128Mi }
+    limits:   { cpu: 500m, memory: 512Mi }
+
```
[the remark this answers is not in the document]
mozhaev-dev: Valid — fixed. Updated the Step 5 verification pod/Secret list (now expects -identity-resolution, .NET -identity only conditionally), the pre-install checklist (identityResolution.deploy is the required one, identity.deploy stays optional), and the placeholder-reference notes to match.

src/backend/services/identity-resolution/src/infra/db/persons_repo.rs
```diff
@@ -114,6 +117,63 @@ pub async fn resolve_person_id_by_email_any_tenant(
...
+    );
+
+    match db.query_one(stmt).await? {
+        Some(row) => {
+            let bytes: Vec<u8> = row.try_get("", "person_id")?;
+            Ok(Some(Uuid::from_slice(&bytes)?))
+        }
+        None => Ok(None),
+    }
+}
```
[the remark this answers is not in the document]
mozhaev-dev: Good catch on the general principle, but not actionable as a code change here: (source_type, external_id) collision across tenants would need TWO DIFFERENT real external IdPs sharing the same source_type label within one deployment. That can't happen today — `idp.issuer_url`/`idp.source_type` are single values per authenticator deployment, so every tenant behind one authenticator logs in through the SAME real IdP, and external_id is only as unique as that one IdP's own id space (Entra oid / Keycloak's internal user id — both effectively globally unique within their own directory). This is the exact same invariant the pre-existing email-based any-tenant lookup already relied on, not a new risk this PR introduces.\n\nAdded a comment on resolve_person_id_by_source_any_tenant documenting the invariant explicitly and flagging it as something to revisit when multi-IdP config lands (tracked as a #1960 follow-up) — that'\''s the point where two distinct real IdPs could plausibly share a label and this would need an explicit uniqueness check.

charts/insight/templates/secrets.yaml
```diff
@@ -214,6 +219,13 @@ stringData:
   # on the per-user claim (fail closed when neither is present).
   APP__gears__authenticator__config__idp__tenant_claim:      {{ .Values.authenticator.oidc.tenantClaim | default "tenant_id" | quote }}
   APP__gears__authenticator__config__idp__default_tenant_id: {{ .Values.authenticator.oidc.defaultTenantId | default (default "" (default dict .Values.global).tenantDefaultId) | quote }}
+  # The identity-resolution source_type this IdP is seeded under (e.g.
+  # "ms-entra") — required; drives the login-bootstrap person lookup
+  # (GET /internal/persons/by-external-id?source_type=...&external_id=...).
+  APP__gears__authenticator__config__idp__source_type: {{ required "authenticator.oidc.sourceType is required" .Values.authenticator.oidc.sourceType | quote }}
```
cyberantonz, line 225: Do not forget to put proper values to the gitlab gitops

cyberantonz (approved):

mozhaev-dev:
Rebased onto current main (the branch was 144 commits behind) and force-pushed. CI is green; the branch now needs review only.

**History was rewritten** — the ten commits (six real ones plus three stale "merge main" commits) are collapsed into a single signed commit. Every commit on the branch was missing `Signed-off-by`, so DCO failed on all of them, and that is only fixable by rewriting. The previous tip was `68124d0e` if anyone wants the old sequence.

**Three things main changed under this branch**, all of them fixed here:

1. `resolve_person_ids_by_emails` no longer exists on main — the metrics `person_id` cutover replaced the email-keyed visible-persons path. It is dropped; only the `__override` doc note on `resolve_person_id_by_email_any_tenant` survives. This was the one textual conflict.
2. main added a `return_to_prefix` config test that builds its fixture from `valid_config()`, while this branch made `validate()` require `idp.source_type`. The helper now sets it — otherwise the fixture builds a config `validate()` refuses. Git merged both sides cleanly; only the test run caught it.
3. Two e2e helpers still resolved a seeded person through the removed `/internal/persons/by-email/{email}` and answered 404. **This is what the red `identity-rust` job was failing on before the rebase** — it was not a flake. They now use `/internal/persons/by-email-override`.

**Verified on the rebased tree**, which closes the two unchecked boxes in the test plan above:

- `cargo test -p authenticator` 61/61, `-p identity-resolution` 88/88 against a live MariaDB; clippy `--all-targets --all-features` and `cargo fmt --all --check` clean.
- `./e2e.sh test identity/` — **149 passed, 3 skipped**, with the runner image rebuilt from this branch so the new internal routes are actually the ones under test.
- helm render-contract 14/14, identity chart contract 17/17, `deploy/seed/test_identity.py` 2/2, `docker compose config` valid.

The rollout notes in the description still stand unchanged: this is a breaking chart upgrade, and logins 403 until `persons` carries `value_type='id'` rows under the configured `source_type`, so the seed has to run before the deploy.

ktursunov (approved):

cyberantonz (approved):

Issue #1964, opened 2026-07-28
[pres] Create presentation DB + wire analytics to presentation_ro


cyberantonz:
Phase A. Empty `presentation` database; point analytics at `presentation_ro`. Parent: #1803.

Issue #1962, opened 2026-07-28
[pres] Single-SELECT gate on the query path


cyberantonz:
Phase A. Reject anything but one `SELECT`/`WITH` on the public query path (write + run). Parent: #1803.

Issue #1803, opened 2026-07-17
Split Insight into an engineering (data contract) layer and a presentation layer


dzarlax:
## Summary

Split Insight into two clearly separated parts:

- **Engineering layer** — collection, unification, and identity resolution. Produces
  a **contract**: a set of tables that behave as an API. Optimized for correctness,
  stability, reliability.
- **Presentation layer** — a BI-like layer on top of the contract: metric math,
  methodology, visualization. Optimized for fast, safe iteration.

The two halves have different engineering profiles and are developed, tested, and
scaled independently.

## Scope: the division, and only the division

Delivered as quick wins, not a big migration. The boundary itself — read-only role +
single-`SELECT` gate, a writable `presentation` namespace, saved-query CRUD + console,
a tenant filter, contract versioning, manual preview environments — is **shipped**; all
15 sub-issues are closed (see Delivered).

Two ambitions that grew out of this epic now have their own homes and are **not tracked
here**:

- **Semantic layer → #2213.** Definitions-as-data with one compiler over datasets
  (ADR-001, PR #2207): the declarative registry (#1974), passports (#1975), the
  definition core (#2208), the compiler (#1976), the no-SQL editor (#2219), and the
  query→dashboard-card ladder (#1978). The division needs nothing from it — experiments
  run against a stand's real data through the saved-query API and the existing metric
  endpoints. #1979 (physically relocating legacy gold) moved there too.
- **Self-service preview environments → #1981.** Manually `helm`-provisioned previews
  ship here; CI/UI-driven provisioning is a separate epic (#2372–#2375, #2402).

What is left under this epic's own banner is the **hardening the division still owes** —
see Remaining.

## Contract boundary

- The contract is **silver (normalized facts) — including `fct_*` / `mtr_*` and the
  `class_*` tables — plus identity artifacts** (`person.*`, `identity.*`).
- Identity resolution stays in the engineering layer; only its **artifacts** flow to
  presentation, which does not recompute them.
- Changes are **additive only**; the contract is versioned so presentation builds on
  stable ground (old views keep working on new tables). The surface is named (#1968) and
  its version stamped and pinned in analytics (#1969).

## Gold: relabel, do not migrate

- **Legacy gold stays in the `insight` DB**, read by presentation through the read-only
  role — treated as read-shared output, never rewritten.
- A new `presentation` DB holds **new** artifacts only (saved-query results, scratch,
  new gold). Default is a plain `VIEW`; promote to refreshable MV / scheduled
[...]

PR #2019, opened 2026-07-29
feat(analytics): presentation_ro read-only ClickHouse role (#1963)

### LINKED ISSUE #1963: [pres] presentation_ro ClickHouse role
Phase A. `SELECT` on the contract; `CREATE/INSERT/ALTER/DROP` only in `presentation`. Parent: #1803.

cyberantonz:
## What

Adds the `presentation_ro` ClickHouse role — the second barrier behind the single-SELECT gate (#1962) in the presentation-layer split. Contract reads run under a role that makes writing the source impossible **by construction**, not by convention.

Grant matrix (matches the governed PRD/DESIGN — `SELECT` on the contract, `CREATE`/`INSERT` only in `presentation`, **no `DROP`/`ALTER`/`TRUNCATE` anywhere**):

| Scope | Grants |
|---|---|
| `silver.*`, `person.*`, `identity.*`, `insight.*` (contract, incl. legacy gold) | `SELECT` |
| `presentation.*` (writable namespace) | `SELECT`, `INSERT`, `CREATE` |

## Changes

- **`bootstrap-db/presentation-role.sql`** — canonical, idempotent role DDL.
- **`apply-ch-migrations.sh`** — provisions the role, **guarded + non-fatal**: creating a role needs `access_management` on the applying admin, so an admin without it is skipped with a warning instead of aborting the deploy. Bootstrap picks this up via its existing step-4 call.
- **`clickhouse-access-management.xml`** + README — enable `access_management` on the local/throwaway (official-image) ClickHouse so bootstrap provisions the role; the official image ships it disabled.
- **`tests/test_presentation_role.py`** — opt-in integration test (repo's skip-unless-`*_CH_URL` pattern) asserting the full allow/deny grant matrix against a real ClickHouse. This is the adversarial-write half of NFR `cpt-presentation-nfr-source-immutability`.
- **PRD/DESIGN** — mark `cpt-presentation-fr-read-only-role` implemented; add the `cpt-presentation-component-read-only-role` component. `cfs` per-artifact validate + `check-language` pass clean.

## Verification

- Integration test passes against a throwaway CH `25.7.5`; the exact allow/deny matrix (contract read-only; `presentation` create/insert-only; DROP/ALTER/TRUNCATE denied everywhere) was confirmed with a probe user carrying only the role.
- The guarded provisioning step was verified to warn-and-continue (no `set -e` abort) on a CH whose admin lacks `access_management`.

## Scope / follow-ups

- Wiring the analytics service to *connect* as `presentation_ro` (needs the runtime user + sealed secret) and creating the empty `presentation` database are the next slices (#1964 + connection wiring), per the DESIGN implementation plan.
- ⚠️ **Prod enforcement depends on the bitnami CH admin having `access_management`.** If it does not, the guard silently skips the role (WARN). This is harmless today (no consumer yet), but the analytics-wiring PR must confirm/enable it before relying on prod enforcement.

Closes #1963
Part of #1803

## Summary by CodeRabbit

* **New Features**
  * Added automatic provisioning of a `presentation_ro` ClickHouse role.
  * Grants read-only access to contract data and controlled create, insert, and read access within the presentation area.
  * Provisioning is safe to repeat and does not block deployments when permissions are unavailable.

* **Documentation**
  * Updated design and product documentation to describe the role’s permissions, setup, and deployment behavior.
  * Added local testing guidance for enabling access management.

* **Tests**
  * Added integration coverage verifying allowed operations and blocked destructive changes.


cyberantonz:
Addressed the review in `c4d504c4`:

**Semgrep (`dynamic-urllib`, test)** — the test now pins `CH_URL` to an `http(s)` scheme before `urlopen` (rejecting `file://` etc.) and suppresses the audited rule with justification. The value is an operator-supplied test endpoint, but the scheme guard makes that explicit and safe.

**CodeRabbit — don't mark enforcement shipped** — correct: the role is *provisioned* by #1963 but not yet the active query-path identity (analytics still connects as the admin until the connection wiring lands). Left `cpt-presentation-fr-read-only-role` and the DESIGN component **unchecked** and stated the role is provisioned-but-dormant.

**CodeRabbit — document `person.*`** — `person` is contract (identity domain), so I kept the grant and added `person` to the PRD's listed contract databases.

**CodeRabbit — cover every contract namespace** — the read-only test is now parameterized over `silver`/`person`/`identity`/`insight` (each gets a probe table; reads allowed, writes/DDL denied).

**CodeRabbit — grant before default role** — fixed: `CREATE USER` → `GRANT presentation_ro` → `ALTER USER … DEFAULT ROLE`.

Also, per maintainer feedback, reverted the earlier `deploy/gitops/system/clickhouse/values.yaml` comment so gitops config stays untouched — the role is provisioned by the existing `clickhouse-migrate` hook and the bitnami admin already has access-management.

The connectors-ddl regen reminder is a false positive here: this PR adds a role + provisioning, no bronze/silver/gold schema change.

cyberantonz:
The failing `secrets (diff)` check is a **false positive not introduced by this PR** — TruffleHog's `Lob` detector matching the test function name `test_divergent_watermark_fetches_its_own` in `src/ingestion/connectors/git/bitbucket-cloud/tests/test_request_budget.py:199` (a file this PR doesn't touch; it also fails #2023). Tracked in #2025. This PR's own diff scans clean.

dzarlax (approved):

PR #2045, opened 2026-07-30
fix(hr): make class_people a true one-row-per-person snapshot

### LINKED ISSUE #2037: Employees appear twice in the people registry after their HR record changes
## Summary

Updating an existing employee in an HR source can create an additional current person record, inflating people counts and any listing that counts those rows.

## Steps to Reproduce

1. Sync a BambooHR or Workday source and run its dbt models.
2. Change an attribute of an existing employee in the HR source, such as their job title.
3. Sync the same source again and rerun its dbt models.
4. Check the current Silver rows for each source employee:

   ```sql
   SELECT
       workspace_id,
       source,
       source_person_id,
       count() AS row_count,
       groupArray(toString(valid_from)) AS observed_versions
   FROM silver.class_people FINAL
   GROUP BY workspace_id, source, source_person_id
   HAVING count() > 1;
   ```

**Expected:** The updated employee has one current row containing the latest source values.

**Actual:** The updated employee appears in multiple current rows with the same `source_person_id`, so row-based people counts increase.

## Additional information

- The repeated rows contain complete employee records; they are not empty or partially populated placeholder rows.
- The source identifies the records as the same employee.
- The repeated Silver rows have different `valid_from` values.
- Employees whose source record has not changed remain represented by one row.

## Notes

Related to #749 and #2038.

mitasovr:
## Problem

A data-quality audit on virtuozzo reported two findings against `silver.class_people`:

- *"19 duplicate active emails — the SCD2 dedup is broken"*
- *"active-count drift: silver has 20 more active people than bronze"*

They are **one root cause**, and the dedup is not broken — ReplacingMergeTree is behaving correctly. Live numbers: **1440 rows for 1420 employees**, **414 `active` rows for 394 active people** (~5% headcount inflation), `countIf(valid_to IS NOT NULL) = 0`.

## Root cause: two defects compounding

Neither is sufficient alone.

1. **The `__to_class_people` staging views read append-only bronze without `FINAL`.** Airbyte appends a *full snapshot per sync*, and bronze is `ReplacingMergeTree(_airbyte_extracted_at)`, which only collapses on background merge. A bare read therefore emits every unmerged snapshot row. (Caught live: bronze went from 1420 rows / 1 part to 2840 rows / 2 parts mid-investigation.)
2. **Those views appended a version axis to `unique_key`** (`lastChanged` / `createdDateTime` / `whenCreated` / `Last_Functionally_Updated`), as ADR-0004 required. That turns each *transient* bronze duplicate into a *permanently* distinct silver key — so the versionless RMT can never collapse it, and **`FINAL` cannot either**, because the keys genuinely differ. `SELECT count() FROM silver.class_people FINAL` returns the same 1440.

`valid_to` was hardcoded `NULL` in every producer, so nothing marked which row was current.

Because the table is `materialized='table'`, the surplus is **regenerated on every run** and drifts with bronze merge timing — which is exactly why it surfaced as two separate symptoms. The nightly rebuilt it mid-investigation and the surplus moved 19 → 20.

### The arithmetic, end to end

```
1440 rows = 1400 entities x 1 row + 20 entities x 2 rows
1420 distinct (source, source_person_id)      <- exactly bronze's 1420 employees
 414 active rows = 374 singles + 20 doubles   -> 394 distinct active entities
```

The residual vs bronze's 393 `status='Active'` is **+1**, and it is *not* a status disagreement between versions. It is the `ELSE 'active'` catch-all: exactly one bronze row has `status=''` with `employmentHistoryStatus='Third party'` and falls through to `active`. 393 + 1 = 394. ✔

## Why the contract allowed this

Two normative documents disagreed:

| Source | Claim |
|---|---|
| `ADR-0004:58` | version axis in `unique_key` is **mandatory** for `class_people` |
| `DESIGN.md:178` | `class_people` is a snapshot **"with no event history kept at this layer"** |
| `hr-directory/README.md:168` | full SCD2, *"exactly one row per person may have `valid_to IS NULL`"*, enforced by an *"SCD2 Merge component"* |
| `ADR-0001:109` | exempted `class_people` from the read-dedup cleanup because it *"already collapse[s] to one row per key"* |

`ADR-0001:109` is the load-bearing error: "one row per key" only equals "one row per person" while `unique_key` *is* the entity key — which ADR-0004 had already stopped being true.

**History is genuinely covered elsewhere**, which is what makes the snapshot reading safe: each connector already has `*_employees_snapshot` / `*_fields_history` (`incremental`+`append`, 17 tracked columns) that accumulate across syncs. `class_people` is `materialized='table'` and cannot retain history by construction. Note `class_hr_events` is *not* that history — it reads `leave_requests` only.

## Changes

- **All four `__to_class_people` views** (active-directory, bamboohr, ms-entra, workday): entity-level `unique_key`, `valid_to` dropped, bronze read with `FINAL`. All four projections verified aligned 1:1 (25 columns) for the positional `UNION ALL`.
- **`ELSE 'active'` → `'unknown'`** in all four, with `accepted_values` updated. Defaulting an unrecognised source status to `active` silently inflates headcount.
- **Same missing `FINAL` in `bamboohr__working_hours` / `workday__working_hours`.** No row inflation there (`unique_key` was already entity-level), but `WHERE status='Active'` ran *before* dedup, so a leaver could still qualify via a stale snapshot row.
- **New `assert_class_people_one_row_per_person`** data-quality check (`data_quality`, `tier: error`), plus `unique`/`not_null` on `class_people.unique_key`. There was previously **no** uniqueness test on this table and **zero** `data_quality` tests touching it.
- **Contract reconciled**: ADR-0004, ADR-0001, `union_by_tag` docstring, HR README, four connector `schema.yml`, glossary.

## Verification (live virtuozzo, read-only)

Simulating the fixed transform against live bronze:

| | before | after |
|---|---|---|
| rows / distinct keys | 1440 / 1420 | **1420 / 1420** |
| `active` | 414 rows (394 entities) | **393** |
| `unknown` | — | **1** |

The new DQ check returns **exactly the 20 current violations** against the deployed table, and goes green after a rebuild.

Also demonstrated *why* `FINAL` is mandatory rather than defensive: the versionless `LIMIT 1 BY unique_key` in `union_by_tag` has no `ORDER BY`, so the surviving row is undefined. At `max_threads=1` it picks the **stale** row for **20 of 20** changed entities; at `max_threads=4/16`, the fresh one.

## Consumers

- `gold/metric_entity_cohorts_current.sql` — already correct (`LIMIT 1 BY` ordered by `valid_from DESC`); keeps working, now over a unique input.
- `crm-gold-views.sql` (4x `silver.class_people FINAL`) — `FINAL` was a false safety signal; the `people` CTE is `LEFT JOIN`ed on `lower(email)`, so duplicates fan out and double `count()`/`sum()`. Currently latent (verified 0 overlap between CRM owner emails and the 20 duplicated emails); fixed at the source by this PR.
- `insight.people` — reads bronze directly with `argMax`; unaffected.
- No reference to `class_people` in `services/analytics`. **`valid_to` had zero readers anywhere in `src/`.**

## Migration

**None needed.** `class_people` is `materialized='table'`, so one `dbt run` replaces it wholesale and the surplus rows disappear — unlike the column-order heal precedent in `20260716000000_class_contract_heal.sql`, which exists for *incremental* tables with positional inserts.

```
argo submit --from workflowtemplate/dbt-run -n insight-v2 -p dbt_select="tag:bamboohr+" -l workflows.argoproj.io/controller-instanceid=argo-v2-insight-v2
```

Expect `count() = uniqExact(source, source_person_id) = 1420` and `countIf(status='active') = 393`.

## Deliberately not in this PR

- **`audit_rmt_read_dedup.py` has a false negative that hid this bug.** It matches `source('<name>','<table>')` against the physical `<schema>.<table>` recorded by `promote_bronze_to_rmt`, so connectors declaring an unprefixed source name (`bamboohr`, `workday` — both `class_people` producers) are silently classified non-RMT and skipped; `ms-entra` is caught because its source *is* named `bronze_ms_entra`. The one-line fix surfaces 5 further genuine gaps. Left out because the file predates the repo's `ruff` config and touching it trips 31 pre-existing lint errors plus a whole-file reformat, which would bury this change. **Follow-up needed.**
- `silver/hr/schema.yml:7` references `dbt_utils.unique_combination_of_columns`, but `dbt_utils` is not installed anywhere (no `packages.yml`, no `dbt deps`, not vendored) — a pre-existing latent breakage. Follow-up needed.
- `hr_events` models also read bronze without `FINAL`, but `union_by_tag`'s **versioned** dedup path (`QUALIFY ROW_NUMBER ... ORDER BY _version DESC`) protects them, unlike the versionless path used here.

### DIFF
--- a/docs/components/connectors/hr-directory/README.md
+++ b/docs/components/connectors/hr-directory/README.md
@@ -64,7 +64,19 @@ prd: pending
 
 ### 1.1 Architectural Vision
 
-The HR Silver Layer transforms raw HR directory data from Bronze source tables (BambooHR, Workday, LDAP/Active Directory) into two canonical, workspace-isolated Silver tables: `class_people` and `class_org_units`. Both tables follow the SCD Type 2 pattern: each row captures a point-in-time snapshot of a person or organisational unit, with `valid_from`/`valid_to` interval semantics. The current state of any entity is always the row where `valid_to IS NULL`.
+> **Implementation status.** This document is the original HR Silver design. Parts of it describe components that were never built, and they must not be read as the current contract:
+>
+> - There is **no "SCD2 Merge" component**. Bronze → Silver is plain dbt (`<source>__to_class_people` staging views unioned by `union_by_tag` into `silver.class_people`). Every section describing row-closing writes (`UPDATE ... SET valid_to = ...`) is unimplemented design.
+> - **`class_people` has no `valid_to`** and holds no version history — see below.
+> - **`class_org_units` does not exist** as a table or a dbt model. `class_people.org_unit_id` is consequently unpopulated; department attribution currently flows through `department_name`.
+>
+> For the shipped data-flow contract see [ADR-0001](../../../domain/ingestion-data-flow/specs/ADR/0001-rmt-with-version-and-unique-key.md) and [ADR-0004](../../../domain/ingestion-data-flow/specs/ADR/0004-unique-key-formula.md).
+
+The HR Silver Layer transforms raw HR directory data from Bronze source tables (BambooHR, MS Entra, Workday, LDAP/Active Directory) into two canonical, workspace-isolated Silver tables: `class_people` and `class_org_units`.
+
+`class_people` is a **current-state snapshot**: exactly one row per person per source, keyed on an entity-level `unique_key`. `valid_from` records when the source last changed the record; there is no `valid_to`. It is `materialized='table'` and rebuilt in full on every run, so it cannot accumulate history by construction.
+
+HR attribute history lives in the per-source SCD2 chain — `<source>__<entity>_snapshot` and `<source>__<entity>_fields_history` — which are `incremental`/`append` and therefore genuinely accumulate versions across syncs, tracking strictly more fields than `class_people` exposes. Do not add version rows to `class_people`: see ADR-0004 and ADR-0001 for the headcount-inflation bug this caused.
 
 The design makes the HR domain the authoritative source for person identity and organisational structure within the Insight platform. `class_people` serves as the canonical person registry consumed by all other Silver streams that carry `person_id`. `class_org_units` provides the org hierarchy needed for hierarchical scoping of metrics — the depth, path, and head of each organisational unit — data that cannot be reliably derived from `class_people` alone.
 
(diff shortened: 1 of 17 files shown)


cyberantonz (approved):

aleksdotbar (approved):

mitasovr:
Superseded by #2125 — same change, rebased onto current `main` (178 commits) and moved from the `mitasovr` fork into this repository so CI runs with repository secrets. No content change beyond the rebase; the `active_directory__to_class_people.sql` conflict was resolved in favour of main's `manager_person_id` / `hire_date` / `termination_date` retyping.

PR #2046, opened 2026-07-30
feat(identity): make persons-seed CLI-only with a scheduled CronJob (#1690)

### LINKED ISSUE #1690: Team view shows a stale roster — manager and org changes never appear
## Summary

A manager's Team view lists the wrong people: anyone whose manager changed since the last org-tree build is still shown under their old manager and missing from the new one.

## Steps to Reproduce

1. Find a person whose ingested manager has changed at least once:

```sql
-- ClickHouse
SELECT count() FROM (
  SELECT source_account_id
  FROM identity.identity_inputs
  WHERE insight_source_type = 'bamboohr' AND value_type = 'parent_email'
  GROUP BY source_account_id
  HAVING uniqExact(value) > 1);     -- 527
```

2. Open that person's **new** manager's Team view (`/ic/<new-manager-email>/team`) → they are absent from Direct reports.
3. Open the **old** manager's Team view → they are still listed there.
4. Call `GET /api/identity/v1/persons/<old-manager-email>` → the `subordinates` tree returns the same stale membership the UI shows.

**Expected:** Team view rosters follow the latest ingested manager/org data.
**Actual:** the roster still reflects the manager the person had at the last org-tree build.

Matched comparison, same person, one variable changed — the two stores disagree:

```sql
-- ClickHouse gold → returns the NEW manager
SELECT supervisor_email FROM insight.people WHERE person_id = '<person-email>';
```

In the identity database, that person's open `org_chart` edge (`valid_to IS NULL`) still points at the **old** manager, and the new manager's children do not include them.

## Additional information

- The org tree's newest write is weeks old while its input is same-day fresh:

```sql
-- ClickHouse: newest manager observation in the identity inputs
SELECT max(_synced_at) FROM identity.identity_inputs
WHERE insight_source_type = 'bamboohr' AND value_type = 'parent_email';   -- today
```
```sql
-- identity database: newest edge ever written to the org tree
SELECT MAX(valid_from), MAX(valid_to) FROM org_chart;                     -- 2026-05-22
```

- Every edge in `org_chart` shares that single most-recent `valid_from`; nothing has been written to the tree since.
- 527 people have ≥2 distinct `parent_email` values in `identity_inputs`. For one verified person the newer value synced 2026-07-01 and carries a higher `_version`, yet their open `org_chart` edge is unchanged.
- Reproduced again 2026-07-29 on a separate deployed instance with the same result. State on that instance: `bronze_bamboohr.employees` ~2.2k rows extracted same-day; `identity.identity_inputs` ~36.8k rows synced same-day; `insight.people` ~1,000 people across 51 org units with 0 unresolvable `supervisor_email`; the persons-seed output in the identity database was empty.
- `identity.bootstrap_inputs` → `staging.bamboohr__bootstrap_inputs` on that instance is a view with no counterpart in the dbt project. Noted only so it is not chased separately.

mozhaev-dev:
Closes #1690.

## Problem

The identity org projection (`persons` / `account_person_map` / `org_chart` in MariaDB) is **materialized** and was only rebuilt by the persons-seed at bootstrap. Nothing re-ran it, while connectors keep its input (`identity.identity_inputs` in ClickHouse) fresh daily — so every manager/org change ingested after the last seed run never reached the Team view (527 re-parented people invisible on the reporting Virtuozzo instance).

## How it works now

The seed becomes a first-class **scheduled operation of the identity-resolution service**; the HTTP trigger is removed (team decision — the .NET service and its `POST /v1/persons-seed` are being deleted together).

```
CronJob (30 6 * * * UTC, or a manual Job)
  └─ identity-resolution seed [--mode link-by-email] [--force]
       1. GET_LOCK('persons-seed:<tenant>', 0)   ← per-tenant MariaDB advisory lock,
          held on a dedicated single connection for the whole run
       2. zombie sweep                            ← reclaims operations rows a killed
          Job left `queued`/`running`
       3. operations journal row                  ← author = nil UUID (SYSTEM_AUTHOR),
          request records {"trigger": "cli"}
       4. read identity_inputs → input guards     ← see below
       5. the same domain pipeline the endpoint used
          (resolve → INSERT IGNORE persons → rebuild account_person_map + org_chart,
          one transaction), bounded by a 10-minute timeout
       6. journal → completed (summary) / failed
```

**Exit codes** (diagnosable straight from the Job status): `0` ok · `1` failed · `2` another run holds the lock · `3` refused by an input guard.

**Concurrency.** Runs serialize on the advisory lock — it lives on the MariaDB server, so a cron Job, a manual Job, and even a second Insight instance sharing the same database all serialize through it, crash-safe by construction (the lock dies with the connection). `concurrencyPolicy: Forbid` on the CronJob is belt-and-braces for cron-vs-cron only. A concurrent run fails fast (exit 2) rather than queueing a stale re-run.

**Input guards** (both `--force`-overridable, both journaled as a `failed` operation so `GET /v1/persons-seed` explains why nothing was written):
- *empty input* — 0 `identity_inputs` rows means a broken/misconfigured pipeline (wrong ClickHouse URL/database, wiped stand), not "no people";
- *wrong tenant* — `persons` holds rows under other tenants and none under the configured one: seeding would mint a parallel person universe under a wrong tenant (the #1550 failure mode), which the append-only log cannot undo. A genuinely fresh install (empty `persons`) passes.

**What stays / what goes.** `POST /v1/persons-seed`, the in-process queue/worker and the 503 queue-full path are gone. The read-only journal routes `GET /v1/persons-seed` and `GET /v1/persons-seed/{id}` stay as the observability window over CLI runs. The subcommand paths now install a tracing subscriber — previously `migrate` logged into the void.

**Helm.** The CronJob lives in the identity-resolution subchart (same image/config/secret wiring as the deployment; distinct pod labels so the Service never routes traffic to seed pods). Values: `identityResolution.seed.{enabled, schedule}` (+ subchart-only `seed.tenantDefaultId` that env-overrides the Secret for standalone installs). The umbrella fails the render when the seed is enabled but no tenant is configured — only on the path where the umbrella composes the config Secret itself (`credentials.autoGenerate`); pre-created-secret installs are the operator's source of truth.

## Running it manually

Inside the cluster, from the CronJob (same image/env/config — nothing to remember):

```bash
kubectl -n <ns> create job --from=cronjob/<release>-identity-resolution-seed seed-manual-$(whoami)
kubectl -n <ns> logs -f job/seed-manual-$(whoami)
kubectl -n <ns> get job seed-manual-$(whoami)   # exit code semantics above
```

Watching a run / history over HTTP (admin JWT):

```bash
GET /v1/persons-seed            # list runs, ?status=&limit=
GET /v1/persons-seed/{id}       # one run: status, summary, error_message
```

Overriding the guards knowingly (e.g. a legitimate first seed while another tenant's data exists):

```bash
kubectl -n <ns> create job seed-force --image=<identity-resolution image> \
  -- /app/identity-resolution -c /app/config/insight.yaml seed --force
```

(or simplest: `kubectl create job --from=cronjob/... seed-force`, then edit the Job's args to append `--force` before creating — the CronJob itself never runs forced.)

## Tests

- **Unit** (67): guard decision table (empty input, wrong tenant, `--force`, fresh-install and steady-state pass), mode/tenant validation, pipeline fakes.
- **Contract e2e** (`src/ingestion/tests/e2e/identity/`, both CI lanes):
  - CLI trigger cases (rust lane): completed run journaled with nil author + `trigger: "cli"`, empty-input guard (exit 3, journaled), wrong-tenant guard (exit 3, journaled), lock busy (exit 2 against a held `GET_LOCK`);
  - **inputs → org_chart correspondence**: after a seed the open edges match `identity_inputs` (resolvable manager → edge, unresolvable → NULL-parent membership, top-of-tree → Path-B row);
  - **the #1690 regression proper**: a newer `parent_email` lands in the inputs → a re-run moves the open edge to the new manager and closes the old one (SCD2 history intact);
  - POST cases are gated to the dotnet lane and die with the .NET service; the rust coverage gate carries an approved SKIP for the removed POST.

## Verification

- `cargo fmt` / `clippy -D warnings` / 67 unit tests green; release binary smoke-tested (`seed --help`, exit codes).
- Identity contract suite: rust lane 113 passed (three back-to-back runs — flake check), dotnet lane 108 passed; both endpoint-coverage gates rc=0.
- `helm lint` both charts; render matrix: umbrella fails fast on enabled-seed-without-tenant, renders with `global.tenantDefaultId` / `seed.tenantDefaultId` / `seed.enabled=false`; functional-ci values render unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added a scheduled **persons-seed** CronJob and seed job configuration (schedule, resources, history limits) with **force** support.
  * Added a Rust **CLI seed** workflow that records progress via the existing operations journal (write trigger removed; status/journal reads remain).
* **Bug Fixes**
  * Helm/seed rendering now **fails fast** when seed is enabled without a tenant, including umbrella-chart guard behavior.
  * Improved tenant handling by supporting safe tenant inference when unconfigured (and refusing ambiguous/empty cases).
* **Tests**
  * Expanded Helm contract and end-to-end suites to validate rendering, labels, locking/guard behavior, and journaling semantics.


ktursunov (approved):

.github/workflows/identity-resolution-helm.yml
```diff
@@ -0,0 +1,50 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      # Pinned to full SHAs (mutable tags are repointable — supply-chain
+      # hardening, same as semgrep.yml / trivy.yml).
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
+
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
+        with:
+          python-version: "3.12"
```
cyberantonz, line 38: Why so old?

mozhaev-dev: Fair catch — the action SHAs were copied from semgrep.yml and lagged behind; bumped to the current releases in ddf94908 (checkout v7.0.1, setup-python v7.0.0). `python-version: 3.12` itself is intentional: it's the repo-wide standard (every workflow, the e2e runner image, and `requires-python` in the e2e pyproject all pin 3.12) — happy to bump repo-wide in a separate PR if we want to move.

src/backend/services/identity-resolution/src/seed_runner.rs
```diff
@@ -0,0 +1,327 @@
...
+    // Work pool + a dedicated single-connection session for the advisory
+    // lock: GET_LOCK is session-scoped, so the lock must not ride a pooled
+    // connection that could be swapped mid-run.
+    let db = db::connect(&config.database_url).await?;
+    let lock_conn = db::connect_single(&config.database_url).await?;
+    if !db::try_acquire_seed_lock(&lock_conn, tenant).await? {
+        return Err(SeedRunError::LockBusy);
+    }
+
+    let result = tokio::time::timeout(RUN_TIMEOUT, run_locked(&db, config, tenant, mode, force))
```
cyberantonz, line 139: If you crash here, you will have lock held forever. I think it should be done via RAII pattern, i.e. lock held until the variable is in the scope. When you leave scope - the lock automatically removed.

src/backend/services/identity-resolution/src/seed_runner.rs
```diff
@@ -0,0 +1,327 @@
...
+    config: &GearConfig,
+    mode: &str,
+    force: bool,
+) -> Result<SeedSummary, SeedRunError> {
+    if mode != LINK_BY_EMAIL_MODE {
+        return Err(SeedRunError::Failed(anyhow::anyhow!(
+            "unsupported mode '{mode}'; only '{LINK_BY_EMAIL_MODE}' is available"
+        )));
+    }
+    if config.tenant_default_id.is_empty() {
```
cyberantonz: It is true for one installation. But at least dev has no default tenant id at all and in perfect world you need to grab all tenants from DB and run for each. Currently you will break multitenancy.

src/backend/services/identity-resolution/src/seed_runner.rs
mozhaev-dev: Done in ff8d3afb — `SeedLockGuard` now owns the lock's session (RAII), every exit path releases it. Worth noting: `GET_LOCK` is session-scoped, so a crash already released it server-side — the guard formalizes that invariant.

src/backend/services/identity-resolution/src/seed_runner.rs
mozhaev-dev: Agreed on direction, but a per-tenant loop is blocked one level lower: the identity_inputs reader reads the whole table with no tenant filter (HOTFIX(#1550) — the dbt producer hashes tenant ids), so each tenant's run would ingest every other tenant's rows. Documented this at the validation site in ff8d3afb (grep `HOTFIX(#1550)` for the full blast radius); the runner below is already per-tenant (lock, journal, writes), so the loop drops in once the producer fix lands.

mozhaev-dev: Follow-up in 6c68bfbd: since existing config Secrets (dev included) can't be touched right now, an EMPTY tenant_default_id no longer hard-fails — the runner infers the SOLE distinct tenant from the persons log (WARN-logged); zero or several tenants still refuse with a clear message, so nothing is ever guessed ambiguously. Dev works out of the box with this.

mitasovr (approved):

Issue #1873, opened 2026-07-23
GRAPH 1b: Identity mapping engine — resolve accounts to a canonical person


dzarlax:
## Goal

Merge each person's accounts across systems (HR, git, AI tools, comms) into
**one canonical person** with a **reported match rate**, so every metric, cohort,
and lineage link resolves to a real person. Foundational for #1455, #1660, #1617,
and all person-grain gold views. ([VISION.md](https://github.com/constructorfabric/insight/blob/main/docs/product/VISION.md) §7.2, Person entity.)

This epic is the **resolution behaviour** — the matching, review, and lifecycle
logic. The service that hosts it is being ported to Rust in its companion epic
(#1602); this epic can be built on either implementation.

## How it works

Auto-match high-confidence cases; **escalate disputed / ambiguous ones to admin
review**; support manual override + aliases. Match rate is reported.

## Scope

- Resolve HR / Entra directory + git author emails + AI-tool emails + comms into
  one `person_id`.
- Auto-match vs an admin-review queue; manual override + aliases.
- Report match rate; **unresolved authors are excluded (honest-NULL), not guessed**.
- Honest reality: git → HR match ≈ 68% in the internal experiment; the **reviewer
  namespace (GitHub login / Bitbucket display-name) ≠ author email** —
  cross-namespace resolution is needed before review-side metrics.

## Product capabilities (per PRD)

- **Operator review queue** — anything not confidently matched goes to an operator
  queue to link / ignore / create; operators review only the unresolved set.
- **Conflict handling** — when one identity is claimed by two people, surface it
  for an operator decision; never silently auto-merge.
- **No false merges (safety)** — auto-linking is conservative; ambiguous cases go
  to review (a wrong merge corrupts analytics).
- **Merge / split (reversible, audited)** — an operator can merge two records for
  the same human and reverse it; every merge/split is auditable.
- **GDPR / right-to-erasure** — on request, purge a person's identity data so it's
  no longer resolvable anywhere.
- **Tenant isolation** — one tenant's identities never leak into another.

## Scope — people model

- The **canonical person**, including a **golden-record view** (the merged
  best-value profile), is the product outcome here.
- **Org hierarchy** (reporting lines / team model) is in scope via #1605.

## Out of scope

- **RBAC / access governance** → #1635 — identity ≠ permissions.
- The service port to Rust → #1602.
- Metric aggregation; connector implementation.

## Success measures

[...]

Issue #719, opened 2026-06-04
[IR-05] GDPR purge implementation — alias erasure endpoint + archive


il10241024:
> 🔗 **Mirrored from** [cyberfabric/cyber-insight#191](https://github.com/cyberfabric/cyber-insight/issues/191)
> Originally opened by **dzarlax** on 2026-04-17T10:13:41Z

---

## Context

The Identity Resolution DESIGN specifies a GDPR-compliant data subject erasure flow with an \`alias_gdpr_deleted\` archive table. The storage side is described but the endpoint and enforcement logic are not implemented.

## Summary

Implement GDPR-compliant erasure: operator/API request deletes all aliases for a person, archives them to \`alias_gdpr_deleted\` with tombstone record, and propagates purge downstream (Silver identity-enriched rows).

## Scope

- \`POST /v1/gdpr-purge\` endpoint (identity-admin RBAC)
- Request payload: person_id, reason, requested_by, requested_at
- Move rows from \`aliases\` → \`alias_gdpr_deleted\` (ReplacingMergeTree with tombstone version)
- Block future auto-matching of purged aliases (add to deny-list in match engine)
- Produce \`insight.identity.resolved\` event with purge type for downstream propagation
- Audit log entry in \`insight.audit.events\` (actor, person_id, reason, timestamp)
- Integration test: full purge lifecycle + verify Silver re-resolution drops person
- Operator documentation / runbook

## References

- Identity Resolution DESIGN: [docs/domain/identity-resolution/specs/DESIGN.md](https://github.com/constructorfabric/insight-spec/blob/main/docs/domain/identity-resolution/specs/DESIGN.md) — GDPR & tables sections

## Dependencies

- IR-04 (API skeleton)
- Audit Service (#716) for audit events
- Redpanda topic bootstrap (#717)

---

PR #2065, opened 2026-07-30
feat(identity): `sync` CLI — copy the persons log into ClickHouse (metrics person_id, iteration 1)

mozhaev-dev:
## What

A new `identity-resolution sync` CLI subcommand (+ Helm CronJob + read-only journal GETs) that copies the MariaDB `persons` observation log into ClickHouse **`identity.identity_persons`** — the table the metrics dbt builds will resolve `email → person_id` against.

Consumer-side foundation for the identity mapping engine epic (#1873): however matching improves there, its output reaches metrics through this pipe; the dbt side (resolve macro + `person_id` in the observation models + activity-weighted match-rate view) follows in a companion PR.

## How

**Trigger model** — CLI-only from birth, mirroring the persons-seed shape (#1690): `identity-resolution sync` runs one copy and exits (Helm `CronJob` scheduled 15 min after the seed's — the seed rewrites the log, the sync publishes it; manual `kubectl create job --from=cronjob/...-sync`). Exit codes `0 ok / 1 failed / 2 lock busy / 3 guard`. HTTP keeps only the admin-gated journal: `GET /v1/persons-sync` + `/{id}`.

**Copy semantics** — full snapshot per run, atomic publish:
- per-run staging table → RowBinary insert → **count-verify** (a short write is never swapped in) → `EXCHANGE TABLES` → drop; readers only ever see a complete snapshot, and a failed run leaves the live one untouched;
- whole run holds a **global MariaDB `GET_LOCK`** (RAII guard, released even on process death); the `_synced_at` watermark guard before the swap is a backstop for anything bypassing the runner; stale stagings from crashed runs are GC'd after 1h;
- the log is copied **verbatim** (source provenance columns, NULL vs `''` preserved) — all resolution semantics stay downstream in the dbt macro, and per-source resolution stays possible;
- **empty-log guard**: publishing a 0-row read would atomically erase a populated snapshot (usually a misconfigured DB, not "no people") — refused with exit 3, `--force` overrides deliberately.

**Journal** — every run (guard refusals included) lands in `operations` under the resolved tenant with the `SYSTEM_AUTHOR` nil UUID; the summary carries `rows` / `max_id` / `max_created_at` / `synced_at` — the resolution watermark for "why did the numbers change between builds".

## Testing

- `cargo test -p identity-resolution` — 81 unit tests (incl. run_sync orchestration on fakes, guard decisions, wire-row mapping, live-gated lock mutual-exclusion);
- clippy pedantic clean, fmt clean;
- e2e (containerized stack, `E2E_IDENTITY_IMPLEMENTATION=rust`): new `identity/test_persons_sync.py` — CLI exit 0 + terminal journal row, ClickHouse snapshot assertions against the fixture dataset (fixed person UUIDs, shared-email pair stays unresolved — the copy is a log, not a resolution), replace-not-append semantics, journal type-scoping vs the seed surface, 401/403/404;
- full identity contract suite: **124 passed, 5 skipped**;
- `helm lint` + template render (2 CronJobs) clean.

Refs #1873, #1690.

## Summary by CodeRabbit

* **New Features**
  * Added scheduled synchronization that publishes identity data to ClickHouse.
  * Added a command-line sync operation with concurrency protection and safeguards against empty snapshots.
  * Added admin-only APIs to list and retrieve synchronization history, including status filtering.
  * Added atomic snapshot replacement so ClickHouse reflects the latest successful run.
* **Bug Fixes**
  * Improved reliability by preventing overlapping runs and protecting newer snapshots from being overwritten.

### DIFF
--- a/src/backend/services/identity-resolution/helm/values.yaml
+++ b/src/backend/services/identity-resolution/helm/values.yaml
@@ -93,6 +93,36 @@ seed:
       cpu: 500m
       memory: 512Mi
 
+# Persons-sync CronJob: publishes the `persons` log into ClickHouse
+# `identity.identity_persons` (the metrics email→person_id resolve source).
+# Scheduled after the seed by default — the seed rewrites the log, the sync
+# publishes it.
+sync:
+  enabled: true
+  # Journal-row tenant (UUID); same override semantics as seed.tenantDefaultId
+  # (the copy itself is tenant-agnostic — this only scopes the operations
+  # journal the admin GET routes read).
+  tenantDefaultId: ""
+  # Daily, 15 minutes after the seed's 06:30 run.
+  schedule: "45 6 * * *"
+  # Belt-and-braces for cron-vs-cron only — manual Jobs and other instances
+  # serialize on the MariaDB advisory lock regardless.
+  concurrencyPolicy: Forbid
+  # Retries for transient connect blips; the lock + snapshot-swap idempotency
+  # make a repeated run safe.
+  backoffLimit: 2
+  # Caps a wedged pod well past the in-binary 5-minute sync timeout.
+  activeDeadlineSeconds: 600
+  successfulJobsHistoryLimit: 3
+  failedJobsHistoryLimit: 3
+  resources:
+    requests:
+      cpu: 100m
+      memory: 128Mi
+    limits:
+      cpu: 500m
+      memory: 512Mi
+
 # TCP gate before start: the service connects to MariaDB at boot. Same
 # busybox wait as the .NET identity chart.
 waitForMariadb:
--- a/src/backend/services/identity-resolution/src/infra/db/ops_repo.rs
+++ b/src/backend/services/identity-resolution/src/infra/db/ops_repo.rs
@@ -18,6 +18,8 @@ use uuid::Uuid;
 /// `operation_type` value of persons-seed runs — shared by the CLI runner
 /// (writes) and the `GET /v1/persons-seed*` journal endpoints (filter).
 pub const PERSONS_SEED_OP: &str = "persons-seed";
+/// Operation type of the persons→ClickHouse sync (`sync` subcommand).
+pub const PERSONS_SYNC_OP: &str = "persons-sync";
 
 /// Lifecycle phase of an operation. DB column is a `VARCHAR(16)`.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
--- a/src/backend/Cargo.lock
+++ b/src/backend/Cargo.lock
@@ -1433,6 +1433,7 @@ dependencies = [
  "bnum",
  "bstr",
  "bytes",
+ "chrono",
  "cityhash-rs",
  "clickhouse-macros",
  "clickhouse-types",
--- a/src/backend/Cargo.toml
+++ b/src/backend/Cargo.toml
@@ -33,7 +33,7 @@ dbg_macro = "deny"
 
 [workspace.dependencies]
 # ClickHouse client
-clickhouse = { version = "0.14", features = ["uuid"] }
+clickhouse = { version = "0.14", features = ["uuid", "chrono"] }
 
 sqlparser = "0.62"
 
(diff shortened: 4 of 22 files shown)


mozhaev-dev:
Both bot findings addressed (branch also rebased onto current main):

- **CronJob controls via values**: `sync.concurrencyPolicy` / `sync.backoffLimit` / `sync.activeDeadlineSeconds` added with defaults matching the previous literals; the template consumes them. (The seed CronJob keeps its literals — aligning it is a separate change outside this PR's scope.)
- **Stale-staging GC identifier validation**: tightened to the exact shape this code mints — `identity_persons_staging_` + 32 hex chars — before the name reaches the identifier-interpolated `DROP`, closing the quoted-identifier escape the prefix check alone left open.

Checks re-run locally: clippy pedantic 0, fmt clean, 81 unit tests, helm contract suite 17/17.

src/backend/services/identity-resolution/helm/values.yaml
```diff
@@ -93,6 +93,36 @@ seed:
...
+# Persons-sync CronJob: publishes the `persons` log into ClickHouse
+# `identity.identity_persons` (the metrics email→person_id resolve source).
+# Scheduled after the seed by default — the seed rewrites the log, the sync
+# publishes it.
+sync:
+  enabled: true
+  # Journal-row tenant (UUID); same override semantics as seed.tenantDefaultId
+  # (the copy itself is tenant-agnostic — this only scopes the operations
+  # journal the admin GET routes read).
+  tenantDefaultId: ""
```
cyberantonz, line 105: The insight-gitops should be modified for customer cluster or this value should be set from values.

mitasovr (approved):
Reviewed against the #1873 epic. Architecturally sound iteration 1 of the consumer path: the MariaDB persons log is the right publication point per ADR-0002 (any future matching improvements flow through the same pipe), the verbatim copy keeps resolution semantics downstream, and the full-snapshot + EXCHANGE swap with empty-log guard / count-verify / watermark backstop is proportionate to the blast radius of a bad publish. Known trade-offs (time-based seed→sync coupling, lockstep schema copies, PII now in a second store for #719 to account for) are acknowledged in-code and fine as follow-ups. Approving.

src/backend/services/identity-resolution/src/infra/db/ops_repo.rs
```diff
@@ -18,6 +18,8 @@ use uuid::Uuid;
 /// `operation_type` value of persons-seed runs — shared by the CLI runner
 /// (writes) and the `GET /v1/persons-seed*` journal endpoints (filter).
 pub const PERSONS_SEED_OP: &str = "persons-seed";
+/// Operation type of the persons→ClickHouse sync (`sync` subcommand).
+pub const PERSONS_SYNC_OP: &str = "persons-sync";
```
cyberantonz, line 22: Some strings declared twice will drift.

src/backend/services/identity-resolution/helm/values.yaml
mozhaev-dev: Already covered on the gitops path: `deploy/gitops/scripts/compose-app-secrets.sh` (the "First-admin bootstrap inputs" block) writes `APP__gears__identity-resolution__config__tenant_default_id` into `insight-identity-resolution-config` whenever `TENANT_DEFAULT` is configured for the cluster — and the sync CronJob consumes that same Secret via `envFrom`, so customer clusters need no gitops change. `sync.tenantDefaultId` here is the standalone-install override, mirroring the seed's pattern from #1690. And the tenant is only journal-row scoping for the sync — a cluster without it still runs via the sole-tenant inference fallback.

src/backend/services/identity-resolution/src/infra/db/ops_repo.rs
mozhaev-dev: Within the service each op-type string is declared exactly once: `grep '"persons-sync"'` over src/ hits only this const, and every consumer (sync_runner journal writes, the GET journal filter) goes through `ops_repo::PERSONS_SYNC_OP` — same for the seed const. The other spellings live across wire boundaries (the `/v1/persons-sync` route path, helm component labels, the e2e assertions) which can't share a Rust declaration; the contract e2e pins them together against the live service (`test_persons_sync` asserts `operation_type == "persons-sync"` end-to-end), so a drift fails CI rather than going quiet.

PR #2066, opened 2026-07-30
feat(dbt): resolve email → person_id in the metric observation tables (metrics person_id, iteration 1)

mozhaev-dev:
## What

The dbt half of the metrics person_id rework — companion to #2065 (which delivers `identity.identity_persons`, the persons-log copy in ClickHouse). Six commits, one step each:

1. **on-run-start hook** creating an empty `identity.identity_persons` when missing — a build on an environment where the sync never ran degrades gracefully (all resolves NULL, pipeline behaves exactly as today) instead of failing on a missing relation. Same dbt-owns-DDL / external-process-owns-data split as `staging.jira__task_field_history`; the schema is a documented lockstep copy of the service's `COLUMNS_DDL`.
2. **`resolve_person_id()` macro** — THE resolve point: collapses the observation log into the current `email → person_id` map (latest-observation-wins, `created_at DESC, id DESC`), normalized `lower(trimBoth())` byte-identical to the models' `entity_id`. Deliberately a macro, not a model: per-source maps / tenant scoping / as_of resolution are future body changes every consumer picks up on its next build.
3. **`person_id` in `git_metric_observations`** (pilot) — one LEFT JOIN on the final projection; join misses stay honest NULL via an explicit `if` (ClickHouse's LEFT JOIN default would mint the zero UUID). `entity_id` and everything the runtime reads is byte-identical — nothing consumes the column yet; it lands first so coverage is measurable and the later API cutover is a compiler/contract change, not a data change.
4. **The remaining four models** (`ai`/`collab`/`task`/`wiki`; six join sites — collab's final projection is two UNION ALL branches) via two companion macros so every site is one greppable shape.
5. **`person_id` in `metric_entity_cohorts_current`** — the peer-comparison membership, where an unmatched email distorts a whole team's percentiles. View caveat documented inline (join runs per peer query; cheap at the log's size). NOTE: cohorts are slated for replacement by the identity org chart — this column is a deliberate bridge, isolated to the view's file.
6. **`identity_resolution_coverage`** — activity-weighted match rate per source (+ `hr_cohorts`): `unresolved_rows` weighs what dashboards actually lose, `unresolved_people` counts the operator decisions that would close the gap. The "match rate is reported" outcome of #1873 for the metrics pipeline; baseline tracker for the ≈68% git→HR expectation.

## Verification

- every step verified through the metrics e2e rig (bronze seed → dbt build → batch API): one case per source green, metric values unchanged;
- resolve semantics live-tested on ClickHouse 25.7: re-linked email follows the newest binding, case/whitespace variants collapse, a second email of the same person maps independently, non-email observations excluded;
- retroactivity proven: seeding one email binding and rebuilding re-attributed all 11 of that author's observation rows — no backfill, just the next build;
- `/check-dbt-conventions` (incl. ADR-0001 read-dedup rules + the deterministic `audit_rmt_read_dedup.py`): 0 violations — the two new reads are of physically-unique producers, documented in-file.

Runtime is untouched: the API still keys on `entity_id` (email); switching the compiler/contract to `person_id` is the next iteration, gated on coverage.

Refs #1873; companion: #2065.

### DIFF
--- a/docs/domain/metrics/specs/DESIGN.md
+++ b/docs/domain/metrics/specs/DESIGN.md
@@ -31,6 +31,7 @@ tenant_id String,
 source_key String,
 entity_type String,
 entity_id String,
+person_id Nullable(UUID),
 metric_date Date,
 observed_at Nullable(DateTime64(3)),
 measure_key String,
@@ -45,6 +46,14 @@ Rules:
 - `source_key` identifies the logical source.
 - `measure_key` identifies the source measure.
 - `entity_type` and `entity_id` identify the measured entity.
+- `person_id` is the canonical person resolved from the identity log at
+  build time (`resolve_person_id` dbt macro; NULL = identity does not know
+  the email). ADDITIVE, not yet consumed: the runtime still keys on
+  `entity_id`, and the schema validator's `OBSERVATION_COLUMNS` /
+  `COHORT_COLUMNS` deliberately exclude it until the person_id API cutover
+  — probing for a column nothing reads would gate metric availability on
+  the next dbt rebuild after a deploy, for no reader's benefit. The cohort
+  view carries the same column under the same rule.
 - `observed_at` is reserved for future point-in-time semantics.
 - `subject_key` carries the counted subject for distinct-count measures (a
   date, a tool) and is NULL on every other measure's rows.
--- a/src/backend/services/analytics/src/domain/metric_definitions/validator.rs
+++ b/src/backend/services/analytics/src/domain/metric_definitions/validator.rs
@@ -394,6 +394,15 @@ impl MetricDefinitionValidator {
     }
 }
 
+// The columns the RUNTIME actually reads — a deliberate subset of the tables'
+// physical shape. `person_id` (present in every observation table and the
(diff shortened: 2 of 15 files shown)


mitasovr (approved):
Reviewed against the #1873 epic as the consumer half of iteration 1. Closes exactly the acceptance points reachable without the matching engine: activity-weighted match rate per source (the epic's "match rate is reported"), honest-NULL on join misses (explicit if() instead of ClickHouse's zero-UUID default), single resolve point in the macro, and exemplary rollout hygiene (additive column, on-run-start empty-table fallback, validator deliberately not probing person_id, cutover gated on coverage). One v1 limitation worth a follow-up, not a blocker: LIMIT 1 BY email silently awards a contested email to the latest observation — the epic's "never silently auto-merge" stance would prefer uniqExact(person_id) > 1 → NULL once operator edits / multi-source bindings appear. Also worth queuing: a CI check that the macro DDL stays byte-identical to the service's COLUMNS_DDL, and a trailing window for the coverage view. Approving.

cyberantonz (approved):

Issue #1744, opened 2026-07-13
New environment: Git + Task Delivery KPIs (Commits, PRs Merged, Tasks Completed, Cycle Time, …) never load until silver tables are manually reset


ktursunov:
## Summary

On a freshly built environment you connect a source (Bitbucket for **Git**, Jira for **Task Delivery**, M365 for **Collaboration**) and the sync reports success — but the dashboard KPIs never populate. They only start working after an engineer manually drops some internal tables and re-runs. The sync says "success", so the gap is silent and easy to mistake for "no data".

KPIs that stay blank until the manual reset: **Git** — Commits, PRs Created/Merged, PR Cycle Time, PR Size, LOC / Clean LOC; **Task Delivery** — Tasks Completed, MTTR, Pickup/Dev Time, Estimation Accuracy, On-Time Delivery, Flow Efficiency, Reopen Rate, Logged Work; **Collaboration** — Meeting Hours/Count, Teams Chats, Emails.

## Steps to Reproduce

1. Build a fresh environment (so the internal "starter" placeholder tables are freshly created).
2. Connect Bitbucket (or Jira / M365), let it sync — it reports success; confirm bronze is populated.
3. Open the **Git** (or **Task Delivery**) dashboard for a period covering that work.

**Expected:** the tiles (Commits, PRs Merged, …) show numbers.
**Actual:** every tile says "No data for this period", though the rows are in bronze — the processed rows land in a temp staging table but never reach the table the dashboard reads:
```sql
SELECT count() FROM bronze_bitbucket_cloud.commits;               -- 84  (sync OK)
SELECT count() FROM silver.class_git_commits;                     -- 0   (final table empty)
SELECT count() FROM silver.class_git_commits__dbt_new_data_<id>;  -- 84  (stuck in temp staging)
```

## Root Cause

The deploy-time runner [`apply-ch-migrations.sh`](https://github.com/constructorfabric/insight/blob/main/src/ingestion/scripts/apply-ch-migrations.sh) (the `clickhouse-migrate` Helm hook) bootstraps in a fixed order: create databases → [run `create-bronze-placeholders.sh`](https://github.com/constructorfabric/insight/blob/main/src/ingestion/scripts/apply-ch-migrations.sh#L45) → apply migrations → build gold.

[...]

PR #2077, opened 2026-07-30
feat(bootstrap-db): audit staging -> silver field parity + connectors-ddl CI gate

mitasovr:
## Why

Two contracts are invisible to every existing lane, because both are only observable in a real ClickHouse where every model is materialised at once:

1. **The committed `connectors-ddl` snapshot.** A gold/staging/bronze relation added without regenerating it simply will not exist on the next fresh cluster (#1744). Today the only guard is a sticky reminder comment, and an ignored reminder reintroduces exactly that drift.
2. **staging -> silver field parity.** A silver `class_*` model is a `UNION ALL` of every staging model tagged `silver:<target>` (`union_by_tag`). ClickHouse matches UNION branches **by position** and takes the column names from the first branch, so a contributor that renames, reorders or retypes a column does not fail the build — it silently misaligns data, or widens the published silver type depending on which connectors happen to be enabled.

This PR adds the audit for (2) and a CI lane that gates both.

## What

**`src/ingestion/scripts/bootstrap-db/check-field-parity.py`** — reads `system.columns` for the structure and `dbt/target/manifest.json` for the staging -> silver mapping (that mapping lives only in the dbt tags, not in the database), then fails on any divergence:

1. **coverage** — every non-ephemeral model in the manifest has a relation. Without this, a connector whose `discover` failed would shrink the comparison silently and the audit would pass for the wrong reason.
2. **columns** — contributor and target expose the same column names.
3. **order** — identical column positions, because the UNION is positional.
4. **types** — byte-identical `system.columns.type`; `Nullable(T)` vs `T` counts, since that split decides the published silver type.

No baseline file, no warning tier: everything is a hard failure. Same `CLICKHOUSE_*` env contract as the sibling scripts, and it sources the local `.env` the way `bootstrap-db.sh` does (`--no-env-file` to audit another cluster).

Ephemeral contributors are covered too. `jira__task_field_history` is an ephemeral pass-through over `staging.jira__task_field_history`, a table written by the `jira-enrich` Rust binary whose DDL lives in the `create_task_field_history_staging` macro (ADR-003) — so it has no dbt DAG edge to its silver target. The audit follows the `source()` dependency to that table, which means a future YouTrack twin of the enrich table gets the same guard for free, as long as it keeps the shape (ephemeral pass-through + the `silver:<target>` tag).

**`.github/workflows/connectors-ddl.yml`** — one empty ClickHouse pinned to the production version from `pins.env`, then straight into `bootstrap-db.sh` (real connector `discover`, the real destination-clickhouse connector, real dbt, real migrations) -> re-dump -> fail on any snapshot diff -> audit field parity.

The committed snapshot is deliberately **not** applied first. `create-bronze-placeholders.sh` issues `CREATE ... IF NOT EXISTS`, so a relation the snapshot still carries but the current code no longer produces would survive into the fresh dump, hide its own deletion from the diff and break convergence — the same reason `bootstrap-db.sh` sets `BOOTSTRAP_SKIP_SNAPSHOT=1` during generation.

`pull_request` (never `pull_request_target`) plus a head-repo guard, so the lane runs only for branches in this repository: the job executes PR code and needs the HubSpot / Salesforce credentials whose CDK `discover` calls a live API, and secrets never reach fork PRs. Those secrets are checked up front, so a missing one fails in seconds instead of 20 minutes into the bootstrap. Drift fails with the regeneration command, the first 200 diff lines inline and the full diff as an artifact; the job never commits.

## Findings this already surfaces

**The snapshot was stale.** `feat(gold): add metric evidence serving tables` (6e2c56b1) added six gold relations without regenerating it, so a fresh cluster would not pre-create them. Regenerated here in its own commit — without it the new lane is red on arrival.

**Field parity, over a full bootstrap warehouse** (270 relations, 39 union targets): 74 findings — 66 nullable-only, 8 representation, 0 coverage gaps, 0 unchecked. Column sets and column order are clean everywhere. The 8 representation-level ones:

| Divergence | Where |
|---|---|
| `commit_order` `UInt8` vs `Int64` — **overflows at the 256th commit in a PR** | github, gitlab -> `class_git_pull_requests_commits` |
| `hire_date`, `termination_date` `Nullable(Date)` vs `Nullable(DateTime)` | active_directory, ms_entra -> `class_people` |
| `visited_page_count` `Nullable(Int64)` vs `Nullable(Decimal(38, 9))` | m365 onedrive -> `class_collab_document_activity` |
| `close_date` `Nullable(Date)` vs `Nullable(Date32)` — `Date` cannot hold pre-1970 | hubspot -> `class_crm_deals` |

Fixing those is deliberately **not** in this PR. Note the consequence: **the lane is red until they are fixed** — either land the type alignments first, or merge this and treat the lane as advisory until it is made a required check.

## Verification

Every check was proven to trip, by mutating an empty table in a throwaway warehouse and restoring it (`SHOW CREATE` identical before/after):

| Case | Result |
|---|---|
| `ADD COLUMN probe_extra String` | `FAIL columns … extra` |
| `DROP COLUMN report_period` | `FAIL columns … missing` |
| `MODIFY COLUMN calls_count Int32` | `FAIL types … [representation]` |
| `MODIFY COLUMN user_name … AFTER calls_count` | `FAIL order … positional UNION mismatch` + both position lists |
| enrich table absent | `FAIL coverage … read by the ephemeral contributor … but absent` |
| enrich table retyped / column dropped | `FAIL … (via ephemeral pass-through) …` |

The lane's own steps were rehearsed against a real warehouse: `dump-ddl.sh` + the diff gate is what caught the stale snapshot above, and the strict audit is the run reported in "Findings". An earlier draft also applied the committed snapshot to an empty cluster and audited that; measured, it compared 1 of 141 contributors (153 relations absent, since the snapshot keeps only the gold-referenced staging tables), so it was dropped.

The pass-through detector was unit-checked against `SELECT *` (with a `-- depends_on` hint and a trailing `;`) plus the shapes it must reject: explicit column list, `SELECT *, 1 AS extra`, `WHERE`, `JOIN`, CTE — those resolve to UNCHECKED rather than being compared against the wrong relation.

## Needs doing before this lane can pass

- **Repository secrets**: `HUBSPOT_ACCESS_TOKEN`, `SALESFORCE_CLIENT_ID`, `SALESFORCE_CLIENT_SECRET`, `SALESFORCE_INSTANCE_URL`. Their CDK `discover` calls a live API, so fake values fail. The workflow checks them up front and fails in seconds rather than 20 minutes into the bootstrap. The documented alternative — seeding those two connectors' bronze from the snapshot instead — is not wired up here, because it would freeze exactly the schemas the gate is meant to watch.
- **No path filter**, matching `e2e-bronze-to-api.yml`'s convention (PR-only, all paths). If the ~40 minutes on every PR is too much, add a `paths:` filter on `src/ingestion/**`.
- `connectors-ddl-reminder.yml` is now redundant with a real gate; retiring it is a separate call, so it is left in place.

## Commits

Five, each droppable on its own: the audit script, the `connectors-config.yaml` regeneration (two spec-required fields moved), the CI lane (plus three doc references that described the snapshot as "CI-generated" — CI regenerates it to compare, but the committed file is produced by a human), the snapshot regeneration for the metric-evidence tables, and the follow-up that drops the snapshot-apply phase.

## Summary by CodeRabbit

* **New Features**
  * Added new metric evidence and task worklog data tables to support richer reporting and traceability.
  * Added automated validation for connector schema consistency, including field coverage, ordering, and data types.
* **Bug Fixes**
  * Pull requests now detect connector DDL drift before changes are merged.
  * Improved connector configuration examples and updated Salesforce date settings.
* **Documentation**
  * Expanded setup guidance and documented staging-to-silver field parity auditing and validation results.


mitasovr:
Superseded by #2080 — same commits, but the branch now lives in this repository instead of a fork, so the new `connectors-ddl` lane this PR adds can actually run on it (the lane is guarded by `head.repo.full_name == github.repository`, because it needs the HubSpot / Salesforce secrets that fork PRs never receive).

PR #2082, opened 2026-07-30
fix(analytics): authorize person ids on metric-results

### LINKED ISSUE #2035: Any logged-in user can view any employee's personal dashboard by changing the email in the URL
## Summary
Any authenticated user can view any other employee's personal dashboard — their metrics, name, and job title — by changing the email in the URL, with no authorization check.

## Steps to Reproduce
1. Log in as any user (e.g. an IC whose only reports are a small team, or none).
2. Navigate to `/ic/<another-employee-email>/personal` for someone with no reporting relationship to you — a colleague in a different department, or a C-level executive.
3. That person's name, job title, and personal metrics render (Focus Time, Emails Sent, Files Shared, Pull requests merged, AI usage, etc.).

The SPA route just calls the analytics serving endpoint with a client-supplied person id, which returns `200` with that person's data:

```
POST /api/analytics/v1/metric-results
{"entity":{"type":"person","ids":["<any-employee-email>"]},
 "period":{"from":"2026-06-30","to":"2026-07-29"},
 "metrics":[{"metric_key":"collab.focus_time_pct","views":[{"view":"period"},{"view":"peer"}]}, ...]}

→ 200 OK, body contains the target person's metric values
```

The distinguishing signal is that the response is `200` with real data, not `403`. It holds for any email of a real person in the tenant, verified across departments and up to C-level — none with a reporting relationship to the caller.

**Expected:** A user can load only their own personal dashboard (and, for a manager, their reports); requesting an unrelated person returns `403`/empty.
**Actual:** The serving path returns any person's metrics to any authenticated caller in the same tenant.

aleksdotbar:
## Problem

`POST /v1/metric-results` returned any person's metrics to any authenticated caller in the tenant. The handler read the `SecurityContext` for its tenant id only; nothing checked that the requested `entity.ids` were the caller or within their reporting scope.

## Fix

Identity answers the authorization question, analytics enforces it.

- **`POST /v1/visible-persons`** (identity-resolution): takes emails, returns the subset the caller may see. The existing visible-set predicate — caller, active grants, whole tenant on a wildcard grant, `org_chart` descendants — materialised once and joined against the resolved candidates, instead of one traversal per person. A wildcard grant short-circuits the traversal entirely.
- **The gate** (analytics): requested person ids are filtered through that endpoint; any id outside the answer refuses the whole request with `403`. No partial responses — a filtered `200` is indistinguishable from absent data.

No semantic change to who may see whom. The `admin` role still confers no visibility, and an email is visible when any person it resolves to is, matching `POST /v1/profiles`.

Failing to reach identity is a server error, never a denial, so an authorization backend that is down or unconfigured cannot read as "you may not see this".

Emails resolve by input position: `persons.value_id` compares case- and accent-insensitively, so the stored spelling cannot be mapped back to the request reliably. A candidate whose stored email merely compares equal — a different person under a loose collation — is dropped, since consumers key metric rows by the email bytes.

## Verification

Live-MariaDB cases assert the predicate: a caller with no reports still sees themselves, a manager sees a transitive descendant and not an unrelated person, an explicit grant reaches outside the reporting line, a wildcard grant covers the tenant, and an `admin` role holder sees no one extra. Gate cases cover admit, refuse, whole-request refusal, service-subject bypass, unconfigured identity, and an identity that predates the endpoint (`404`/`405` → 5xx, not 403). Route-table cases drive the real operation table.

## Known gaps, not addressed here

- `POST /v1/metrics/{id}/query` and `POST /v1/metrics/queries` interpolate client-supplied `person_id` / `org_unit_id` filter values with no visibility check — the same exposure class, reachable without touching this endpoint. Now documented as planned rather than implemented.
- `identityResolution.deploy` defaults to `false` while `identity_url` is set unconditionally, so a default chart install refuses person metric requests. Needs a chart change and a release note.
- `migrations_and_bootstrap_are_idempotent_against_live_mariadb` fails on a pristine migrated database, independent of this branch — no migration or SQL file is touched here.

ADR-0015 records the non-admin read, relaxing ADR-0012 for one self-scoped query.

Closes constructorfabric/insight#2035


src/backend/services/analytics/src/infra/identity/mod.rs
```diff
@@ -126,12 +145,17 @@ pub struct IdentityClient {
...
     /// Create a new client. `base_url` is the identity service root,
     /// e.g. `http://insight-identity-resolution:8082`.
-    #[must_use]
-    pub fn new(base_url: &str) -> Self {
-        Self {
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the HTTP client cannot be built.
+    pub fn new(base_url: &str) -> anyhow::Result<Self> {
```
cyberantonz, line 152: #[must_use] lost

src/backend/services/identity-resolution/src/api/visible_persons.rs
```diff
@@ -0,0 +1,181 @@
...
+    Ok(Json(VisiblePersonsResponse { visible }))
+}
+
+fn normalize_emails(emails: &[String]) -> Result<Vec<String>, CanonicalError> {
+    if emails.len() > MAX_EMAILS {
+        return Err(invalid(&format!("at most {MAX_EMAILS} emails per request")));
+    }
+
+    let mut seen: HashSet<String> = HashSet::with_capacity(emails.len());
+    let mut out: Vec<String> = Vec::with_capacity(emails.len());
```
cyberantonz, line 108: you can get in emails
```
[
`person@email.nonpresent`,
`   person@email.nonpresent`,
]
```
Those are dirrefent strings in hash obviously.

After the trim you will have Vec with two entries 
```
[
`person@email.nonpresent`,
`person@email.nonpresent`,
]
```

Why not to use HashSet for the output?

cyberantonz:
The queries is heavy, can we add caching for the requests?

src/backend/services/analytics/src/infra/identity/mod.rs
aleksdotbar: anyhow::Result is itself #[must_use], so discarding the return value already warns via the type. Adding the attribute to the function is redundant and clippy flags it

src/backend/services/identity-resolution/src/api/visible_persons.rs
aleksdotbar: Trim happens before the dedup check, and the key is the trimmed+lowercased form:

```rust
let email = email.trim();
if email.is_empty() { continue; }
if seen.insert(email.to_ascii_lowercase()) { 
    out.push(email.to_owned());
}
```

aleksdotbar:
> The queries is heavy, can we add caching for the requests?

I dont think we need it here. At least not in this PR. We can measure and see how it behaves and then decide.

cyberantonz (approved):

PR #2083, opened 2026-07-31
fix(ingestion): resolve nltk and langchain-core to fixed versions (#2061)

### LINKED ISSUE #2061: Connector and harness dependency graphs resolve to CRITICAL nltk and langchain-core
Eight resolved dependency graphs land on packages that have a released fix for a CRITICAL. Visible on main as `trivy-fs`.

| Package | Resolved | Fixed in | CVE | Where |
|---|---|---|---|---|
| `nltk` | 3.9.1 | 3.9.3 | CVE-2025-14009 | 7 CDK connectors + the nocode test harness |
| `langchain-core` | 0.1.42 | 0.3.81 | CVE-2025-68664 | the nocode test harness only |

## Cause

`airbyte-cdk` pins both to exact versions, so no downstream project can move them by declaring a floor:

| CDK | nltk | langchain-core |
|---|---|---|
| 6.60.x (harness) | `==3.9.1` | `==0.1.42`, unconditional |
| 7.23.x (connectors) | `==3.9.1` | `>=1.2.5`, only under the `vector-db-based` extra |

The connector **images** are already clean: each Dockerfile installs `nltk==3.10.0` with `--no-deps` over the pin. What is broken is that the declared graph disagrees with the shipped image, so anything resolving from `pyproject.toml` — the scanner, or `uv sync` on a developer machine — still gets 3.9.1.

Neither package is reachable from the code these projects run: `nltk` serves the file-based unstructured parser and `langchain-core` the vector-db sink.

## Fix

`[tool.uv] override-dependencies` in all eight projects, pointing at the fixed versions.

Raising the harness to CDK 7 instead is **not** viable. Measured against `--suites-only`:

| CDK | Result |
|---|---|
| 6.60.16 (current pin) | 42 passed, 1 skipped |
| 7.23.8 | 6 failed + 1 error, 36 passed |

Every failure is in `collaboration/zoom/tests/test_participants.py` (the parent-child substream suite): CDK 7 issues a different parent request, `GET /v2/metrics/meetings?type=past&page_size=100&from=…&to=…`, which no mock matches, so the stream yields 0 records instead of 2. That is a change in the HTTP the declarative engine emits, not a fixture detail, and it belongs in its own change with the zoom manifest.

## Verified

`uv pip compile` + `trivy fs --severity CRITICAL` on each of the eight projects: **0 CRITICAL**, resolving `nltk==3.10.0` and `langchain-core==1.5.3`. Harness suites with the overrides applied: 42 passed, 1 skipped — identical to baseline.

Refs #2020

### LINKED ISSUE #2102: No .gitignore rule for build output, so stale artifacts can be committed
Nothing tracked in this repository produces `obj/` or `bin/` output, but `.gitignore` carries no rule for them. A stale working tree still holds such directories, and a broad `git add` commits them back — which happened in #2083 and forced that branch to be rebuilt to drop 25 files.

These artifacts also embed machine-local absolute paths, so they are noise in any diff.

## Fix

Add ignore rules for `obj/` and `bin/`. Verified safe: the repository tracks 0 files under any such directory, so nothing is retroactively ignored.

Gregory91G:
Closes #2061, closes #2102.

`airbyte-cdk` pins `nltk==3.9.1` in all seven CDK connectors and, in the 6.60.x line the nocode harness runs on, `langchain-core==0.1.42`. Both have a released fix for a CRITICAL. An exact upstream pin cannot be moved by declaring a floor downstream — a resolver reports the conflict and fails — so the version is stated as `[tool.uv] override-dependencies`.

| Package | Was | Now | CVE |
|---|---|---|---|
| `nltk` | 3.9.1 | 3.10.0 | CVE-2025-14009 |
| `langchain-core` | 0.1.42 | 1.5.x | CVE-2025-68664 |

The connector images were already clean: their Dockerfiles install `nltk==3.10.0` with `--no-deps`, which skips resolution entirely, so the fixed version lived only inside the image while the declared dependencies still resolved to 3.9.1. Scanners and `uv sync` read the declared side. This makes the two agree.

`exclude-newer = "7 days"` accompanies each override: an override that only sets a floor would otherwise resolve straight into a release published minutes ago, which is where a compromised maintainer account surfaces first.

Raising the harness to CDK 7 was measured and rejected: 7 failures in the zoom parent-child substream suite, because CDK 7 emits a different parent request. Details in #2061.

## Test plan

- [x] `uv pip compile` + `trivy fs --scanners vuln --severity CRITICAL` on each of the 8 projects — **0 CRITICAL** (was 8), resolving `nltk==3.10.0` and a fixed `langchain-core`
- [x] Re-verified after adding the cooldown: resolution still succeeds and still lands above both fixes; the cooldown demonstrably held the newest `langchain-core` back one patch, as intended
- [x] Harness suites with the overrides resolved by `uv`: `pytest --suites-only` → 42 passed, 1 skipped — identical to the pre-change baseline
- [x] Connector runtime unchanged: `source-gitlab` emits a valid `SPEC`, and the image already carried 3.10.0
- [ ] `trivy-fs` on this branch reports 8 fewer CRITICAL than the base

### DIFF
--- a/src/ingestion/connectors/ai/github-copilot/pyproject.toml
+++ b/src/ingestion/connectors/ai/github-copilot/pyproject.toml
@@ -22,6 +22,9 @@ dev = [
 source-github-copilot = "source_github_copilot.source:main"
 
 [tool.uv]
+# Do not resolve a version published in the last week: a freshly pushed release is
+# where a compromised maintainer account shows up first.
+exclude-newer = "7 days"
 # airbyte-cdk pins nltk==3.9.1 (fixed CRITICAL CVE-2025-14009); the image installs 3.10.0
 # over it (see Dockerfile), so keep the resolved dependency graph on the same version.
 override-dependencies = ["nltk>=3.10.0"]
--- a/src/ingestion/connectors/ai/github-copilot/pyproject.toml
+++ b/src/ingestion/connectors/ai/github-copilot/pyproject.toml
@@ -21,6 +21,11 @@ dev = [
 [project.scripts]
 source-github-copilot = "source_github_copilot.source:main"
 
+[tool.uv]
+# airbyte-cdk pins nltk==3.9.1 (fixed CRITICAL CVE-2025-14009); the image installs 3.10.0
+# over it (see Dockerfile), so keep the resolved dependency graph on the same version.
+override-dependencies = ["nltk>=3.10.0"]
+
 [tool.pytest.ini_options]
 testpaths = ["tests"]
 
--- a/src/ingestion/connectors/crm/hubspot/pyproject.toml
+++ b/src/ingestion/connectors/crm/hubspot/pyproject.toml
@@ -22,6 +22,11 @@ dev = [
 [project.scripts]
 source-hubspot-insight = "source_hubspot.source:main"
 
+[tool.uv]
+# airbyte-cdk pins nltk==3.9.1 (fixed CRITICAL CVE-2025-14009); the image installs 3.10.0
+# over it (see Dockerfile), so keep the resolved dependency graph on the same version.
+override-dependencies = ["nltk>=3.10.0"]
+
 [tool.pytest.ini_options]
 testpaths = ["tests"]
 
--- a/src/ingestion/connectors/crm/hubspot/pyproject.toml
+++ b/src/ingestion/connectors/crm/hubspot/pyproject.toml
@@ -23,6 +23,9 @@ dev = [
 source-hubspot-insight = "source_hubspot.source:main"
 
 [tool.uv]
+# Do not resolve a version published in the last week: a freshly pushed release is
+# where a compromised maintainer account shows up first.
+exclude-newer = "7 days"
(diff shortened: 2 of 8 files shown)


cyberantonz:
Why you change removed .NET insight?

cyberantonz:
By the way, how the python can get old nltk if the installed version is already 3.10 ?

Gregory91G:
> Why you change removed .NET insight?

fixed

Gregory91G:
> By the way, how the python can get old nltk if the installed version is already 3.10 ?

`trivy fs` is checking `pyproject.toml`, not an actual image. So the fix fixes dependency graph in `pyproject.toml`.

And `nltk>=3.10.0` under [project.dependencies] cannot be used it because it contradicts with strict `nltk==3.9.1` from `airbyte-cdk`

.gitignore
```diff
@@ -423,3 +423,8 @@ trivy-fs.sarif
 trivy-image.json
 trivy-image.sarif
 trufflehog-findings.jsonl
+
+# .NET build output. No .csproj is tracked today, but a working copy predating the
```
cyberantonz: useless comment and ignored files

cyberantonz (approved):

Issue #2020, opened 2026-07-29
ci(security): complete vulnerability reporting (SCA, SAST, secrets, images)


Gregory91G:
Bring the backend's vulnerability reporting to full coverage: every dependency ecosystem, every published image, every Helm chart, and the secret classes a refs-based scan cannot reach. Reporting only — enforcement is tracked separately.

Frontend equivalent: constructorfabric/insight#2079.

Follows #1478, which set up the first scanners, and #1797, which added the Semgrep gate. Neither a TruffleHog nor a Trivy workflow existed in this repository before this work.

## Where coverage stood

`semgrep.yml` was the only security workflow. `trivy fs` reported **0 vulnerabilities**, which read as good news and was not: Trivy resolves dependency versions from lock files and this repository commits exactly one, `Cargo.lock`. Five `.csproj` and eleven `pyproject.toml` projects were skipped in silence. Seven of nine Helm charts were skipped too, because they `required` values at render time. No image was ever scanned.

## Where it stands now

| Scope | Findings | Note |
|---|---|---|
| Repository (`trivy-fs`) | **149** — 9 CRITICAL, 61 HIGH, 47 MEDIUM, 32 LOW | was 0; **102 come from Python**, .NET resolves clean |
| Helm charts (`trivy-helm`) | **124** — 16 HIGH, 29 MEDIUM, 79 LOW | was 21; a new class appears — `KSV-0109`, secrets in a ConfigMap, in identity-resolution |
| Images, 14 of them (`trivy-image:*`) | **664** — 73 CRITICAL, 588 HIGH | 693 of them trace to one line: all seven connectors build `FROM airbyte/python-connector-base:2.0.0`, two majors behind 4.1.1 |
| Secrets | 253 findings, **0 live-verified** | all triaged as detector false positives on synthetic fixtures, docs prose and dbt identifiers |
| SAST (`semgrep`) | 172 | 76 of them one rule — mutable action tags |

## Delivered by

- #2002 — TruffleHog secret gate and Trivy SCA, plus the image matrix over all 14 published images
- #2016 — resolve lock files for scanning, render the Helm charts, report every severity, add the GitHub API secret pass, and make every check report-only

Closed by #2016. Enforcement is deliberately outside this issue and lives in #2022, which stays a draft until the preconditions listed there are met.

## Two scanner defaults worth remembering

[...]

Issue #1340, opened 2026-06-15
[Security][Critical ] trivy-config — 8 HIGH IaC and Dockerfile misconfigurations (root containers, frontend securityContext)


SharedQA:
## Summary

The `trivy-config` gate (added in #1317) surfaces 8 HIGH IaC/Dockerfile misconfigurations. They are pre-existing and mostly on build/test images; waived in `.trivyignore` (4 rule IDs, documented) so the gate stays blocking for any *new or other* misconfiguration. This issue tracks fixing them and removing each waiver.

## Findings (4 rules / 8 occurrences)

| Rule | Sev | Where | Note |
|---|---|---|---|
| AVD-DS-0002 (container runs as root) | HIGH ×4 | `compose/seed/Dockerfile`, `src/ingestion/tests/e2e/compose/Dockerfile.runner`, `src/ingestion/tools/declarative-connector/Dockerfile`, `src/ingestion/tools/toolbox/Dockerfile` | build/test/tooling images — add non-root `USER` where feasible |
| AVD-KSV-0118 (Deployment missing securityContext) | HIGH ×2 | `src/frontend/helm/templates/deployment.yaml` | **real prod hardening** — set `runAsNonRoot`, drop caps |
| AVD-KSV-0014 (root filesystem not read-only) | HIGH ×1 | `src/frontend/helm/templates/deployment.yaml` | set `readOnlyRootFilesystem: true` |
| AVD-DS-0029 (apt-get without --no-install-recommends) | HIGH ×1 | `src/ingestion/tools/toolbox/Dockerfile` | add the flag |

## Acceptance

- Fix the frontend Deployment securityContext (KSV-0118/0014) first — that's the genuine prod gap.
- Add non-root `USER` to build images where it doesn't break them; `--no-install-recommends` on toolbox.
- Remove each rule from `.trivyignore` as it's fixed. Gate stays blocking for everything else.

SharedQA:
Re-scanned with trivy config (0.71.1) against merged main (`08f0527`, post-#1314). Same four rule IDs on the same files — the gears migration introduced no new misconfiguration, and the api-gateway helm templates it *did* change (configmap.yaml/secret.yaml) produce no HIGH finding:

- DS-0002 (root container) x4 — compose/seed/Dockerfile, e2e/compose/Dockerfile.runner, tools/declarative-connector/Dockerfile, tools/toolbox/Dockerfile
[...]

PR #2093, opened 2026-07-31
ci(security): pin third-party actions to digests, add Dependabot

Gregory91G:
Refs #2020, #1340.

A `uses: owner/action@v3` reference is resolved at run time from a mutable tag. Repointing that tag is normal release practice for the action owner — and the exact capability someone gains by taking over the account, after which their code runs on our runners with the job token. A digest cannot be repointed.

Every action is pinned, `actions/*` included: living in GitHub's own organisation lowers the risk but does not remove it, and a partial rule leaves a reader unable to tell a deliberate exemption from an oversight. `dtolnay/rust-toolchain@stable` is the one exception — it tracks a branch by design, and pinning it would freeze the toolchain.

Pinning alone would freeze everything, so `.github/dependabot.yml` refreshes weekly with a seven-day cooldown, so a release published minutes ago is never resolved. Two groups (`actions`, `base-images`) keep a failing base-image rebuild from holding back an unrelated action bump. Review routing comes from `.github/CODEOWNERS` rather than a named assignee, so it follows ownership. `npm` and `cargo` are out of scope for now.

62 references across six workflows. `build-images.yml` (64 more) is a separate PR — it publishes every image and the umbrella chart, so it stays reviewable and revertable on its own.

Spec: `sop/scanner-image-refresh.md` — it covers scanner images, which are invoked in shell steps and so are invisible to Dependabot (manual, quarterly).

## Test plan

- [x] All 15 distinct tags resolved through the GitHub API; 62 references rewritten, count matched per file (ci 18, e2e 22, openapi 8, authenticator 7, gateway 5, ghcr-cleanup 2)
- [x] No reference left on a tag in the six workflows
- [ ] Every workflow in this PR parses and its jobs start
- [ ] Dependabot opens its first grouped PR on the next Monday tick


.github/dependabot.yml
```diff
@@ -0,0 +1,32 @@
...
+# ecosystem cannot see. They are refreshed by hand — see
+# docs/components/deployment/specs/sop/scanner-image-refresh.md.
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+      day: monday
+    assignees: [Gregory91G]
```
cyberantonz: Why not mainteiners?

docs/components/deployment/specs/ADR/0002-pin-actions-to-digests.md
```diff
@@ -0,0 +1,41 @@
+# ADR-0002 — Pin third-party actions to commit digests, refresh with Dependabot
```
cyberantonz: Are we really need this document?

.github/dependabot.yml
Gregory91G: Fixed. Didn't know about the option

docs/components/deployment/specs/ADR/0002-pin-actions-to-digests.md
Gregory91G: I thought it should be documented. If it shouldn't - I'll remove it

cyberantonz: it is `good practices`, why we need to document those practices?

cyberantonz (approved):

PR #2073, opened 2026-07-30
feat(metrics): add metric drilldown endpoint

aleksdotbar:
## Summary

- add `POST /v1/metric-drilldown`: paginated, evidence-backed supporting data preserving the metric's entity, period, filters, and breakdown selection
- server-owned human-facing columns; ratio metrics expose named input columns
- cursor pagination bound to the evidence snapshot; rebuilds mid-pagination fail explicitly
- document drilldown selection rules and the authoring impact in the design doc

## Stack

Merges bottom-up; each PR retargets to `main` as its parent lands.

1.    #2071 — evidence serving tables
1.    #2072 — capability metadata
1. **→** #2073 — drilldown endpoint
1.    #2074 — CSV/XLSX export

UI layer: constructorfabric/insight-front#226

Closes constructorfabric/insight#2069

## Validation

- `cargo test -p analytics`
- e2e api contract tests for the new endpoint

## Summary by CodeRabbit

* **New Features**
  * Added metric drilldown API endpoint with support for filtering, pagination, and entity selection for detailed metric analysis.

* **Documentation**
  * Added API documentation and design specifications for metric drilldown functionality.

* **Tests**
  * Added integration and end-to-end tests for metric drilldown API validation and error handling.

PR #2098, opened 2026-07-31
feat(analytics,identity): metric-results keys on person_id — the identity cutover (⚠️ merge together with the frontend)

mozhaev-dev:
> ## ⚠️ DO NOT MERGE ALONE — breaking wire change, needs the frontend in lockstep
>
> This flips `POST /v1/metric-results` from **email** entity ids to **canonical person UUIDs**. An email in `entity.ids` is now a `400`. The current frontend sends emails (session email, `/ic/<email>/personal` routes), so merging this without the paired `cyber-insight-front` change takes every dashboard down.
>
> **Kept as a draft on purpose** — that is the mechanical guarantee against an accidental merge. Reviewable as-is; flip to ready only when the frontend PR is open and both can go in together (this one first, so the FE never speaks UUID to an email-only backend).

## What

Iteration 2 of the metrics `person_id` rework: the runtime now keys on the canonical person the identity pipeline resolves into the gold tables (delivered by #2065 + #2066). Two people's accounts that resolve to one person finally read as one person; rows identity cannot resolve (`person_id IS NULL`) drop out of every query — the epic's *excluded, not guessed*, now enforced by the key itself.

Also **integrates and re-contracts #2082** (@aleksdotbar's visibility gate). Its commits are merged here, and its whole stack moved from emails to person UUIDs — see the coordination note at the bottom.

## How

**Validation** — `entity.ids` parse as UUIDs into `person_ids: Vec<Uuid>` (parse-don't-validate). The pre-cutover email shape and the nil UUID are both loud `400`s, never silent empty results. The email lowercasing normalizer is gone.

**Compiler** — every view (period / timeseries incl. capped / breakdown / histogram / ranking / peer) filters `person_id IN (…)`, groups and orders by it, projects `toString(assumeNotNull(person_id))`.

**Peer** — the cohort join runs `person_id = person_id` instead of string email equality, the fragile HR-email↔source-email match this rework exists to kill. Two hardenings, both from review:
- the cohort view is unique per **email**, so two emails resolving to one person could double-weight a peer → both cohort CTEs collapse to person grain;
- **contested membership is excluded, not tie-broken**: `HAVING uniqExact(cohort_id) = 1`, evaluated over the person's *complete* membership set (the target-cohort filter applies outside the guard — filtering first would hide the conflicting row and wave the person through). A person whose emails claim different departments leaves peer comparison until the org assignment is fixed.

**Authorization** — the gate now forwards validated person UUIDs to identity's `POST /v1/visible-persons` (which takes `person_ids` and answers UUIDs; its email-resolution layer, needed only as the pre-cutover bridge, is deleted along with `resolve_person_ids_by_emails`). It runs **before** any ClickHouse work; service principals bypass; a single non-visible id refuses the whole request rather than silently dropping an entity.

**Schema validator** — `person_id` joins `OBSERVATION_COLUMNS` / `COHORT_COLUMNS`, as their comment promised: a table without it would 500 every query, so the probe now marks the source unavailable upfront.

**Naming discipline** — everything carrying the person UUID is named `person_id`/`person_ids` through validation, compiler rows, batch demux and builder. `entity_id` survives only as the wire field name, documented as a seam at the single builder/DTO boundary.

**Contracts** — metrics DESIGN, `gold/schema.yml`, request DTO docs, and both OpenAPI specs (analytics + identity, the published one described a contract that no longer existed). The person-only entity contract is stated in code where a first non-person type will land: ids parse as person UUIDs for every type and the gate rules on `person` alone (fail-closed), so a new type must bring both halves.

## Testing

- `cargo test --workspace`: 514 + 89 + … , 0 failed · clippy pedantic workspace-clean · fmt clean.
- Coverage of the touched code: gate `person_visibility.rs` **100%**, `compiler.rs` 97.8%, `builder.rs` 92.9%, `batch.rs` 89.3%, `validation.rs` 87.6%.
- The gate's ignored live tests were rewritten: they spoke the email contract, so the 403 case would have died on validation and the not-denied case asserted only `!= 403` — which that 400 satisfies. Both now use named person UUIDs, and the not-denied case asserts the exact 500 from the unreachable ClickHouse, so it fails if the request stops earlier.
- e2e: **metrics 35/36**, **api 93 passed**, **identity 133 passed**. Endpoint-coverage gates pass: analytics 28/28 operations (metric-results' `403` is now genuinely observed, not BLOCKED boilerplate), identity 19/19.
- The metric yaml rig stays authored in emails — the readable persona key — and owns the translation: personas get deterministic uuid5 ids, `identity.identity_persons` is seeded before the gold build so the dbt resolve macro attributes rows, requests translate email→UUID and responses back. Two latent traps closed: the stub's visible set is derived from the yaml under test (a new persona can't fail as a phantom 403), and a UUID collision in the reverse map raises instead of silently dropping a spelling.

**Known, not ours:** `metrics/ai_assistant_activity` fails here and on `main` alike (`silver.class_ai_assistant_usage` builds empty, before any `person_id` is involved) — filed separately, not a cutover regression.

## Coordination — #2082

@aleksdotbar — your visibility gate is merged into this branch (commits preserved), but its contract had to change: `entity.ids` is UUIDs now, so `/v1/visible-persons` takes `person_ids` and the email→person resolution step is gone. Visibility itself was always `person_id`-native inside identity, so the gate got *simpler*, not harder. Please look at how it landed and say which you prefer: **(a)** #2082 merges first as-is and this PR rebases to re-contract it, or **(b)** this PR supersedes it and #2082 closes with a pointer here.

## Rollout order

The gold `person_id` column is populated by a dbt build that reads a ClickHouse
mirror of the identity persons log. Deploy in this order:

1. **`persons-sync`** — populates `identity.identity_persons`. The CronJob runs
   daily at 06:45, and **nothing triggers an initial run**, so on a fresh stand
   run the Job by hand or every `person_id` resolves NULL until the next tick.
2. **dbt gold build** — resolves `person_id` into the observation and cohort
   relations. Until it runs, the old gold tables lack the column and the
   analytics schema validator reports `schema_status: error` for them.
3. **analytics + identity-resolution** — the services in this PR.

Between steps the failure mode is honest emptiness, not a 500: a metric whose
`person_id` is NULL contributes no rows.

### Tenant stamp caveat

Compiled queries filter on the warehouse `tenant_id`, bound from the caller's
gateway JWT. The platform has **no defined mapping** between the control-plane
tenant id and the `tenant_id` string stamped at ingestion. On a stand where the
two differ, every metric comes back empty with no log line saying why — a
configuration problem that reads exactly like "no data for this period".

Tracked in #2107 (deploy-time warning when the JWT tenant matches no stamp). The
predicate itself is not from this PR; it arrived with the metric-results
authorization work and is inherited here.

Refs #1873; builds on #2065, #2066; integrates #2082.

## Summary by CodeRabbit

* **New Features**
  * Added `POST /v1/visible-persons` to return person IDs visible to the authenticated caller.
  * Added profile and person lookup using canonical person UUIDs.
  * Analytics metric results now support UUID-based person filtering with visibility enforcement.

* **Bug Fixes**
  * Invalid, unknown, nil, or legacy email identifiers now return clear errors.
  * Improved tenant isolation and fail-closed authorization behavior.
  * Prevented empty synchronization results from replacing existing data unless explicitly forced.

* **Documentation**
  * Updated API specifications and guides with UUID requirements, visibility rules, validation constraints, and response details.


mozhaev-dev:
Paired frontend PR: **constructorfabric/insight-front#255** — merge the two together.

Merging either alone breaks the dashboards: this PR alone makes the API reject the emails the current SPA sends (400 on a non-UUID `entity_id`), and the frontend PR alone makes the SPA send UUIDs the current API resolves against nothing (empty metrics).

The frontend side also exercises the new `value_type: "person_id"` mode of `POST /v1/profiles` added here, and keeps legacy `/ic/<email>` URLs working by redirecting them through the email mode once.

mozhaev-dev:
### Note on the red `connectors-ddl snapshot + field parity` check

Not caused by this PR, and it needs a decision from whoever owns that gate (it is not a required check, so it does not block the merge).

The gate was merged today and has never been green. On `main` it fails at the **snapshot drift** step, because the `silver.contract_version` view landed with its `;` on the SELECT line while the dumper puts every terminator on its own line. This branch fixes that one line, which lets the job proceed to the next step — and that step, **Field parity**, was being skipped on `main` all along.

So the 13 failures now visible are pre-existing and were simply masked:

```
staging.m365__collab_document_activity_onedrive  visited_page_count  Nullable(Int64)  vs Nullable(Decimal(38,9))
staging.salesforce__crm_{accounts,activities,contacts,deals,users}  custom_fields  Nullable(String) vs String
staging.hubspot__crm_deals  close_date  Nullable(Date) vs Nullable(Date32)
staging.{github,gitlab}__pull_requests_commits  commit_order  UInt8 vs Int64
staging.{active_directory,ms_entra}__to_class_people  hire_date/termination_date  Nullable(Date) vs Nullable(DateTime)
```

None of these relations are touched here — this PR changes the metrics runtime and identity, not any connector staging model or class contract. Reverting the one-line snapshot fix would only re-mask them.

Everything else is green, including the required **Run E2E suite** and the **metrics** e2e job.

mozhaev-dev:
@coderabbitai Went through all findings from both reviews. Fixed in `8197f563`:

**Fixed**

- **Cap the ids before parsing** (`validation.rs`) — correct, and the second half of your point was the sharper one: the cap measured the *parsed* count, so blanks or duplicates could pad a request past the documented bound. It now counts submitted ids before parsing, with a test for each padding shape.
- **Prose → test, non-person entity types** — replaced the `DELIBERATE CONTRACT` comment with `an_entity_type_with_no_authorization_rule_fails_closed` plus `a_service_principal_bypasses_the_gate_for_any_entity_type`. Neither rule had a test before; that was the real gap.
- **`#[expect]` with a reason** — both new test modules (`person_visibility.rs`, `visible_persons.rs`).
- **Unwrap `authorization` at the guard** and **count the unmatched ids instead of collecting them** — both applied; the count is all that ever leaves the function anyway, since which ids were refused is not the caller's to learn.
- **Metrics DESIGN, Authorization section** — was worse than stale: it said entity-level scoping is deferred *and* that warehouse tenant isolation is absent. This PR lands the first, and the compiler already leads every observation/cohort read with `tenant_id = ?`. Rewritten, plus the nil-UUID rule and the submitted-vs-parsed cap semantics.
- **Cohort model description** (`gold/schema.yml`) — right, the email-grain uniqueness claim no longer explains anything; fan-out protection is the person-grain collapse in `compile_peer_batch_query`.
- **`maxItems` / `minItems` on `person_ids`** — added, matching the enforced bound.
- **Nil-UUID e2e case** and **status assertions** on the two visibility cases — added (identity suites: 23 passed; api metric-results: 9 passed).
- **Concrete row type** — `list[tuple]` → a named `ObservationRow` alias, applied to the pre-existing helper too.

**Not doing, with reasons**

- **`///` → `//` in service crates.** The guideline is real, but these two crates already carry ~3,200 doc-comment lines across ~130 files. Converting only my 34 new lines would make this code inconsistent with everything around it. That is a repo-wide sweep, not a PR-scoped fix.
- **Move the visibility decision / person resolution into the domain layer** (3 sites). No behaviour change, and it would restructure handlers that follow the existing layout of both services. Not while this PR is the blocking half of a two-repo lockstep merge.
- **Strip the 11 explanatory comments.** The behaviours listed are already pinned by named tests (`metric_results_forbids_a_person_outside_the_callers_visible_set`, `get_person_forwards_authorization_then_5xx_on_dead_identity`, the builder's wire-contract tests). What remains is *why*, which the guideline allows — and on a security gate it is the part a reviewer needs. I did remove the one comment that genuinely restated a rule better expressed as a test.
- **Worker-scope the identity database in the metric rig.** The proposed fix would break resolution: `resolve_person_id` names `identity.identity_persons` literally, so a per-worker suffix would leave gold reading an unseeded table. The suite is serial by design and documented as not xdist-safe for deeper reasons (shared analytics process, shared dbt `target/` dir). Added a comment at the seeding site so a future xdist attempt knows the macro has to become schema-aware first.
- **Typed errors for `person_exists` / the identity client.** Both callers map every failure to one canonical error; a typed error would add a layer with no reader.
- **Visibility on `POST /v1/metrics/{id}/query` and `POST /v1/metrics/queries`.** Correct and worth its own issue — but pre-existing, and already documented as a known gap in the analytics DESIGN ("Planned — IDOR prevention on org units"). Out of scope for a PR that must merge in lockstep with a frontend change.

mozhaev-dev:
## Codex review — both points addressed

**[P1] The analytics persons facade was still email-only — fixed in `9bf194f3`.**

Confirmed and it was the right call: `GET /v1/persons/{email}` sent the path value to identity as `value_type: "email"`, so a person id read off a metric result came back 404. Two contradicting contracts on one service.

Took the first option — converted the facade rather than removing it:

- Route is now `GET /v1/persons/{person_id}`, summary updated, OpenAPI regenerated.
- `IdentityClient::get_person` takes a `Uuid` and sends `{"value_type": "person_id", "value": "<uuid>"}`.
- A pre-cutover email path, or the nil UUID, is a **400** with a field violation — not a 404, which would read as "no such person". The parse runs **before** the `is_configured` check, so that client error does not depend on deployment state.
- The e2e identity stub resolves both key forms, so all three outcomes are observed: `test_persons.py` now covers 200 / 404 / 400-on-email / 400-on-nil. The endpoint-coverage ledger drops 400 from the boilerplate set for this route, since it is reachable now.

Removal was the alternative — the new SPA already calls `POST /v1/profiles` directly, so the facade has no in-repo consumer. Kept it because deleting a published route silently breaks anything outside the repo, and converting is reversible. Worth revisiting once we can confirm nothing external calls it.

Evidence: api suite **96 passed, 14 xfailed** (was 93 + 3 failing on the stale runner image — rebuild needed after the Rust change), endpoint-coverage gate green with only the pre-existing `metric-definitions` advisory, `cargo clippy -D warnings` clean, 526 analytics + 89 identity unit tests, OpenAPI drift check OK.

**[P2] identity README — fixed.** The `POST /v1/profiles` row now spells out all three modes: `email` (tenant-wide), `id` (source-native, needs both source fields), `person_id` (canonical UUID).

One related thing that is **intentional**, so nobody removes it: the identity `openapi.json` still lists the retired `GET /v1/persons/{email}`. The endpoint-coverage gate uses the committed spec as its universe and carries a SKIP entry for that op ("dropped in the Rust successor"). Pulling the path out of the spec would strand the skip.

aleksdotbar (changes requested):
## Keep the canonical ID in `entity_id`

The metrics contract is explicitly polymorphic: `entity_type + entity_id`. After this cutover, the canonical person UUID is the entity identity, so it should replace the email value in `entity_id`.

The current shape instead keeps:

- `entity_id` = source-native email
- `person_id` = canonical UUID and actual runtime key

This creates two competing identity columns. `entity_id` no longer identifies the measured entity despite its name and documented role, while the entire runtime is specialized around `person_id`.

This causes concrete inconsistencies throughout the stack:

- All five observation tables remain ordered by the obsolete email key:
  - `src/ingestion/gold/ai_metric_observations.sql`
  - `src/ingestion/gold/collab_metric_observations.sql`
  - `src/ingestion/gold/git_metric_observations.sql`
  - `src/ingestion/gold/task_metric_observations.sql`
  - `src/ingestion/gold/wiki_metric_observations.sql`

  Their `ORDER BY` is still `(source_key, measure_key, entity_id, metric_date)`, while runtime queries filter and group by `person_id`. Because `entity_id` sits before `metric_date`, the old email key also prevents effective pruning by the canonical person and subsequent date component.

- The gold models group by both `entity_id` and `person_id`. Therefore, multiple emails belonging to one canonical person remain separate rows in the serving tables and must be recombined by every runtime query. The gold observation grain remains source-email grain rather than canonical-entity grain.

- `metric_entity_cohorts_current` also remains email-grained. The peer compiler now needs additional grouping and conflict-handling CTEs to collapse those rows back to person grain. This is compensating in the query layer for a non-canonical serving model.

- `OBSERVATION_COLUMNS` and `COHORT_COLUMNS` now require both `entity_id` and `person_id`, making the duplicated identity shape part of the permanent runtime contract.

- `gold/schema.yml` describes `entity_id` as a lowercased email while declaring `person_id` to be the actual runtime key. This contradicts the metrics design statement that `entity_type + entity_id` identifies the measured entity.

- Validation, compiler rows, batching, builders, and query results were renamed from `entity_id` to `person_id`, only for the builders to translate `person_id` back into the wire field `entity_id`. That boundary translation is a symptom that the internal model no longer matches the public entity abstraction.

- Comments in the gold models still state that `entity_id` is the runtime key, which is no longer true after this PR.

Source email may still be required for resolution coverage and provenance, but it should be represented explicitly as `source_entity_id` or retained in an upstream resolution model. `identity_resolution_coverage` can count unresolved source identifiers from that layer; it does not require the source email to remain the canonical gold `entity_id`.

The intended serving shape should be:

- `entity_type = 'person'`
- `entity_id = <canonical person UUID>`
- optional `source_entity_id = <source-native email>` in the resolution/provenance layer

The runtime can then continue using `entity_id` throughout, the observation tables can be materialized at canonical entity grain, the sort key remains aligned with runtime access patterns, and future entity types fit the existing abstraction without another domain-wide rename.

Please implement the final canonical shape in this cutover rather than shipping `person_id` as a second permanent identity column and adapting every downstream layer around that duplication.

mozhaev-dev:
## Backend review — `c19a8534`

**[MAJOR] IDOR via the legacy query endpoints — closed in the branch, not deferred.**

Confirmed: `POST /v1/metrics/{id}/query` and `POST /v1/metrics/queries` bound `person_id eq …` / `person_id in (…)` from `$filter` straight into the query, with the MVP TODO still sitting there. Since the cutover declares entity-level scoping enforced, documenting the hole was not good enough — the ids now go through the same `authorize_entity_ids`, same fail-closed matrix, before any ClickHouse work.

Two details worth naming:

- A `person_id` filter value that is **not** a person UUID is now a **400**. Post-cutover the column is a UUID, so a pre-cutover email would have reached ClickHouse as a cast failure (a 500 the caller cannot act on) and it names nobody the gate could check either.
- `org_unit_id` stays ungated on purpose: no visibility rule for org units exists, and failing those closed would take out every department distribution metric. Filed **#2108** with what the fix actually needs (a visibility predicate for org units first), and the DESIGN now states which half is enforced and which is planned instead of covering both with one "Planned" paragraph.

Low breakage risk: the new SPA does not call these endpoints at all, and the gate only refuses ids the caller may not see.

**[MAJOR, rollout] Tenant predicate against an undefined mapping.** Agreed, and I did not touch the predicate — it arrived with the metric-results authorization work and this branch only merges it into the person-grain CTEs. Two actions:

- **#2107** for the deploy-time check, with the contract-version sweep named as the home (it already polls ClickHouse on an interval), and a warning rather than a hard failure so a stand mid-backfill still boots.
- A **Tenant stamp caveat** section in the PR description: on a stand where the JWT tenant and the ingestion stamp differ, every metric is empty with no log line saying why, and that is configuration, not data.

Your evidence is the strongest part of that finding — the e2e fixtures in this very branch had to be moved off the nil tenant onto the rig's JWT tenant by hand. They do not line up on their own.

**[minor] Deploy order.** Added to the PR description as a numbered **Rollout order**: `persons-sync` → dbt gold build → services, including the trap you spotted — the CronJob is daily at 06:45 and **nothing triggers an initial run**, so a fresh stand needs the Job run by hand or `person_id` resolves NULL for up to a day. Also stated that between steps the failure mode is honest emptiness, not a 500, and that old gold tables report `schema_status: error` until the dbt build lands.

**[minor] TOCTOU in the sync guard.** Real, and the same class as #1550 — fixed. Rather than re-checking `rows.len()` in the runner, the guard moved into `run_sync`, where it judges the rows about to be published; a seed emptying the log mid-run can no longer slip past a stale `count()`. The now-unused `count()` is gone, and the guard tests moved with it — they exercise the real path now, including the property that actually matters: **a refusal writes nothing**, so the populated snapshot survives.

**[minor] Two `MAX_PERSON_IDS`.** Each constant now says it must equal the other and why: analytics forwards a whole cleared request to identity, so a smaller cap there would reject what analytics already accepted. A shared-contract test would need a crate both can see; the comments are the honest cheap version.

On the error text: analytics says "entity ids" because its wire field **is** `entity.ids`, and identity says "person ids" because its field is `person_ids`. Each message names the field the caller sent, which is what a client-facing message should do — leaving as is.

**[nit] `entity_id` carrying UUIDs.** Contract is frozen and documented, and the free moment has passed — a rename now would mean re-cutting the frontend PR and re-running both suites for a naming change. Leaving it.

## Verification

`cargo fmt` · clippy `-D warnings` clean · **530 analytics + 87 identity** unit tests · e2e `api/` + `identity/` **237 passed, 3 skipped, 14 xfailed** (rebuild the runner image after Rust changes — a stale baked binary silently tests the old code) · OpenAPI drift check OK · endpoint-coverage gate green.

## Frontend

Everything in your frontend section was already fixed before this message landed — `f0da25d` and the two commits before it. nil-UUID guard in `isPersonId` (mocks inherit it), the redirect sending the normalized email, `findIdentityNodeByEmail` deleted, the dashboard's "person not available" state, and the mock email casing plus its 400-vs-404 split. Details in [insight-front#255](https://github.com/constructorfabric/insight-front/pull/255) — 441 unit tests, 100% diff coverage.

mozhaev-dev:
## Heads-up from merging `main` — the new drilldown endpoint needs the same treatment (#2109)

`main` landed `POST /v1/metric-drilldown` (#2073) while this branch was in review. Merged in `3c3f39e4`; everything compiles and passes, but the new endpoint is now the odd one out on two counts, and one of them is a security gap:

1. **It is email-keyed.** `validate_request` lowercases `entity.id` and the compiler filters `evidence.entity_id`. Every other person-keyed route takes a canonical person UUID after this PR, and the SPA holds UUIDs — so its first frontend caller would get a silently empty drilldown. Cutting it over is not a one-liner: this PR added `person_id` to the **observation** and **cohort** relations, not to the five `*_metric_evidence` relations the drilldown reads.

2. **It has no visibility gate.** No `authorize_entity_ids` on the route, and it serves per-person evidence rows for whatever entity id the caller names — the same IDOR class `/v1/metric-results` answers 403 for and the one the legacy `$filter` paths were closed for here.

Neither is caused by this PR and neither regresses anything today (the endpoint has no frontend caller yet), so I did not expand this PR to fix it — filed **#2109** with the ordering: `person_id` into evidence → validator/compiler → gate → e2e 403 case.

What the merge did do is keep the seam visible: the drilldown used the shared `normalize_entity_id`, which this PR deletes (its email semantics are what the cutover removes). Rather than resurrect it, the drilldown now has a local `normalize_evidence_entity_id` with a comment pointing at #2109 — so the email keying reads as a known, tracked gap instead of looking like the cutover missed a call site. `normalize_key` keeps `main`'s widened visibility, which the drilldown needs.

**Verification after the merge:** clippy `-D warnings` clean · **540 analytics + 87 identity** unit tests · e2e `api/` + `identity/` **239 passed, 3 skipped, 14 xfailed** · OpenAPI doc regenerated (the drilldown paths are in it now) and drift-check OK.

mozhaev-dev:
@coderabbitai The one actionable finding and all three nitpicks are on `metric_drilldown/validation.rs`, which is `main`'s code (#2073) — this PR only added a three-line local helper there while merging. Taking one of them; the rest belong with that endpoint's owner.

**Taken — issue number out of the source comment** (`8d604ab4`). Correct against our own rule: `src/backend/CLAUDE.md` says no issue numbers in source, that context lives in the issue and the PR. The comment now states the invariant a future edit could actually break — this endpoint reads email-keyed evidence, so it must **not** be handed the person UUID every other route takes — and the tracking stays in #2109 where I put it.

**Not taken**

- **Reject oversized fields before normalizing.** The premise that the allocation is unbounded does not hold: these handlers take `Json<…>`, and axum 0.8 applies `DefaultBodyLimit` (2 MB) unless it is explicitly disabled, which nothing here does. So the worst case is bounded at parse time, before any of this code runs, and reordering the checks buys a fraction of one already-bounded body. A tighter per-field cap may well be worth having, but it is a change to the drilldown's contract (new 400s on inputs it accepts today) and belongs with #2109, not with a merge commit.
- **Separate pure validation from external I/O** and **table-driven normalization test** — both are refactors of code this PR does not own, with no behaviour change. Reshaping `main`'s new endpoint from inside the blocking half of a two-repo lockstep merge is how a merge turns into a review round it did not need.

For the record on that file: the email keying and the missing visibility gate on `POST /v1/metric-drilldown` are both tracked in **#2109**, with the ordering — `person_id` into the evidence relations, then validator/compiler, then the gate, then an e2e 403 case.

cyberantonz (approved):

aleksdotbar (changes requested):
## Finish the canonical `entity_id` cutover

The gold observation changes address the original storage concern:

- `entity_id` now contains the canonical person UUID.
- `person_id` is no longer a second observation/cohort column.
- Source aliases collapse to canonical entity grain.
- Observation sort keys and runtime SQL use `entity_id`.

However, the cutover remains incomplete in the runtime and evidence paths.

### 1. The runtime still treats `person_id` as the real domain key

`ValidatedMetricResultsRequest` still combines:

- `entity_type: String`
- `person_ids: Vec<Uuid>`

Validation parses `entity.ids` into `person_ids` for every entity type, compilers bind `req.person_ids` into `entity_id` predicates, and builders use `values_by_person` / `by_person` before translating back to response `entity_id`.

The builder comment explicitly calls `entity_id` a “pre-cutover” wire name. That is incorrect after this change: `entity_type + entity_id` is the canonical contract, and for a person the UUID is the `entity_id`. There is no compatibility translation at the response boundary.

Please model this as a typed entity selection, for example:

`ValidatedEntitySelection::Person { ids: Vec<Uuid> }`

The compiler and builders can then operate through the canonical entity abstraction while retaining UUID type safety. Do not replace UUIDs with unvalidated strings; make the entity specialization explicit in the type instead of pairing a generic `entity_type` string with person-specific fields.

This request does not apply to legitimate `person_id` uses in the identity service, visibility API, or frozen legacy metric schemas.

### 2. Drilldown reverses the canonical identity back into current emails

The drilldown path currently does:

`canonical entity.id → person_id → current source emails → evidence.entity_id`

This undoes the cutover at read time. Observations were attributed using the identity map from the gold build, while drilldown uses the current `identity.identity_persons` mapping. If an email is reassigned between those points, the drilldown can omit evidence included in the selected metric or show evidence that was not included.

Calling the evidence advisory does not resolve that contradiction: drilldown exists to explain the selected metric and must use the same attribution.

The compiler also duplicates the `resolve_person_id()` algorithm in Rust despite the dbt macro claiming to be the single resolution point. Tenant scoping, source-specific mappings, or future as-of semantics can consequently diverge between the two implementations.

The cursor only snapshots the evidence table UUID. It does not snapshot the identity mapping, so identity membership can change between pages without invalidating the cursor.

Please persist one build-scoped identity-resolution snapshot and use it consistently for:

- resolved evidence;
- observations;
- cohorts;
- resolution coverage;
- drilldown.

Resolved evidence should expose the canonical UUID as `entity_id` and retain the source-native email explicitly as `source_entity_id` for provenance and coverage. Drilldown can then filter `evidence.entity_id = ?` directly, without reverse-resolving through the live identity log. The evidence query must also follow the runtime’s tenant-scope policy rather than having no tenant predicate.

### 3. Cohorts and observations currently use different identity snapshots

Observation relations are materialized at gold-build time, but `metric_entity_cohorts_current` is a live view that resolves identity on each peer request.

After an identity sync and before the next gold build, cohort `entity_id` values can therefore disagree with observation `entity_id` values. That can produce missing targets, missing target values, or incorrect peer membership.

The cohort relation should depend on the same persisted resolution snapshot as the observations, not the current identity log.

### 4. Existing dbt integrity tests still enforce the email contract

These tests still reject any `entity_id` that does not contain `@`:

- `assert_collab_observations_entity_id_shape.sql`
- `assert_task_observations_entity_id_shape.sql`
- `assert_wiki_observations_entity_id_shape.sql`

Every valid canonical UUID now violates those predicates. The Git equivalent still documents email semantics and no longer verifies the canonical UUID shape.

Please update the observation identity checks to validate canonical UUID `entity_id` values across the managed relations.

### Cleanup

- Remove the unused `resolved_person_id_column()` macro and its obsolete example.
- Correct `schema.yml`, which describes `identity_resolution_coverage` as a view although it is materialized as a table.
- Remove comments that describe canonical `entity_id` as a legacy or pre-cutover field.

The storage direction is now correct. These remaining changes are needed so the runtime, evidence, peer, pagination, and validation contracts all follow the same canonical identity model rather than translating between `entity_id`, `person_id`, and current email mappings at different boundaries.

mozhaev-dev:
## Finish-the-cutover review — all four done (`409b94ca`, `6eb954ef`)

**1. Typed entity selection.** `ValidatedEntitySelection::Person { ids: Vec<Uuid> }` replaces the `entity_type: String` + `person_ids` pair. The specialization is in the type, ids keep UUID safety, compilers/batching/builders speak `entity_id` end to end, `by_person` naming and the "pre-cutover wire name" comment are gone — you were right that the field's name *is* the canonical contract, not a compatibility seam. A first non-person type now adds a variant plus its own gate rule instead of inheriting person semantics.

**2 + 3. One build-scoped resolution snapshot.** Conceded in full — read-time reverse-resolution was undoing the cutover, and "advisory" was not an answer to drilldown disagreeing with the metric it explains. Resolution now happens exactly once, in the evidence wrappers: `entity_id` = canonical UUID ('' when unresolved — those rows stay for coverage and reach no serving relation), `source_entity_id` = the source email for provenance. On the same snapshot, in the same build:

- observations read already-resolved evidence — the resolve join is gone;
- `metric_entity_cohorts_current` is a **table**, so peer membership cannot disagree with observations after a sync;
- coverage counts unresolved rows straight off evidence, no join;
- drilldown binds `evidence.entity_id = ?` — the Rust copy of the resolve algorithm is **deleted**, the dbt macro is the single resolution point again, the existing evidence-snapshot check now covers identity membership for the cursor, and the evidence read follows the runtime tenant-scope policy.

One consequence your review implied but did not name: with `entity_id` canonical, several source-day `record_id`s (date:measure:dims-hash) became identical across one person's accounts, breaking the evidence uniqueness grain and cursor ordering. `record_id` is now account-qualified in the wrapper (hashed — the id reaches the client and stays opaque).

**4. Shape asserts.** The sharpest catch of the round — every canonical UUID violated the '@' predicates, and it was latent for two mundane reasons: the e2e rig runs gold via `dbt run` (no tests) and bootstrap builds gold empty (asserts pass on zero rows). All five relations (ai had no assert at all) now validate the canonical UUID shape, and I verified them **against built data**, not an empty stand.

**Cleanup:** `resolved_person_id_column()` deleted, `schema.yml` corrected (coverage is a table; `source_entity_id` documented), the stale "pre-cutover" description removed.

**Verification** — unit 545 + 88 · live MariaDB `--include-ignored` 601 + 88 · e2e api+identity+metrics **277 passed** (1 failed = the known pre-existing `ai_assistant_activity`) · served data inspected directly: all five observation relations pass the UUID shape check with data, evidence carries both keys (alias fixture: two source emails under one canonical id), evidence grain has zero duplicates, coverage reads the build snapshot, cohorts is a MergeTree table · OpenAPI drift OK.

aleksdotbar (approved):

Issue #1981, opened 2026-07-28
[EPIC] Self-service preview environments (CI-driven provisioning)


cyberantonz:
Automate preview-experiment provisioning so a non-technical user can spin up a branch's frontend on a stand and tear it down — self-service, no helm/kubectl/git.

Standalone epic (split out of #1803).

## Background
#1855 (Envoy Gateway) unblocked this: each preview experiment is now a Gateway API `HTTPRoute` (chart at `deploy/preview/`) that attaches to the shared Gateway over xDS — creation is a plain Kubernetes API operation with no proxy reload and no central config rewrite. The manual flow (`helm upgrade --install preview-<name>` / `helm uninstall`) works today; this epic turns it into a managed service with a UI.

## Approach
A **`previews` backend service** (a gear, in gears-rust terms — `#[toolkit::gear]` host like `identity-resolution`) manages the per-experiment `Deployment` + `Service` + `HTTPRoute` trio via kube-rs. **Kubernetes is the source of truth** — the objects *are* the experiment, metadata as annotations; no database. It sits behind the gateway (gateway-JWT via the `oidc-authn-plugin` gear) and is gated on the authenticator's `experiments_enabled`, so it is structurally inert on a production stand. A `/previews` section in the existing SPA drives it.

Guardrails (a cluster-mutating service, boxed by construction): dedicated `insight-previews` namespace, RBAC scoped to Deployment/Service/HTTPRoute in that namespace only, tag-only image input (repo hardcoded to the frontend image), fixed pod shape, DNS-1123 name validation, a live-count cap, and a TTL sweep.

## Sub-tasks
- [x] Phase -1 — FE serves under any /exp/<name> prefix (relative base + runtime basepath): #2402
- [x] Phase 1 — `previews` gear, kube-rs experiment API: #2372
- [x] Phase 2 — deploy wiring (chart, namespace-scoped RBAC, gateway route): #2373
- [x] Phase 3 — frontend self-service UI + capability gate: #2374
- [x] Phase 4 — polish (registry tag listing, TTL/cap surfacing): #2375

Phase 0 (flip `experiments_enabled` + hand-provision one experiment end-to-end to de-risk #1972/#1973 against the Envoy edge) needs no separate issue — tracked here.

## Prerequisite already met
Branch-tagged `preview-*` frontend image builds exist.

Refs #1855, #1803, #1971, #1972, #1973, #1994.

cyberantonz:
[...]

Issue #1973, opened 2026-07-28
[pres] Preview backend pinned to synthetic data


cyberantonz:
Phase A. Preview stack reads synthetic/dev only, never customer prod (R1.4). Parent: #1803.

Issue #1972, opened 2026-07-28
[pres] Preview auth: Redis-state return path


cyberantonz:
Phase A. Redis-backed opaque `state` over one fixed Entra callback. Parent: #1803.

PR #2123, opened 2026-08-03
feat(preview): path-based per-experiment FE preview bundle (#1971)

### LINKED ISSUE #1971: [pres] Preview environments (path-based, manual)
Phase A. `/exp/<name>` single host, one route object per experiment, manual kubectl/helm. Parent: #1803.

cyberantonz:
## What

Adds `insight-preview`, a small standalone Helm chart at `deploy/preview/` that provisions one FE preview experiment per release: a `Deployment` + `Service` + a single prefix-strip `Ingress` served under `/exp/<name>` on one shared host.

Provisioning is manual (no GitOps controller): `helm upgrade --install preview-<name> ...` adds the `/exp/<name>` path and `helm uninstall preview-<name>` removes it. The controller merges same-host route objects, so no central config is ever rewritten. Only the FE image varies per experiment; the backend never does.

## Design points

- **Prefix-strip route** — `rewrite-target: /$2`, `path: /exp/<name>(/|$)(.*)`, `pathType: ImplementationSpecific`, mirroring the existing fakeidp chart. `/api/...` is emitted unprefixed by the FE and is not matched here, so it flows to the shared backend route.
- **Controller-agnostic** — nginx `Ingress` now; becomes a Gateway API `HTTPRoute` after the Envoy move (#1981).
- **Guards at template time** — the experiment slug must be a DNS-1123 label and at most 55 chars (so `preview-<name>` never truncate-collides at the 63-char limit); `ingress.host` and `image.tag` are required; OIDC and the dev-impersonation escape hatch are mutually exclusive.

## Tests

- `deploy/preview/tests/test_render.py` — 14 render-contract assertions (prefix-strip route, per-experiment naming, service→pod selector, and every guard).
- `.github/workflows/preview-helm.yml` — `helm lint` + the pytest suite on `deploy/preview/**`.
- Extended the `yamlfmt` pre-commit exclude to cover `deploy/*/templates` (Go-templated, not parseable YAML).

## Specs

Extended the presentation PRD and DESIGN to record the shipped serving-path slice. The auth return path (#1972) and the synthetic-data pin (#1973) stay open, so the `cpt-presentation-fr-preview-envs` FR and the `cpt-presentation-component-preview-router` component remain unchecked.

Closes #1971
Part of #1803

## Summary by CodeRabbit

* **New Features**
  * Added Helm-based preview environments for individual experiments.
  * Preview applications are available under shared-host paths such as `/exp/<name>`.
  * Added configurable replicas, images, resources, services, and ingress routing.
  * Added validation for experiment names, required hosts, and image tags.
  * Preview authentication is handled externally by the gateway.

* **Documentation**
  * Documented preview deployment, routing, naming, URLs, authentication limitations, and cleanup procedures.

* **Tests**
  * Added automated checks for rendering, routing, resource configuration, and invalid settings.


cyberantonz:
Addressed the CodeRabbit review in a4a78de:

- **use-regex** (ingress) — added `nginx.ingress.kubernetes.io/use-regex: "true"` so the capture-group prefix-strip path matches as a regex; test asserts it.
- **image digest** — the template renders `repository:tag`, invalid for a digest, so documented `image.tag` as a build tag only (values + README).
- **experiment length** — the ≤55-char guard was already in the helper; added the limit to the values/README/DESIGN docs.
- **CI checkout** — set `persist-credentials: false` (repo convention).
- **MD018** — reworded the README intro so no line starts with `#1971`.
- **DESIGN wording** — "controller-agnostic" → "nginx-specific today, portable by intent".

**Not actioned (moot):** the note to document the full `devUserEmail`↔`oidc.{issuer,clientId}` exclusivity — that block was removed from the preview chart earlier in this PR (auth is the gateway+authenticator's job, wired in #1972), so there is nothing left to document.

ktursunov (approved):

PR #2200, opened 2026-08-04
refactor(analytics)!: retire the legacy API surface

### LINKED ISSUE #1663: backend: legacy per-metric thresholds 500 on every read once a row exists (DECIMAL value vs f64 entity)
## Problem

The legacy per-metric thresholds API (`/v1/metrics/{id}/thresholds*`) is unusable: the moment the first threshold row is created, **every read of the table returns a canonical 500** — the create's own read-back, list, update, and delete. The INSERT itself succeeds, so one successful-looking create permanently poisons the endpoint for that tenant. This is an analytics-api entity/schema mismatch, not a UI or data-pipeline issue.

## Context

`thresholds` is the legacy per-metric threshold store (predates the `metric_threshold` admin table). Its `value` column is `DECIMAL(20,6)`, but the SeaORM entity maps it as a bare `f64` — and sqlx-mysql cannot decode MySQL `NEWDECIMAL` into `f64`. Writes coerce fine server-side; reads fail decoding.

## Steps to reproduce

Fastest isolated check (any deployed instance or the e2e rig):

```bash
# 1. create a metric (201) — $API is the analytics base URL, $T a tenant id
MID=$(curl -s -X POST "$API/v1/metrics" -H "X-Insight-Tenant-Id: $T" -H 'Content-Type: application/json' \
  -d '{"name":"thr-repro","description":"x","query_ref":"SELECT 1 AS one FROM system.one"}' | jq -r .id)
# 2. create a threshold → HTTP 500 (the INSERT lands; the read-back fails)
curl -si -X POST "$API/v1/metrics/$MID/thresholds" -H "X-Insight-Tenant-Id: $T" -H 'Content-Type: application/json' \
  -d '{"field_name":"one","operator":"ge","value":1.0,"level":"good"}' | head -1
# 3. every subsequent read of the endpoint also 500s
curl -si "$API/v1/metrics/$MID/thresholds" -H "X-Insight-Tenant-Id: $T" | head -1
```

**Expected:** 201 with the created threshold; 200 with `{items:[…]}`.
**Actual:** 500 `application/problem+json` (`cf.core.err.internal.v1~`) on both.

## Evidence

- Server log (the decode failure, from a live reproduction):

  ```
  analytics::api::handlers: failed to list thresholds error=Query Error: error occurred
  while decoding column "value": mismatched types; Rust type `core::option::Option<f64>`
  (as SQL type `DOUBLE`) is not compatible with SQL type `DECIMAL`
  ```

- The row IS inserted despite the 500 (`SELECT HEX(id), field_name, operator, value FROM thresholds` returns the created row with `value = 1.000000`), confirming write-path OK / read-path broken.
- Found by the e2e endpoint contract suite in #1661 (`api/test_metric_thresholds.py`, currently xfail'd against this issue).

## Root cause

- Column: [`m20260414_000001_init.rs#L86`](https://github.com/constructorfabric/insight/blob/main/src/backend/services/analytics/src/migration/m20260414_000001_init.rs#L86) — `Thresholds::Value` is `decimal_len(20, 6)`.
- Entity: [`entities.rs#L38`](https://github.com/constructorfabric/insight/blob/main/src/backend/services/analytics/src/infra/db/entities.rs#L38) — `#[sea_orm(column_type = "Decimal(Some((20, 6)))")] pub value: f64`.
- Every entity read decodes `value` and hits the sqlx type mismatch; the handlers wrap it as internal: [`handlers.rs#L1054`](https://github.com/constructorfabric/insight/blob/main/src/backend/services/analytics/src/api/handlers.rs#L1054) (create read-back), [`handlers.rs#L992`](https://github.com/constructorfabric/insight/blob/main/src/backend/services/analytics/src/api/handlers.rs#L992) (list).

The admin table documents the same pitfall and avoids it with raw-SQL casts ([`repository.rs#L4`](https://github.com/constructorfabric/insight/blob/main/src/backend/services/analytics/src/domain/admin_threshold/repository.rs#L4): "All SELECTs use raw SQL with `CAST(... AS DOUBLE)` for the DECIMAL"). Note SeaORM's `select_as` attribute is NOT a fix on MySQL/MariaDB — the cast expression is not re-aliased, so decoding then fails with `no column found for name: value` (verified). The API is `f64` end-to-end (`CreateThresholdRequest.value` in, `Threshold.value` out), so the DECIMAL column adds only a silent 6-dp quantization between two `f64` endpoints.

## Other notes

`metric_threshold` (admin) is unaffected — different read path. A DECIMAL→DOUBLE migration was validated end-to-end on the #1661 branch before being scoped out of that PR (test-only); the e2e lifecycle passed with it.

### LINKED ISSUE #1664: backend: duplicate admin threshold create returns 500 internal + drift alarm instead of declared 409
## Problem

Creating a duplicate admin threshold (same `metric_id` + tenant scope target) returns a **canonical 500 internal error** and fires the "schema/mapper drift" alarm — for what is a routine client conflict. The OpenAPI spec declares `409` for this operation; clients get a misleading "internal error, retry later" (retrying can never succeed), and the drift alarm is a false positive that trains operators to ignore it. analytics-api error-mapping bug.

## Steps to reproduce

Fastest isolated check (any deployed instance or the e2e rig):

```bash
# $MID must be a metric_catalog row id (e.g. from POST /v1/catalog/get_metrics)
BODY='{"metric_id":"'$MID'","scope":"tenant","good":0.0,"warn":0.0}'
curl -si -X POST "$API/v1/admin/metric-thresholds" -H "X-Insight-Tenant-Id: $T" -H 'Content-Type: application/json' -d "$BODY" | head -1   # 201
curl -si -X POST "$API/v1/admin/metric-thresholds" -H "X-Insight-Tenant-Id: $T" -H 'Content-Type: application/json' -d "$BODY" | head -1   # 500 (expected 409)
```

**Expected:** `409` conflict (`already_exists`) — the code every admin operation already declares in `docs/components/backend/analytics/openapi.json`.
**Actual:** `500` — `{"type":"gts://gts.cf.core.errors.err.v1~cf.core.err.internal.v1~","title":"Internal","status":500,"detail":"An internal error occurred. Please retry later.",…}`

## Evidence

- Wire capture from a live reproduction: first POST → 201; byte-identical second POST → the 500 envelope above.
- Server log on the duplicate: `admin-crud: DbErr with no matching CHECK mapper — schema/mapper drift` — the alarm that is supposed to mean "schema and mapper have drifted" fires on a normal duplicate create.
- Found by the e2e endpoint contract suite in #1661 (`api/test_admin_thresholds.py::test_create_409_duplicate`, currently xfail'd against this issue).

## Root cause

The duplicate violates the UNIQUE composite [`uq_metric_threshold_scope_target` (`m20260522_000002_metric_threshold.rs#L210`)](https://github.com/constructorfabric/insight/blob/main/src/backend/services/analytics/src/migration/m20260522_000002_metric_threshold.rs#L210). `map_db_err` matches only the five named CHECK constraints; a UNIQUE violation (MariaDB 1062) matches none and falls through to the internal-500 fallback with the drift alarm: [`error_map.rs#L132`](https://github.com/constructorfabric/insight/blob/main/src/backend/services/analytics/src/api/admin/error_map.rs#L132). Mapping-side, not schema-side: the constraint is intended (the DESIGN's per-scope-target uniqueness); only the error translation is missing.

## Other notes

A mapping of the `uq_metric_threshold_scope_target` substring to a 409 `already_exists` envelope was validated end-to-end on the #1661 branch before being scoped out of that PR (test-only).

### LINKED ISSUE #1670: backend(analytics): legacy body endpoints return non-canonical errors + 422 (plain axum::Json, not CanonicalJson)
## Summary

Six legacy write/query handlers in `src/backend/services/analytics/src/api/handlers.rs`
extract the request body with plain `axum::Json<T>` instead of the in-repo `CanonicalJson<T>`
extractor (`src/backend/services/analytics/src/api/canonical_json.rs`):

- `POST /v1/metrics`
- `PUT  /v1/metrics/{id}`
- `POST /v1/metrics/{id}/query`
- `POST /v1/metrics/queries`
- `POST /v1/metrics/{id}/thresholds`
- `PUT  /v1/metrics/{id}/thresholds/{tid}`

Because of that, on a bad body they answer with **Axum's default plain-text envelope**
(not RFC-9457 `application/problem+json`) and with inconsistent status codes:

- **415** — missing/wrong `Content-Type`
- **400** — malformed JSON (`JsonSyntaxError`)
- **422** — well-formed JSON that fails the schema/type (`JsonDataError`)

The admin-CRUD and catalog endpoints already use `CanonicalJson`, which emits a canonical
Problem body and maps the schema case to **400** (never 422). So the service is inconsistent
with itself in both **status** (422 vs 400 for an off-schema body) and **envelope**
(plain-text vs RFC-9457 Problem). DNA `REST/API.md §7` requires the canonical Problem
envelope on every error.

## Related (same root cause, path params)

The `{id}` / `{tid}` segments bind `Path<Uuid>`, whose deserialization failure
(`FailedToDeserializePathParams`) is also a non-canonical plain-text **400** on every route —
there is no canonical Path extractor.

## Suggested fix

Point the six handlers at the existing `CanonicalJson<T>` extractor — it is a drop-in; the
module already exists and is used elsewhere. That collapses the off-schema case 422 → canonical
400 and unifies the envelope. Optionally add a canonical `Path` extractor so path-param
rejections also carry the Problem envelope.

## Tests

`src/ingestion/tests/e2e/api/` already pins the *intended* 400 for the off-schema case with
strict-xfail cases named `test_*_400_schema_mismatch` (metrics create/update/query/batch,
threshold create/update). They flip green automatically once the extractor is unified — no
test change needed to close this bug.

aleksdotbar:
## What

Analytics served two generations of API. Old one has no consumer and is broken. Removed it.

30 → 10 operations.

**Kept** — what the engineering/presentation split needs:

`GET|POST /v1/queries` · `GET|PUT|DELETE /v1/queries/{id}` · `POST /v1/queries/{id}/run` · `POST /v1/metric-results` · `GET /v1/metric-definitions` · `POST /v1/metric-drilldown` · `POST /v1/metric-drilldown/export`

**Removed:**

| Group | Ops |
|---|---|
| legacy metric CRUD + OData query engine | 7 |
| legacy per-metric thresholds | 4 |
| admin metric-thresholds | 5 |
| catalog read (`POST /v1/catalog/get_metrics`) | 1 |
| column catalog | 2 |
| person passthrough | 1 |

−37k LOC.

## Why safe

- Frontend calls none of them. Its catalog client layer was already deleted; it uses metric-definitions, metric-results and `/v1/queries*` only.
- `/v1/columns*` is not schema discovery for query authoring — that reads ClickHouse `system.columns`. Both ops and `table_columns` date from the first analytics commit, untouched by any feature since. Table ships empty, no write endpoint, no seed.
- Admin threshold CRUD had no client. Every `metric_threshold` row comes from a seed migration; the admin UI it was built for was never written.
- `metric_catalog` has no surviving reader. `metric_definitions` is self-seeding and owns its own tables.
- `GET /v1/persons/{person_id}` orphaned. The visibility gate uses `visible_person_ids`, a different call.

## Fallout

- `infra/cache/` goes — the catalog cache was analytics' only Redis consumer. Redis itself stays wired: the service, `redis_url`, the chart/compose/secret plumbing, the crate dep and the CI container are all intact, because Redis is a platform staple and the next caching feature should write code, not plumbing. Only the `infra::cache::live_tests` filter goes, since it names a module that no longer exists.
- Boot no longer gates on a `metric_threshold` floor. `metric_catalog` integrity is still covered by the CHECK probe and the metric-definition validator.
- `schema_validator` reduces to its metric-key parser, the one piece metric-results validation still needs.
- Config key gone: `metric_catalog.tenant_default_id` — it only fed the tenant auth trait. `metric_catalog.enforce_tenant_scope` stays; metric-results and drilldown read it.
- OpenAPI regenerated from the route table: 19 → 7 paths, 30 → 10 ops, 81 → 51 schemas.
- e2e rig: the session fixture seeded the legacy `metrics` table for every test, so `lib/metric_seed.py` + `seed/metrics.yaml` go with the call site. Gateway downstream-verify probes `/v1/metric-definitions`.
- stand suite: its analytics half asserted the removed endpoints. `test_metrics`, `test_thresholds`, `test_catalog` and `test_columns` go; `operations.py` drops 30 ops to 10; `coverage.py`'s BLOCKED keeps only surviving routes; `schemas/analytics.py` regenerated from the spec. The seed stops writing `table_columns`, so the manifest, its model and `PROFILE.md` lose that half.
- `test_request_contracts.py` loses its two-extractor split: `CanonicalJson` went with the catalog and admin surfaces that used it, so every body route now answers Axum's 422 and closing #1670 means adding an extractor rather than repointing routes at one.

## Out of scope

- **Tables not dropped.** DDL and every seed migration stay. Separate forward-only migration later.
- **Specs untouched.** `docs/domain/metric-catalog/specs/` and `docs/components/backend/analytics/DESIGN.md` are now stale. Two carve-outs must be relocated, not lost: the RFC 9457 error-envelope contract that `api/error.rs` anchors to, and the MariaDB CHECK rationale that `check_probe.rs` cites.
- Surviving POST endpoints keep Axum's default rejection envelope — `canonical_json.rs` had no other consumer.

## Verification

`cargo fmt --check`, `clippy --all-targets -D warnings` clean at every commit boundary. `cargo test -p analytics`: 378 passed. `cargo check --workspace` clean. OpenAPI drift gate OK. Coverage-gate hygiene: 10 spec ops, zero stale `BLOCKED`, `SKIP_LIST` still empty.

Stand-side: `ruff check` clean, and the operation catalogue, `BLOCKED` and the committed spec cross-check at 10 operations with no drift in either direction.

Not run here — needs live infra: `--include-ignored`, `./e2e.sh`, the stand suite, gateway downstream-verify.

Closes constructorfabric/insight#1663
Closes constructorfabric/insight#1664
Closes constructorfabric/insight#1670
Part of constructorfabric/insight#1803

## Summary by CodeRabbit

* **Changes**
  * Streamlined analytics around metric definitions, results, drilldowns, exports, and saved queries.
  * Removed legacy metric and threshold management, catalog, column-discovery, and person-lookup endpoints.
  * Removed optional single-tenant metric configuration and related environment settings.
  * Removed analytics catalog table-column metadata and associated seed configuration.
* **Documentation**
  * Updated API documentation, endpoint references, and service configuration guidance for the streamlined analytics API.


ktursunov (approved):

cyberantonz:
Ref github.com/constructorfabric/insight/issues/2167

cyberantonz (approved):

cyberantonz (approved):

Issue #2194, opened 2026-08-04
Broker Phase 0: compose PoC — config-cli-only realm, refresh + logout passthrough (go/no-go)


cyberantonz:
Part of EPIC #2193 (ADR-0003). Confirmation gate: nothing else in the EPIC starts until this proves out.

- [x] Compose PoC: Keycloak realm defined **only** in keycloak-config-cli YAML, brokering one upstream OIDC provider; the **unchanged** authenticator completes login against the broker issuer and mints a gateway JWT carrying the single `tenant_id` claim.
- [x] Validate refresh-token passthrough through the broker: background refresher works, `offline_access` honored, refresh-token rotation survives the two-hop chain.
- [x] Validate logout propagation: upstream back-channel logout reaches the broker and the broker's back-channel logout reaches the authenticator (session revoked).
- [x] Record findings (pass/fail per item, gaps, workarounds) in ADR-0003's Confirmation section — feeds EPIC Phase 5.

Deliverable: throwaway compose setup + a short findings note. Refs #1782.

cyberantonz:
**Findings — first checkbox (compose PoC: broker realm as config-cli code, brokered login end-to-end)** — PR #2202

**Result: PASS.** A broker realm defined only in keycloak-config-cli YAML, brokering one upstream OIDC provider (a second realm standing in for a customer IdP), lets the **unchanged** authenticator — the stock published image, with only `issuerUrl` + client secret re-configured — complete the two-hop login and mint a gateway JWT carrying the single string `tenant_id`. An `/api/analytics` call through the gateway with the brokered session returns 200, so the downstream verifier (which fails closed without `tenant_id`) accepted the token.

What was confirmed along the way:

- **config-cli against a running server works as the gitops phase assumes**: both realms applied idempotently to the already-running compose Keycloak (26.4, config-cli `6.5.1-26.1.0`), no `--import-realm`, no admin-UI step anywhere.
- **Env-var substitution keeps secrets out of realm YAML** (`IMPORT_VARSUBSTITUTION_ENABLED=true`). One sharp edge: substitution scans the *whole file including comments*, so a literal `$(env:...)` in a YAML comment fails the import with `Cannot resolve variable 'env:...'`. Worth remembering when writing realm YAML docs in Phase 1.
[...]

Issue #2193, opened 2026-08-04
EPIC: Keycloak identity broker rollout (ADR-0003)


cyberantonz:
## Goal

Adopt Keycloak as the identity broker in front of the authenticator, per **ADR-0003** (`cpt-insightspec-adr-auth-0003-keycloak-identity-broker`, PR #2170): customer IdPs (OIDC/SAML) and social logins (GitHub #2163; Google/Facebook/Apple later) federate behind one uniform OIDC issuer; realm content is configuration-as-code only (keycloak-config-cli from gitops, secrets via SealedSecrets); per-provider mappers emit the single `tenant_id` claim; fakeidp is retired once every environment is brokered.

Resolves the investigation from #1782 (report: https://github.com/constructorfabric/insight/issues/1782#issuecomment-5176199376).

## Phases

- [x] Phase 0 — Broker PoC, go/no-go gate: #2194
- [x] Phase 1 — Broker realm as code in gitops: #2195
- [x] Phase 2 — GitHub login: #2163
- [x] Phase 3 — Tenant mapping & claim contract: #2196
- [x] Phase 3b — Host-keyed issuer map in the authenticator: #2197
- [x] Phase 4 — Migrate environments behind the broker; retire fakeidp: #2198
- [x] Phase 5 — Amend ADR-0003 with PoC findings; update the DESIGN reuse map (tracked here, no separate issue)

## Dependencies

Phase 0 gates everything. Phases 1 → 2 deliver GitHub login. Phases 3/3b run parallel to Phase 2. Phase 4 trickles per environment and completes with fakeidp deletion. Phase 5 closes the record.

Refs #1782, #2163, #1583, ADR-0002.

Issue #2195, opened 2026-08-04
Broker Phase 1: realm as code in gitops — config-cli Job, sealed secrets, canonical client scope


cyberantonz:
Part of EPIC #2193 (ADR-0003). Blocked by Phase 0 (go/no-go).

- [x] Add keycloak-config-cli as a chart/gitops sync Job applying versioned realm YAML against the shared Keycloak (idempotent; re-applies on every sync, reverting drift).
- [x] Wire broker client secrets via SealedSecrets + env-var substitution; realm YAML holds env placeholders only (no secret material) — covered by the repository's standard secret scanning, with the authoring rules documented in the gitops README (decided in review of the implementing PR: no bespoke lint).
- [x] Define the canonical `insight` client scope in realm YAML: allow-listed claims only — `sub`, `email`, one string `tenant_id` (matching what the compose realm generator emits today).

Exit criterion: a realm change is a reviewable pull request; no admin-UI configuration channel remains. Refs ADR-0002 (shared pre-provisioned Keycloak).

cyberantonz:
**Box 1 landed — PR #2204 merged (6d3406c).**

What landed: the umbrella chart gained a `post-install,post-upgrade` hook Job (`keycloakConfig.*` values gate, off by default) running `adorsys/keycloak-config-cli:6.5.1-26.1.0` against the environment's Keycloak; the gitops tree gained the `keycloak-broker-realms` Makefile target packing `environments/<env>/keycloak/realms/*.yaml` into the `<release>-keycloak-config-realms` ConfigMap the Job mounts, chained into `deploy-insight`. `IMPORT_CACHE_ENABLED=false` so every sync re-asserts the versioned realm content and reverts drift (Phase-0 finding). Review hardening: rendering fails on a non-https `keycloakConfig.url` unless `keycloakConfig.allowInsecureUrl=true` (in-cluster dev/CI only).

Relevant for the next boxes:
- Box 2 (secrets): the Job already consumes ONE existing Secret via envFrom (`keycloakConfig.existingSecret`, default `insight-keycloak-config`) that must carry `KEYCLOAK_USER`/`KEYCLOAK_PASSWORD` plus every `$(env:VAR)` the realm YAML references — box 2 is the SealedSecrets template + CI guard on top of that hook, not new chart surface.
[...]

PR #2202, opened 2026-08-04
poc(auth): Keycloak broker realm as config-cli code, brokered login end-to-end

cyberantonz:
## What

Throwaway compose PoC for the first Phase-0 checkbox of the Keycloak identity-broker EPIC: a Keycloak **broker** realm defined **only** in keycloak-config-cli YAML, brokering one upstream OIDC provider, with the **unchanged** authenticator (stock published image — only its `issuerUrl`/client-secret config changes) completing login against the broker issuer and minting a gateway JWT that carries the single string `tenant_id` claim.

New, fully self-contained under `deploy/compose/keycloak/broker-poc/`:

- `realms/poc-upstream.yaml` — stands in for a customer IdP: one user (the seeded dev persona, so Identity resolves it) + the broker's confidential client.
- `realms/insight-broker.yaml` — the broker: OIDC identity provider toward the upstream, `hardcoded-attribute-idp-mapper` stamping `tenant_id` (ADR-0003's fixed per-registration tenant), the `insight-authenticator` confidential client with the `tenant_id` protocol mapper, and a browser flow that auto-redirects to the sole IdP (the ADR's single-IdP-realm behaviour — no broker login page).
- `poc-up.sh` — applies both realms idempotently with keycloak-config-cli against the running compose Keycloak (env-var substitution for secrets and host URLs; nothing sensitive in YAML — the same mechanism that will carry SealedSecret values in the gitops phase), then recreates only the authenticator with the broker issuer.
- `verify.sh` — headless end-to-end proof with no shortcut at any step; PASSes only when the gateway JWT carries the single string `tenant_id`.

## Why

ADR-0003 (`0003-keycloak-identity-broker.md`) gates the whole EPIC on this confirmation: broker realm as code only, one brokered upstream, unchanged authenticator, single `tenant_id` in the minted token. This PoC proves that path and hands Phase 1 (realm-as-code in gitops) a working realm shape to lift.

## Verification

Ran locally on the compose stack (`AUTH_MODE=keycloak`, all backend services from published ghcr images — nothing built locally, which is the strongest form of "unchanged authenticator"):

- `verify.sh` → PASS: login starts at the broker realm, broker auto-redirects to the upstream authorize endpoint, credentials go to the upstream form, the code returns through `/realms/insight-broker/broker/poc-upstream/endpoint`, the authenticator exchanges the broker's code and sets `__Host-sid`, and `/internal/authz` returns a gateway JWT whose `tenant_id` is a single string equal to the configured tenant.
- `GET /api/analytics/v1/metrics` through the gateway with the brokered session → 200 (the downstream verifier, which fails closed without `tenant_id`, accepted the token).

Refresh-token passthrough and logout propagation are the next Phase-0 checkboxes and are deliberately out of scope here.

Refs #2194
Part of #2193

### DIFF
--- a/deploy/compose/keycloak/broker-poc/README.md
+++ b/deploy/compose/keycloak/broker-poc/README.md
@@ -1,0 +1,68 @@
+# Keycloak identity-broker PoC (ADR-0003, issue #2194)
+
+Throwaway compose setup proving the first Phase-0 gate of EPIC #2193: a
+Keycloak **broker** realm defined **only** in
+[keycloak-config-cli](https://github.com/adorsys/keycloak-config-cli) YAML,
+brokering one upstream OIDC provider, with the **unchanged** authenticator
+(stock published image, config-surface change only) completing login against
+the broker issuer and minting a gateway JWT that carries the single
+`tenant_id` claim.
+
+## Topology
+
+```text
+browser ── /auth/login ─▶ authenticator ── code+PKCE ─▶ insight-broker realm
+                                                          │ auto-redirect (sole IdP)
+                                                          ▼
+                                                        poc-upstream realm  (stands in for a customer IdP)
+                                                          │ code back to /realms/insight-broker/broker/poc-upstream/endpoint
+                                                          ▼
+                              authenticator ◀── broker code ── insight-broker realm
+                                   │  exchanges code, resolves person via Identity,
+                                   ▼  sets __Host-sid, mints gateway JWT {tenant_id}
+                              gateway auth_request ─▶ analytics / identity
+```
+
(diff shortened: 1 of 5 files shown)


cyberantonz:
Closing unmerged by design: a throwaway PoC should not land in the tree. The findings and the full reproducible assets (realm YAMLs + run recipe) are recorded on the phase issue instead: https://github.com/constructorfabric/insight/issues/2194#issuecomment-5179536169 and https://github.com/constructorfabric/insight/issues/2194#issuecomment-5179618091. Phase 1 (#2195) lifts the realm shape from there.

Issue #2208, opened 2026-08-05
[pres] Semantic Layer Phase 1 — Definition Core


cyberantonz:
Phase B / Semantic Layer **Phase 1 — Definition Core**. Parent: #1803.

Prerequisite for the Phase 2 compiler (#1976). Establishes definitions-as-data as the single runtime source of truth, per the adopted architecture in `docs/domain/semantic-layer/specs/` and **ADR-001** (`cpt-semantic-layer-adr-adopt-compiler-over-datasets`).

Scope:
- Typed definition format (datasets, measures, metrics) as code in the owning service; authored YAML and stored rows are serializations of the same types.
- Definition store as the single runtime source of truth (`cpt-semantic-layer-component-definition-store`), with versioned revisions.
- Field catalog derived from dataset schemas (`cpt-semantic-layer-component-field-catalog`) as the validation universe.
- Shared validators (build-time for product definitions, write-time for tenant definitions) so an invalid definition is never written.

Note: #1974 shipped the first step of this phase (metric definitions authored as data — the YAML registry replacing Rust constants) in the transitional observation-relation shape; this issue rewrites that into the dataset/measure/metric domain model the compiler consumes.

Out of scope: the compiler and executor cutover (#1976, Phase 2); discovery API and runtime editing (Phases 4–5).

---
### Phase 1 slices (tracking)

- [x] **1. Schema (domain-shaped tables)** — `datasets`/`measures`/`metrics`/`definition_revisions` + CHECKs, startup-probe registration. PR #2210.
- [x] **2. Field catalog** — hybrid source (ClickHouse type snapshot + authored roles; dbt `meta:` blocks don't exist), Rust join into the build-time validation universe, drift-gated against live CH. PR #2211.
- [ ] **3. Validators** — `expr.rs` (`sqlparser` ClickHouse dialect + AST allowlist) and the closed-enum filter-tree types; one shared write path.
- [ ] **4. Authoring + store wiring** — sea-orm entities; product definitions as repo YAML embedded + build-time validated; startup seed reconciliation into the new store.
- [ ] **5. Extraction** — transcribe every gold-observation measure into YAML against silver/gold datasets.
- [ ] **6. Structural validation** — CI check against the field catalog + runtime contract-reconciliation probe (self-healing `unavailable` state, never a boot failure).

Exit: every product measure/metric exists as validated YAML reconciled into the new store; the legacy store untouched and still serving (Phase 2 = #1976 compiler + cutover).

PR #2211, opened 2026-08-05
feat(semantic): field catalog — types from ClickHouse + authored roles (#2208)

cyberantonz:
## What

**Phase 1, slice 2** of the Semantic Layer definition core (ADR-001): the **field catalog** — the typed, role-annotated view of each dataset that is the compiler's *validation universe*. This is what makes capability a projection of definitions rather than of stored data.

## The source problem, and the chosen fix

IMPLEMENTATION assumed the catalog is generated from dbt `schema.yml` `meta:` blocks — **but those don't exist** (no `meta:`, no `data_type`, no role annotations anywhere in the dbt schemas). So the source is **hybrid** (the approach chosen for this slice):

| Half | Source | Committed as |
|------|--------|--------------|
| Column **types** | ClickHouse `system.columns` | `types.snapshot.json` (drift-gated against live CH) |
| Field **roles** | authored in the service | `roles.yaml` (tenant / entity / event_time / dimension / value / subject) |

`domain/field_catalog/` (Rust) **joins** the two into a typed `FieldCatalog` and **rejects** any role that names a column the warehouse doesn't have or a ClickHouse type the layer can't model. So the embedded catalog is a consistent validation universe **at build time, with no live warehouse in CI**.

## Why a committed snapshot + drift test

Build-time validation (the "definitions validate against the catalog" gate) must work offline in CI. So types are captured into a committed snapshot; an **ignored drift test** (`INTEGRATION_TESTS_CLICKHOUSE_URL=…`) re-checks every catalogued field's type against live ClickHouse, so schema drift is caught and the snapshot regenerated. Same golden-snapshot + drift-gate idiom as `openapi.json`.

## Scope

Seed datasets are the two git relations with a **direct entity**: `class_git_commits`, `class_git_pull_requests`. Relations without a self-contained entity (e.g. `class_git_file_changes`, which needs a join to commits) are derived-dataset cases catalogued during **extraction (slice 5)**. Inspect with `analytics field-catalog`.

## Test

- **Offline** (normal CI, no CH): builds the embedded catalog; asserts role invariants (exactly one tenant + one entity, ≥1 event_time), type normalization (peels `Nullable`/`LowCardinality`/`DateTime64`), and every rejection path (absent column, unmodeled type, missing entity, absent relation).
- **Live** (proven here): the drift test passes against **ClickHouse 25.7** — the committed snapshot matches `system.columns` for every catalogued field.
- `cargo clippy` / `cargo fmt --check` clean; full offline suite 570 green.

Extends the semantic-layer DESIGN field-catalog component with slice-2 status, recording the hybrid-source deviation from dbt `meta:` blocks.

Part of #2208
Part of #1803

## Summary by CodeRabbit

* **New Features**
  * Added an offline command to view the semantic-layer field catalog without backend access.
  * Added catalog coverage for 12 Git, AI, collaboration, and wiki datasets, including field roles, types, nullability, and read disciplines.
  * Added validation for missing fields, unsupported types, duplicate datasets, and schema inconsistencies.
* **Documentation**
  * Documented catalog coverage, deferred datasets, tenant naming differences, and decimal type support.
* **Tests**
  * Added automated checks for catalog construction and optional live schema drift detection.


cyberantonz:
Dropping this PR. It reinvented a schema-snapshot the repo already produces and gates (`connectors-ddl/silver.sql`, refreshed by the required `connectors-ddl` lane from the real pipeline). The field catalog will be redesigned to source column types + engine from that DDL instead of a bespoke ClickHouse snapshot — tracked as **#2214** under the Semantic Layer epic (**#2213**).

Salvageable from this branch for the redesign: the `FieldType`/`FieldRole`/`Dataset` model, the CH-type normalizer, the authored `roles.yaml`, and the offline consistency tests. What goes: `types.snapshot.json` and the ignored CH drift test, replaced by parsing the committed DDL.

Issue #1995, opened 2026-07-28
analytics-api: /v1/metric-results has no per-person authorization — Identity's visibility rule is bypassed


dzarlax:
## Summary

`POST /v1/metric-results` performs no per-person authorization. It reads only
`ctx.subject_tenant_id()`; the caller's `person_id` is never consulted, no
visibility check is made, and the incoming `Authorization` header is not
forwarded to Identity (unlike `get_person`, which does forward it). Any
authenticated user of a tenant can therefore read the metric values of any
`entity_id` in that tenant — including people above them in the org chart, in
sibling branches, or in unrelated departments.

Identity already implements the intended rule and analytics bypasses it, so the
org-chart boundary users see in the UI is presentational only.

## Where

- `src/backend/services/analytics/src/api/metric_results.rs:30` —
  `query_metric_results` uses the security context for the tenant only.
- `src/backend/services/analytics/src/domain/metric_results/validation.rs:88` —
  `validate_request` takes `tenant_id`, nothing person-scoped.
- Contrast: `src/backend/services/analytics/src/api/handlers.rs:44` propagates the
  bearer to Identity for the person lookup — the pattern exists, it is just not
  applied to metric reads.

For comparison, the rule Identity enforces
(`src/backend/services/identity/src/Insight.Identity.Infrastructure/MariaDb/Sql.Visibility.cs:29`,
`VisibilityService.CanSeeAsync`): visible set = self ∪ active explicit grants ∪
all `org_chart` descendants of anything visible ∪ the whole tenant when a grant
with `viewed_person_id IS NULL` exists. Gates `GET /v1/persons/{email}` and
`POST /v1/profiles`.

## Reproduction / degradation

1. Authenticate as a person with no visibility grant.
2. Open the person dashboard of someone outside their subtree (the frontend has
   no route guard — the URL carries the email).
3. Identity denies the profile, so the page header renders empty. Analytics
   answers `/v1/metric-results` in full, so every KPI, bullet, trend and peer
   percentile for that person renders with real values.

The failure mode is worse than a clean 403: it looks like a rendering bug while
the data is served. It is also not observable in a default local install,
because the dev bootstrap seeds whole-tenant grants (`viewed_person_id IS NULL`)
for ~10 accounts — every one of them passes `CanSeeAsync` for the entire tenant,
so the gap only appears once grants are realistic.

## Why it matters now

v1 ships as three single-tenant installations, so tenant isolation is not the
[...]

Issue #1984, opened 2026-07-28
Team and organisation distributions cannot be built from the histogram response


dzarlax:
## Summary

A histogram is binned separately for each person, so distributions cannot be combined across a team or an organisation — the scope a lead most often asks about.

## What is needed

A way for the caller to ask for **shared bin edges across the requested entity set**. Any of these shapes works for the UI, in order of preference:

1. `{ view: "histogram", bin_scope: "request" }` — bins computed over the union of the requested entities, with the shared `lo`/`hi` returned once and per-entity counts alongside. An absent `bin_scope` keeps today's behaviour.
2. Caller-supplied edges: `{ view: "histogram", bins: { lo, hi, count } }` — the UI picks a readable 1/2/5 ladder and the server counts into it.
3. An aggregate-only variant returning counts for the whole set without per-entity rows, for when only the org shape is needed.

Per-entity binning needs to stay whichever shape is chosen. The Person screens use it to show one individual's own spread.

## Current behaviour

Bin index is computed against each entity's own minimum and maximum, so bin 3 covers a different value range for every person:

```sql
min(event_value) OVER (PARTITION BY entity_id) AS entity_lo,
max(event_value) OVER (PARTITION BY entity_id) AS entity_hi
...
toUInt32(least({max_bin}, toInt64(floor(
    (events.event_value - events.entity_lo) * {bins} / (events.entity_hi - events.entity_lo)
))))
```

- Each response row carries its own `entity_lo` and `entity_hi`, and the bin count is a fixed constant.
- Merging on the client is not a workaround. `mergeEventHistogram` returns `null` when edges differ, when an entity has an unexpected bin count, or when nobody has bins; its own comment records this as an open design question. Probed against a live dataset with a roster of 150+ people, the edges never aligned.
- Two lenses are built around a team-level distribution — Development · Flow and Overview · Contribution breakdown. Both are currently suppressed at org scope or approximated from per-person period values, which loses the within-person spread the histogram exists to show.
- Verified again on 2026-08-25: the per-entity binning is unchanged and no shared-edge option exists. The client-side merge now lives at `src/frontend/src/lib/portal/event-histogram.ts` in this repository; the original report linked it in `insight-front`, which has since been merged here.

Issue #1974, opened 2026-07-28
[pres] Declarative metric registry (single YAML)


cyberantonz:
Phase B. Collapse analytics seed + FE thresholds + catalog spec into one registry. Parent: #1803.

dzarlax:
Concrete input for the registry from the frontend side, measured on a live install today — because "collapse analytics seed + FE thresholds + catalog spec" hides one fact worth stating up front: **there are no threshold values to collapse for any metric the API currently serves.**

```
SELECT COUNT(*) FROM metric_threshold;                    -- 108
SELECT scope, COUNT(*) FROM metric_threshold GROUP BY 1;   -- product-default: 108 (no tenant/role/team rows)

-- namespaces those 108 keys belong to:
ai_bullet_rows 23 · collab_bullet_rows 21 · task_delivery_bullet_rows 16 ·
ai_person_counter_daily 10 · crm_bullet_rows 8 · git_bullet_rows 7 · ic_kpis 7 ·
support_bullet_rows 7 · wiki_bullet_rows 4 · code_quality_bullet_rows 3 ·
collab_person_counter_daily 2

-- unified metric_definitions keys with a threshold at ANY scope:
0 of 59   (ai.* collab.* git.* tasks.* wiki.*)
```

The thresholds were authored against pre-catalog view keys (`ai_bullet_rows.cc_cost`, `git_bullet_rows.*`, …) and did not travel with the metric migration — same shape as #1930 (CRM/support) and the seat-overage metric (see the note on #1607). So #1977 ("thresholds/units from catalog") will find the plumbing ready and the values absent: the FE evaluator already consumes `ResolvedThresholds` (`evaluateStatus` in `src/api/metric-semantics.ts`), and the catalog client already types them.

Two asks for the registry, both cheap to encode in YAML and both things the frontend cannot decide on its own:

1. **A value or an explicit absence per metric.** Many counters are scale-dependent — "how many commits is good" is a team question, not a product one — so a threshold-less metric is a legitimate outcome. What the UI cannot handle is ambiguity: today a missing row is indistinguishable from an unauthored one. If the registry carries `thresholds: none # scale-dependent, use peer comparison` for those metrics, the UI can confidently render value + peer standing and no status, instead of looking for numbers that will never come.
[...]

Issue #1682, opened 2026-07-07
CONFIGURE 8: [EPIC] Admin configuration console — roles, mappings, thresholds & install specifics


dzarlax:
## Goal

A single **tenant-admin product-configuration console** — the place where an admin tunes the installation's **analytical model** and all per-install specifics: **roles + role→metric mappings, metric definitions & thresholds, cohorts, data filtering, framework packs**. It is the **manual / direct** counterpart to the copilot's Configure capability (#1458): the copilot proposes conversationally, this console lets an admin configure directly.

Distinct from the **infra** admin surfaces — Instance Admin UI (#1639, instance/tenant provisioning), Tenants Config UI (#1640, instance defaults + 2FA), and the settings store (#1633). This epic is the **product-config cockpit**; it **surfaces** the config objects owned by their epics, it does not reimplement them.

## Scope

- **Roles & role→metric mapping** — define / split / merge / rename / retire roles, map HR roles to presets or fork-and-edit, author the role→important-metrics mapping (surfaces #1455).
- **Metrics & thresholds** — edit metric definitions / calculation (#1659), set thresholds by function / team / tenant / period (#1616).
- **Cohorts & baselines** — configure comparison cohorts (#1455 / #1613).
- **Data filtering & exclusion** — in / out-of-scope config (#1636).
- **Framework packs** — enable / disable per tenant (#1616).
- One place for all install-specific settings; changes are **validated + versioned + auditable**, and **editable without a release** where the underlying epic supports it (#1659 metrics, #1612 rules, #1455 roles).

## Acceptance Criteria

- A tenant admin configures roles + role→metric mappings, metrics + thresholds, cohorts, data filtering, and framework packs from one console.
- Config changes are validated, versioned, and auditable; editable-without-release where supported (#1659 / #1612 / #1455).
- The console **surfaces** config owned by #1455 / #1616 / #1659 / #1636 / #1613 — it does not reimplement those.
- It is the manual counterpart to the copilot Configure path (#1458).
- Configures Insight itself only — never operates monitored systems (read-only architecture).

## Relations

- Surfaces roles/mappings #1455, packs/thresholds #1616, metric calc #1659, data filtering #1636, cohorts #1613.
- Manual counterpart to the copilot Configure capability #1458.
- Distinct from infra #1639 / #1640 / #1633.
- [VISION.md](https://github.com/constructorfabric/insight/blob/main/docs/product/VISION.md) §9 (customer configuration); read-only architecture.

[...]

PR #2241, opened 2026-08-05
feat(portal): organizational analytics portal behind the insight.portal flag

dzarlax:
## What

An organizational analytics **portal** — a manager-first shell over the existing metric system — gated behind the `insight.portal` localStorage flag (default off; nothing changes for current users). New code lives under `src/frontend/src/components/portal/` and `src/frontend/src/lib/portal/`, plus a few shared libs extracted with tests.

Turn it on: `localStorage.setItem("insight.portal", "true")` and reload (the flag is compared against the literal `"true"`).

**This replaces constructorfabric/insight-front#221**, re-based onto the monorepo now that the frontend lives at `src/frontend`. Same tree, re-cut as five logical commits (charts, metric grammar, URL navigation, shell, zone views); the granular per-task history stays on `dzarlax/insight-front-cf@feat/portal-shell` for anyone who wants the step-by-step.

## Why

The current dashboard is person-centric. Managers need the opposite entry point: the org first (their subtree), domains second, people last. The portal is that shell — and it doubles as a testbed for a **uniform metric grammar**, so every tab renders honestly from whatever connector set a tenant actually has (validated against two real datasets with very different coverage).

## Structure

- **Zones** (left rail): Overview / Directions / Person / People / AI & Cost / Manage.
- **Global topbar**: `Scope` (any node of the viewer's subtree plus a direct-only switch — the permission boundary is the viewer's own tree), `Slice` (dimensions discovered from identity attributes; single-valued and near-unique ones are hidden automatically), and the period selector. The bar is sticky, so the three global controls stay reachable from anywhere in a long dashboard.
- **IC-aware**: a viewer with no reports gets a Person-only shell.

## Everything is in the URL

Navigation state used to live in memory. A reload reset the screen, Back went somewhere unrelated, and a link or a bookmark could not reproduce a view — the main objection in review on the previous PR.

Now: `/portal?zone=&item=&dir=&lens=&scope=&direct=&slice=&period=&from=&to=` plus `/ic/$person/personal|team`. Only true reader preferences (is the portal on, are planned sections shown) remain in localStorage.

- Params are validated, not trusted: an unrecognised value is dropped so a mistyped link degrades to the computed default, and an inverted or over-long custom range falls back to the preset instead of throwing into the error boundary.
- Effect-driven corrections (pinning the landing zone, syncing scope from the route) use `replace`, so Back never steps into a half-built address.
- The path wins over a stale `?zone=` — the URL cannot contradict itself.
- Defaults stay computed rather than serialised, so a shared link pins what the sender actually chose and lets the rest resolve for the recipient.

## Keyed on person ids

Person keys are canonical person ids everywhere: the `/ic/$person/*` route segment (guarded by `isPersonId`, which redirects a pre-cutover email URL instead of sending it to the API as an unactionable 400), `?scope=`, the roster maps, and the metric entity ids. No email-to-id bridge — the frontend and metrics are on ids on dev, and a translation layer would outlive its reason to exist.

Two consequences worth naming: cohorts are keyed by id, so a person with no email still belongs to one and still appears in the employees directory; and the manager step in the person header uses `parent_person_id`.

## The metric grammar (the core of the PR)

Directions and Overview render through ONE config-driven renderer (`DomainLensView`) over a typed section library (`SectionSpec`): headline, stat-tiles, trend, distribution, concentration, composition, participation, event-histogram, attention, direction-cards, coverage-radar. Registries (`DIRECTION_LENSES`, `OVERVIEW_ITEMS`) declare every tab; invariant tests pin nav/registry coverage both ways and the metric-key budget.

Rules the renderer enforces everywhere:

- counters render per-active-person with a period-over-period delta coloured by the server-owned `direction`; ratios and medians are never summed;
- distributions use integer 1/2/5-ladder bins; concentration is a top-decile share with explicit framing (bus-factor risk for git, load balance for collaboration);
- **honest data-gating**: a family never observed for the scope renders a not-ingested state, never zeros (`entityObserved` reads the peer view's nulls, which the backend guarantees via `join_use_nulls = 1`, so zero-filled sums do not count as data);
- self-suppression on degenerate data (cohorts below four, single-bin histograms, single-valued slices), always with a sentence saying why rather than going silent;
- org zones never rank named individuals; names appear only in "needs attention" as click-through pointers into Person;
- trend requests coarsen their bucket (day to week to month) from roster size, and when no bucket fits the row budget the section says so instead of firing a request the API rejects;
- **no status without a threshold.** Deltas are coloured by `direction`, but nothing is labelled good or bad: the unified metric keys carry no threshold rows, and hardcoding good/warn in a component would turn a product decision into invisible frontend trivia. Status lights up the moment the catalog carries values (constructorfabric/insight#1974).

## Charts

Line charts were configured per call site, so the same reading looked different depending on the screen. `ChartLine` now owns the defaults: linear interpolation (a chart must not draw a value nobody measured), adaptive dots (dropped above 31 points, but an isolated reading between two nulls is always marked so a single point is never invisible), and gaps that read as gaps. `ChartYAxis` abbreviates ticks from 10k — the point where four digits still fit the gutter and five clip — and stays exact below it, because a tick sits on a gridline and "2.3k" for 2250 is a label that lies.

## Three layout tiers

`useShellLayout` resolves phone (<768) / narrow (768-1023) / wide (>=1024) in one place: the rail is a drawer on a phone, collapsible on narrow, pinned on wide, and pane state resets when the tier changes instead of restoring a mode that no longer exists. On a phone the drawer shows a one-row zone switcher and folds settings behind a single row — the expanded list pushed the sections below the fold, which is the content the reader opened the drawer to reach.

## Tests

785 unit tests over the new code, semantics-first rather than snapshot-first:

- **libs**: attention flags (outliers, collapses, adverse deltas, cohort isolation, scale-free severity in cohort IQRs, dedup to the strongest flag per person and metric), slice discovery gates, peer quantiles with small-cohort suppression, per-active-person maths, trend bucket selection including the no-bucket-fits case;
- **hooks**: URL-vs-path zone resolution, manager detection, person cohorts, org-scope resolution inside the viewer's own subtree;
- **components**: the section renderer (per-capita and deltas, active-only denominators, not-ingested gate, distribution suppression, concentration framing, composition shares and error card, participation, by-unit under a slice, attention rows, radar suppression), team state (totals vs medians, empty columns dropped), AI & Cost (Claude-only cost caveat, "not tracked" never $0), metric groups, context pane, shell routing across the three tiers, attention list, org-scope gate (an error never masquerades as an empty team), employees directory, Manage.

`src/test/portal-router` is a reactive router fake so navigation is asserted as real hrefs and recorded navigations; `src/test/identity` mints valid person UUIDs from short labels, so fixtures stay readable without drifting from the real key shape.

Verified in the ported tree: `pnpm typecheck` clean, `eslint --max-warnings 0` clean, 785 unit tests pass. The two Storybook browser suites are unchanged by this PR.

## Verified live

Walked every zone and tab at org scope (150+ people) and at narrowed scopes on two real datasets with different connector sets: full coverage on one; the other missing task and git freshness — every gap rendered as an explicit not-ingested state instead of fabricated zeros, which is exactly the behaviour this PR is after.

## Out of scope / follow-ups

- Server-side by-unit aggregation and an org-level histogram with shared bin edges (constructorfabric/insight#1984 — the client merges per-entity bins only when edges align, and falls back honestly otherwise).
- Threshold column with provenance, then editing via `/v1/admin/metric-thresholds` (blocked on values existing: constructorfabric/insight#1974, #1682).
- Analytics authorization for viewing a person above you in the org chart (constructorfabric/insight#1995).
- Development / Repositories / Elements — frontend work on dimensions that already exist (`repository`, `project`, `file_extension`, `change_type`).
- Phase-2 filter model (persistent attribute filters on top of scope and slice).
- Scorecard and Reports zones are ComingSoon scaffolds.

All five commits are signed off (DCO).

## Summary by CodeRabbit

* **New Features**
  * Added a responsive Portal experience with navigation, organization scoping, team and employee views, personal dashboards, metric groups, trends, attention insights, and AI cost reporting.
  * Added configurable overview, direction, and metric-management views with loading, error, empty, and Coming Soon states.
  * Added portal settings for enabling the experience and showing planned sections.
  * Added responsive charts, compact axis labels, period and slice selectors, and persistent period preferences.
* **Bug Fixes**
  * Prevented metric requests when entity lists or date ranges are incomplete.
  * Improved chart handling for sparse and isolated data points.

### DIFF
--- a/src/frontend/src/api/metric-definitions-client.ts
+++ b/src/frontend/src/api/metric-definitions-client.ts
@@ -19,7 +19,7 @@ import type {
 const BASE =
   (import.meta.env.VITE_API_BASE as string | undefined) ?? "/api/analytics/v1";
 
-type MetricDefinitionSchemaStatus = "ok" | "error" | "unchecked";
+export type MetricDefinitionSchemaStatus = "ok" | "error" | "unchecked";
 
 export interface MetricDefinition {
   metric_key: string;
--- a/src/frontend/src/api/metric-results-client.test.ts
+++ b/src/frontend/src/api/metric-results-client.test.ts
@@ -33,6 +33,15 @@ function response(init: {
 describe("queryMetricResults", () => {
   beforeEach(() => mockFetch.mockReset());
 
+  it("refuses an empty entity list instead of letting the server 400", async () => {
+    // The backend answers `entity.ids must not be empty`; a caller that got
+    // here has a gating bug, and a network error hides that.
+    await expect(
+      queryMetricResults({ ...REQUEST, entity: { type: "person", ids: [] } }),
+    ).rejects.toThrow(/empty entity list/);
+    expect(mockFetch).not.toHaveBeenCalled();
+  });
+
   it("returns the parsed response on success", async () => {
     mockFetch.mockResolvedValue(
       response({ ok: true, json: async () => METRIC_RESULTS_RESPONSE_FIXTURE }),
--- a/src/frontend/src/api/metric-results-client.ts
+++ b/src/frontend/src/api/metric-results-client.ts
@@ -183,6 +183,15 @@ export interface MetricResultsResponse {
 export async function queryMetricResults(
   body: MetricResultsRequest
 ): Promise<MetricResultsResponse> {
+  // Refuse a request the backend is guaranteed to reject (400 invalid_argument,
+  // "entity.ids must not be empty"). Callers are expected to keep the query
+  // disabled until they have entities; failing here names the real cause
+  // instead of surfacing a server validation error in the network log.
+  if (body.entity.ids.length === 0) {
+    throw new Error(
+      "metric-results: refusing to request an empty entity list — the caller should stay disabled until the roster resolves",
+    );
+  }
   const res = await fetchWithAuth(`${BASE}/metric-results`, {
     method: "POST",
     headers: { "Content-Type": "application/json" },
--- a/src/frontend/src/components/app-sidebar-footer.tsx
+++ b/src/frontend/src/components/app-sidebar-footer.tsx
@@ -1,0 +1,83 @@
+import { Link, useRouterState } from "@tanstack/react-router";
+import { BookOpenText, Megaphone } from "lucide-react";
+import { useTranslation } from "react-i18next";
+
+import { useViewer } from "@/auth";
+import { SidebarSettings } from "@/components/sidebar-settings";
+import { ThemeSwitcher } from "@/components/theme-switcher";
+import { Avatar, AvatarFallback } from "@/components/ui/avatar";
+import {
+  SidebarMenu,
+  SidebarMenuButton,
+  SidebarMenuItem,
+} from "@/components/ui/sidebar";
+import { getInitials } from "@/lib/insight/get-initials";
+import { useIcPerson } from "@/queries/ic-dashboard";
+
+/**
+ * Shared footer for the sidebar chrome: metric catalog, What's new, view
+ * settings (portal / focus / explanations), theme switch, and the viewer
+ * identity block. Extracted from AppSidebar so the portal shell can surface
+ * the same controls (from the rail's settings popover) without duplicating them.
+ */
+export function AppSidebarFooter() {
+  const { t } = useTranslation();
+  const { email: viewerEmail, personId: viewerPersonId } = useViewer();
+  const viewerQ = useIcPerson(viewerPersonId ?? "");
+  const viewer = viewerQ.data ?? null;
+  const pathname = useRouterState({ select: (s) => s.location.pathname });
+
+  const primaryEmail = viewer?.email ?? viewerEmail;
+  const primary = viewer?.display_name || primaryEmail;
+  const showSecondary = primary !== primaryEmail;
+
+  return (
+    <>
+      <SidebarMenu>
+        <SidebarMenuItem>
+          <SidebarMenuButton
+            isActive={pathname === "/metrics"}
+            render={<Link to="/metrics" />}
+          >
+            <BookOpenText />
+            <span>{t("metric_definitions.nav_label")}</span>
+          </SidebarMenuButton>
+        </SidebarMenuItem>
+        <SidebarMenuItem>
+          <SidebarMenuButton
+            isActive={pathname === "/whats-new"}
+            render={<Link to="/whats-new" />}
+          >
+            <Megaphone />
+            <span>{t("whats_new.nav_label")}</span>
+          </SidebarMenuButton>
+        </SidebarMenuItem>
+      </SidebarMenu>
+      <SidebarSettings />
+      <ThemeSwitcher />
+      {viewerEmail ? (
+        <SidebarMenu>
+          <SidebarMenuItem>
+            <SidebarMenuButton size="lg" className="cursor-default">
+              <Avatar className="size-8 shrink-0">
+                <AvatarFallback className="bg-sidebar-primary text-xs font-semibold text-sidebar-primary-foreground">
+                  {getInitials(primary) || "?"}
+                </AvatarFallback>
+              </Avatar>
+              <div className="flex min-w-0 flex-1 flex-col leading-tight">
+                <span className="truncate text-sm font-medium text-sidebar-foreground">
+                  {primary}
+                </span>
+                {showSecondary ? (
+                  <span className="truncate text-xs text-sidebar-foreground/60">
+                    {primaryEmail}
+                  </span>
+                ) : null}
+              </div>
+            </SidebarMenuButton>
+          </SidebarMenuItem>
+        </SidebarMenu>
+      ) : null}
+    </>
+  );
+}
--- a/src/frontend/src/components/app-sidebar.tsx
+++ b/src/frontend/src/components/app-sidebar.tsx
@@ -21,6 +21,7 @@ import {
   SidebarMenuButton,
   SidebarMenuItem,
 } from "@/components/ui/sidebar";
+import { personIdFromPath } from "@/lib/metrics/entity";
 import { useIcPerson } from "@/queries/ic-dashboard";
 import type { IdentityPerson } from "@/types/insight";
 
@@ -106,8 +107,8 @@ export function AppSidebar() {
   // The URL segment is the person id since the identity cutover; a legacy
   // email URL simply highlights nothing for the moment its redirect takes.
   const activePersonId = useMemo(() => {
-    const m = /^\/ic\/([^/]+)/.exec(pathname);
-    if (m) return decodeURIComponent(m[1]!);
+    const fromPath = personIdFromPath(pathname);
+    if (fromPath) return fromPath;
     if (pathname === "/" && viewerPersonId) return viewerPersonId;
     return null;
   }, [pathname, viewerPersonId]);
--- a/src/frontend/src/components/mock-banner.test.tsx
+++ b/src/frontend/src/components/mock-banner.test.tsx
@@ -1,19 +1,25 @@
 import { render, screen } from "@testing-library/react";
-import { afterEach, describe, expect, it, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 import "@/i18n";
 
 import { MockBanner } from "@/components/mock-banner";
 
-// `isVisible()` also requires import.meta.env.DEV, which vitest sets to true,
-// so the mock flags below are the only knobs the tests need to turn.
+// `isVisible()` also requires import.meta.env.DEV, which vitest sets to true.
+// Both mock flags are stubbed to a known baseline first: vite loads the
+// developer's own `.env.local`, so a machine with VITE_HIDE_MOCK_BANNER=true
+// would otherwise fail the visible case while CI (no `.env.local`) passes.
 describe("MockBanner", () => {
+  beforeEach(() => {
+    vi.stubEnv("VITE_ENABLE_MOCKS", "false");
+    vi.stubEnv("VITE_HIDE_MOCK_BANNER", "");
+  });
+
   afterEach(() => {
     vi.unstubAllEnvs();
   });
 
   it("renders nothing when mocks are not enabled", () => {
-    vi.stubEnv("VITE_ENABLE_MOCKS", "false");
     const { container } = render(<MockBanner />);
     expect(container).toBeEmptyDOMElement();
   });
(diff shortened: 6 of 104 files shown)


dzarlax:
Thanks — six of the eight are real and fixed in `e6f11cd9`. Two are not bugs; details below.

**Fixed**

- **`person-header.tsx` — teammate navigation by email.** Correct, and the one bug in the identity-cutover migration: `goPerson` and `goTeam` were converted, the peers dropdown was not. Now `p.person_id`, with the comparison and the menu key on the id too.
- **`/portal` renders with the preview off.** Correct and worse than it looks: the root shell only swaps in `PortalLayout` while the flag is on, so with it off the route mounted the portal *inside* the app chrome. The route component now redirects home, with tests for both states (`src/routes/portal.test.tsx`).
- **`setCustomRange` throws from an event handler.** Fixed: it drops an invalid range, matching the policy `validatePortalSearch` states. The picker keeps its own validation message.
- **`decodeURIComponent` on the path segment.** Fixed at all three sites (`use-active-zone`, `app-sidebar`, `org-tree` did the same raw decode) via one `personIdFromPath` in `lib/metrics/entity.ts`, which returns null on a malformed sequence so callers fall back to the viewer.
- **Unguarded localStorage reads.** Fixed: one guarded `readKey`, since these run at module scope and a `SecurityError` there takes the bundle down over a preview flag. Covered by a test that makes `getItem` throw.
- **`usePortalScope` identity.** Fixed with `useMemo` on the two primitives; the quadratic re-walk you describe was the real cost.

**Partly taken: route files should stay thin.** Moved the shell-path predicate out of `__root.tsx` into `lib/portal/portal-routes.ts`, so the route file asks instead of deciding. The rest I would keep: `validateSearch` and the `retainSearchParams` middleware are route *configuration* that TanStack requires on the route, and the `isPersonId` / preview-flag guards follow the pattern already in `routes/ic.$person.personal.tsx` upstream. `__root.tsx` is the app shell rather than a screen wrapper, so a shell/chrome decision there is where it belongs.

**Not a bug: the lens readiness filter.** The vocabulary is the other way round from your reading, and it matches the shared helper: in `partitionByReadiness` (`lib/portal/nav-model.ts` line 184) `planned` is roadmap the reader benefits from seeing — always listed, dimmed — while `unbuilt` is our own backlog and appears only when the viewer asked for planned work. `entry.readiness === "planned" || showPlanned` says exactly that, and the comment above it describes the same two rules.

**Not a bug: partial event histograms.** Skipping a member with no bins is deliberate: they had no events in the period, which is a normal reading on any real roster. Returning null there would blank the org chart whenever one person was inactive — the opposite of the intent. The compatibility contract is about *bin edges*, which is still enforced (differing edges or an anomalous bin count return null). The doc comment now says both things outright, since the old wording was what made this ambiguous.

dzarlax:
Correcting myself on the `overview-configs` point above: I wrote that the API has no field for this. That is wrong, and it makes your finding bigger rather than smaller.

The registry carries **59 metrics, all 59 with `description` and `explanation`, 46 with `unit`** ([`registry.yaml`](https://github.com/constructorfabric/insight/blob/main/src/backend/services/analytics/src/domain/metric_definitions/registry.yaml)), `/v1/metric-results` ships both ([`builder.rs#L286`](https://github.com/constructorfabric/insight/blob/main/src/backend/services/analytics/src/domain/metric_results/builder.rs#L286)), and the frontend already parses them ([`collection.ts#L143-L144`](https://github.com/constructorfabric/insight/blob/main/src/frontend/src/lib/metrics/collection.ts#L143-L144)). The portal's section renderer uses neither, while other screens do (`metric-summary-card`, `kpi-row`, `peer-story`). So the config is not just writing copy the catalog owns — the portal is also dropping a populated field the reader needs. On meeting hours the `explanation` carries the caveat that Zoom reports full-session estimates and runs higher than Teams, which is exactly what someone comparing two teams has to know, and we show it nowhere.

Three of the four fields are derivable, one is not:

| Field | Verdict |
|---|---|
| `title` | derivable from the metric label plus the section kind |
| `caption`, first clause ("how many people fall in each commit-count band") | derivable the same way |
| `caption`, second clause ("a long right tail means a few people produce most of the commits") | genuinely editorial — it is about the chart shape, not the metric, and no API field carries it |
| `unitLabel` | derivable from `unit`, with a fallback: 13 of 59 metrics have no unit, and `h` expands to "h per person" where the hand-written label reads "meeting hours per person" |

I have not changed it in this PR, because the honest version of the fix is "the portal shows `explanation`", and that is a new affordance with a placement decision rather than a config cleanup — doing it here would either sprawl the PR or ship the cosmetic half. Filed as #2242 with the three steps in order, and the interpretive caption sentence is the only metric copy that stays in config afterwards.

hello1101n (approved):

ktursunov (approved):

dzarlax:
I had only answered the nine inline findings — the review body carried another 12 minor and 21 nitpick items, which I missed. Worked through all of them in `29021264`. Most were right; the ones I skipped are named at the end.

**Correctness and copy**

- `formatAxisTick` printed `1000k` for 999_600 — a tick naming a magnitude the scale below it already covers. Rolls over to M.
- `distribution` binned a negative value into `counts[-1]`, a property outside the array, so the person disappeared from the histogram and the bars stopped summing to the population.
- `fmtCompact` stopped at thousands, so a million-level bin edge read `1000k`. Fixed with its own million step rather than by delegating to `formatAxisTick` — a bin edge is an exact value off the 1/2/5 ladder, a tick marks a gridline, so the two round differently on purpose. Delegating would have made a 1500 bin read `1500`.
- "All 1 people are within their usual range" — a lead with one report is a real scope. Both attention sentences pluralise.
- The members-table caption interpolated an empty scope label, so the table's accessible name began with a dash.
- Average AI cost hardcoded `currency` / `USD` while the tile directly above already read `format` and `unit` off the metric result.
- The cohort label lowercased any dimension label. Harmless for "Division", wrong the moment identity exposes attributes generically (#1881) and a label is "R&D area" — only a plain capitalised word is lowered now.
- The slice sentinel was the bare string `"team"`, so a roster attribute named `team` collided with it: duplicate React key, and picking the real dimension read as "no slice".
- An unrecognised zone from the URL rendered the reports scaffold copy, claiming pending work unrelated to it.
- The compact date-range trigger is icon-only below `sm` and had no accessible name.
- `useViewerIsManager` answered "not a manager" when identity failed, demoting the viewer to an IC shell instead of letting the org zone show the identity error. Unresolved until a tree arrives.
- Two person-id comparisons were case-sensitive while the rest of the app normalises: the route→scope sync re-fired on a casing difference, and `setScope` read a same-root move as a new root and dropped direct-only.
- An empty live group in the pane rendered a heading with nothing under it, which reads as a load failure rather than a filter — `ThemeNav` already skipped it.

**Performance**

- The manager-node walk called `flattenSubordinates` per manager, re-walking that subtree: O(n · depth). One pass returns each subtree size, with the entry pushed before recursing so the picker keeps its outline order — a post-order walk would have reordered it into depth bands, so there is now a test pinning the order across two branches.
- The participation section rebuilt its metrics × roster × buckets scan on every parent render.
- `usePortalPeriod` read localStorage in the render body. The preference is now subscribed through the external store that already existed in `use-period.ts`, and the returned range and callbacks are stable — they land in the dependency arrays of every metric query.

**Tests**

- `section-trend` had no test file at all. Ten cases: the CSS-safe alias remapping on both series and rows, nulls left as gaps, right-axis derivation, stacking, and the pending / error / empty branches.
- `shell-layout` mocked `useViewerIsManager` twice, so the state-controlled factory was dead. Worth noting it was scaffolding rather than a masked failure — nothing in that file sets the flag false, and the IC cases live in `portal-shell` — but the override is gone.
- `domain-lens-view`'s collection mock kept a counter across the whole file, so each caller's slot depended on how many renders earlier tests happened to do.
- The retry spies in `employees-view` and `manage-view` accumulated across their files, so `toHaveBeenCalledOnce` held only while exactly one test triggered a retry.
- `portal-shell` claimed to test a remount while mounting two live instances; it unmounts first now, and its person fixtures are ids.
- `attention-list` minted person ids with a local helper that broke its own UUID shape past nine — it uses the shared `pid`.
- The router fake assigned `search` directly, keeping cleared keys as `undefined` where the real router drops them and `Link` filters them. It routes through `set`.
- The route→scope guard moved to `lib/portal/route-scope-sync` with a reset, so a later test mounting the same person is not silently a no-op. It needed its own module anyway: exporting a helper from a component file breaks fast refresh.

**Not taken**

- The lens readiness filter and the partial-histogram skip, for the reasons in my earlier comment.
- `overview-configs` static metric semantics: filed as #2242 with a three-step plan, and I posted a correction above — the API does carry `description` and `explanation`, and the portal renders neither, which makes the finding bigger than a config cleanup.
- The stale `portal-store` and `team-state` comments were wrong text over correct code; corrected in place.

Verified on the ported tree: `pnpm typecheck` clean, `eslint --max-warnings 0` clean, 810 unit tests pass (823 in the source branch, which includes the Storybook browser suites).

dzarlax:
Both taken care of in `90b9c65b`, one as a fix and one as a documented rule.

**Period preference between tests — fixed.** Right, and it was my own test's leak: `window.localStorage.clear()` does not touch the in-memory store next to it, so the first case's `"month"` became the default every later case inherited when the URL named no period. The preference is reset in `beforeEach` now.

**Tracking every synchronized person — not taken.** Keeping only the latest person is the intended rule, and a `Set` would trade this scenario for a worse one.

Arriving at a person's team is a navigation, not a remount. With a set: a reader opens A's team, re-scopes to something else from the topbar, looks at B's team, then clicks back to A's team — and gets B's roster, because A was already claimed. The address says A and the screen shows B. That disagreement is the thing this whole migration set out to remove.

The guard's job is narrower than "sync each person once ever": absorb re-renders and remounts of the *same* person, so a route sync never fights a scope the reader picked. `A → B → A` re-claiming is that rule working.

Where you are right is that nothing said so and nothing tested it, which is why it reads as an oversight. `src/frontend/src/lib/portal/route-scope-sync.ts` now spells out the reasoning, and there are four cases pinning it — including `A → B → A` explicitly, so anyone who later reaches for a `Set` sees the intent first.

**On the docstring coverage check:** added docstrings where a line of prose carries information — what an absent URL param means, the scope picker's shapes, why a trend point omits a series key instead of zeroing it, the guard and store contracts. That takes the diff from 52% to 62%. I stopped there rather than reaching 80%: the remaining gap is mostly API-client types that mirror the server DTO field-for-field (`PeriodView`, `TimeseriesView`, `MetricBucket`) and upstream code this branch only touched. A docstring restating a symbol's own name is noise that ages into a lie, so I would rather leave the number short than pad it.

`pnpm typecheck` clean, `eslint --max-warnings 0` clean, 814 unit tests pass in the ported tree (827 in the source branch, which includes the Storybook browser suites).

cyberantonz (approved):

PR #2253, opened 2026-08-06
feat(seed): run the demo seeder on Kubernetes stands

### LINKED ISSUE #2243: seed: make deploy/seed runnable on Kubernetes stands (bake into toolbox image)
## Goal

`deploy/seed` (the 25-person demo org + activity seeder) is compose-only. A chart-deployed stand starts empty — every login 403s (`login_denied_unknown_person`, see #1956) and dashboards show "No data". Make the seeder a first-class way to populate a Kubernetes stand.

## Scope

Proven on the InsightTest stand (2026-08-05) by running the seeder on the published `insight-toolbox` image, with the source shipped via ConfigMaps and deps pip-installed at Job start. Productize that:

- Bake the seeder into the toolbox image: `COPY deploy/seed /seed` + `PyMySQL`, `clickhouse-connect` in `src/ingestion/tools/toolbox/Dockerfile` (toolbox is py3.12; the seeder declares `requires-python >= 3.13` — align one or the other; 3.12 is proven to run it).
- Fix the two compose assumptions that broke on a chart stand:
  - `MARIADB_ANALYTICS_DB` defaults to `analytics`, but the chart's analytics service keeps its catalogue tables (`metric_definitions`, `metric_catalog`, …) in `mariadb.database` — the default should follow the chart convention or be documented as required.
  - `TENANT_DEFAULT_ID` defaults to the compose tenant (`00000000-df51-5b42-9538-d2b56b7ee953`); rows seeded under it are invisible to every login on a stand. Fail loudly or require it explicitly.
- Ship a reference Job manifest (env from the composed `insight-db-creds` Secret; `DEV_USER_EMAIL` anchors the dev-lead persona).

## Acceptance criteria

- [ ] A single `kubectl apply` Job manifest using the published toolbox image runs `seed.py all` successfully on a chart-deployed stand — no ConfigMaps, no runtime pip.
- [ ] Login as `DEV_USER_EMAIL` lands on a populated dashboard.
- [ ] Wrong/missing `TENANT_DEFAULT_ID` fails loudly instead of seeding invisible data.

ktursunov:
Closes #2243.

`deploy/seed` was compose-only, so a chart-deployed stand starts empty: every login is refused and every dashboard reads "No data". This makes populating one a single command.

```bash
./src/ingestion/tools/seed/seed-stand.sh -n <namespace> --email you@example.com
```

## What changed

**The seeder moved into the ingestion tree** — `src/ingestion/tools/seed`. The silver step shells out to the ingestion tree's own DDL and gold-build scripts, so living in that tree is what lets one image carry both at one version, with no chance of the seeder and the migration SQL drifting apart.

**It is published as its own image, `insight-seed`** — deliberately not the toolbox. The toolbox runs migrations against real stands, and a demo-data generator has no business being installed there. Same build context, a narrower COPY set, `insight-seed` as the entry point; the toolbox excludes `tools/seed/` through a per-Dockerfile `Dockerfile.dockerignore`, so both build from the same context and get different content. CI publishes it beside the toolbox (per-arch matrix, digest merge, provenance attestation), the nightly Trivy matrix scans it, and compose builds the same Dockerfile locally. The amd64 leg validates before pushing: the console script resolves, every step module imports (the whole driver set — exactly what was missing when the seeder could not run on a stand at all), the package's tests pass, and dbt is present for the silver step's subprocess. No chart template — nothing in a deployed release runs the seeder.

**The tree is a package plus its tests** — `insight_seed/` and `tests/`. Every caller installs it and invokes a program rather than a module in a directory: `insight-seed <step>` seeds, `insight-seed-realm` generates the compose Keycloak realm from the same roster. `manifest.json` and `PROFILE.md` stay at the tool root, where their readers already name them.

**The environment contract fails loudly.** `TENANT_DEFAULT_ID` and `MARIADB_ANALYTICS_DB` lose their defaults — rows under the wrong tenant are invisible to every login while the run still reports success, and which database holds the catalogue tables is a per-stand fact (compose uses a database of its own; a chart stand keeps them in `mariadb.database`). A new `preflight` module answers every answerable question *before* anything is written and reports the whole list at once.

**It refuses rather than damages.** The identity step is additive and marks its rows, so a tenant already holding other rows is refused. The silver step is *not* additive — it `TRUNCATE`s every table it writes, across all tenants — so a stand holding another tenant's rows anywhere in that reset surface is refused too. The reset surface is one registry that `truncate` itself enforces, kept in step with the generator call sites by a test. `SEED_FORCE=1` overrides either, deliberately.

**`seed-stand.sh` discovers, renders, applies, follows.** Coordinates come from the stand's own `<release>-platform` ConfigMap and its Secrets; the image from `ingestion.seedImage`, which CI pins on publish. It is published state, not deployed state — the chart never renders it; it is there so an operator can ask a stand which seeder matches it. Credentials never pass through the shell — the Job references the release's Secret by key and runs as the *application* MariaDB user, which the umbrella already grants everything the seed writes. It is a plain Job, not a chart hook, so a failed seed never gates or rolls back a release. `--dry-run` prints `seed-job.yaml.tpl` rendered — the reference manifest and the one applied are the same file.

Along the way: the second-tenant fixture is gated (it aborts identity-resolution's scheduled projection on a cluster and is only wanted on compose, where the stand suite asserts against it); the activity window is read in one place so the manifest cannot report a window the rows do not sit in; identity re-runs are idempotent again (`seed_persons`/`seed_person_names` relied on an `INSERT IGNORE` collision that migration 004 removed, so each run appended duplicate email and name observations); and host build artefacts are kept out of both images.

## Acceptance criteria

- [x] **A Job on the published seeder image seeds a chart-deployed stand** — no ConfigMaps of source, no runtime `pip`. `--dry-run` gives the manifest; the script applies it.
- [x] **A wrong or missing tenant fails loudly** — required and UUID-validated, with the analytics database *verified* to hold `metric_definitions` rather than assumed, and the database it looked in named in the error.
- [ ] **Login lands on a populated dashboard** — restated, because the seeder cannot satisfy it alone: the authenticator resolves people by the email claim, so a user with `--email` must already exist in the realm. Documented as a prerequisite in the seeder README, `HELM_DEPLOY.md` (new Step 7) and `CONTRIBUTING.md`. Generating the realm from the seeder's own roster is follow-up work, not part of this change.

## Verification

Run against a local single-node cluster with the umbrella chart deployed and external MariaDB/ClickHouse, using an image built from this branch:

- discovery resolves every value from two flags; the rendered Job passes `kubectl apply --dry-run=client`;
- `--step identity` completes in seconds (26 people, their login rows, names, org-chart edges and account maps); re-running it inserts nothing;
- `--step analytics` writes the tenant-scoped catalogue row — which is what proves the discovered analytics database is the right one;
- both refusals fire on a stand that already holds data, before anything is written;
- `--step silver` end to end against a throwaway ClickHouse of the pinned version: placeholders → generators → migrations → `dbt run --select tag:gold`, 29,306 rows across 19 silver tables, `PASS=18 WARN=0 ERROR=0`; seeding twice in a row works.

Locally: `ruff check`/`ruff format --check` clean, `mypy` clean, 39 unit tests (which CI now runs inside the published image), `PROFILE.md` regenerated and its staleness check green, `docker compose --profile seed config` and the `seed-sample` image build, and the realm generator still builds its 27 users from the shared roster.

The compose path is unchanged in behaviour: `docker-compose.yml` now passes `MARIADB_ANALYTICS_DB` and `SEED_CROSS_TENANT_FIXTURE` explicitly, which it previously inherited from the code defaults this PR removes.

## Note for reviewers

The diff is large because a relocation cannot be split from the rewiring that follows it — every caller has to move in the same commit or the tree is broken in between. The substance is in `src/ingestion/tools/seed/`: `config.py`, `preflight.py`, `seed-stand.sh` and `seed-job.yaml.tpl` are new; everything else under that path is the moved package.

One gap left open on purpose: the package is not registered in `scripts/ci/components.py`, so it is outside the per-component coverage gate. Its suite covers the pure half — the environment contract, the SQL each guard issues, the messages a refusal carries — and the I/O half it deliberately does not touch would put the component well under the 80% line. The tests still run on every ingestion change, in the image build. Raising them to the gate's bar is worth doing separately.

## Summary by CodeRabbit

* **New Features**
  * Added a packaged sample-data seeder for compose and Kubernetes test stands.
  * Added one-command stand seeding with dry-run, step selection, validation, safety checks, and log following.
  * Added generated manifest support and optional cross-tenant fixtures.
  * Added dedicated seeder and multi-architecture image publishing.

* **Bug Fixes**
  * Improved duplicate-prevention and destructive-operation safeguards during seeding.
  * Added preflight checks for configuration, database connectivity, required tables, and existing data.

* **Documentation**
  * Updated setup, deployment, testing, and image documentation for the new seeding workflow.


cyberantonz (approved):

ktursunov:
### `secrets (diff)` is red on a detector false positive — GitHub Action pins

Recording the analysis here rather than changing the scan, as the job's own guidance asks.

**What is flagged.** Two 40-hex strings in `.github/workflows/build-images.yml`, `Github` detector, `Verified: false`, surfaced through the HTML decoder:

| value | what it is |
|---|---|
| `263435318d21b8e681c14492fe198d362a7d2c83` | `docker/build-push-action` commit pin |
| `c299e40c65443455700f0fdfc63efafe5b349051` | `docker/metadata-action` commit pin |

A 40-character hex string is also the shape of a legacy PAT, which is what the detector matches. These are action pins, published in every workflow that uses those actions.

**They are not new here.** `main` carries 9 and 8 occurrences of these exact SHAs in the same file today. This PR adds three more, in the `seed` and `merge-seed` jobs, pinning the same actions at the same versions as the `toolbox` / `merge-toolbox` pair they mirror. Scanning the file in isolation, `main`'s copy yields **5** detector hits and this branch's yields **4** — the file is no worse than the one already on `main`. The diff scan only reads added lines, so `main`'s copies are never re-examined and the new ones are.

**Why it went red only after the last rebase.** Identical content passed this check on the three earlier pushes. Those scans recorded `Misses: 0`, `VerificationTimeSpentMS: 0` over ~190 KB; the failing ones record `Misses: 2` over ~420 KB. The merge base changed what the scan covers — roughly double the bytes — and the larger scan reaches the pins. Re-running the job reproduced the failure exactly (`verified_secrets: 0, unverified_secrets: 2` both times), so it is deterministic, not flaky.

**No credential is involved**, so there is nothing to revoke or rotate. Suppressing it means editing `.github/workflows/trufflehog.yml`, whose header warns against widening the filter — repo-wide scan policy, and out of scope for this PR. Raising it separately.

Every other check is green: 32 pass, 26 skip, including the required `Run E2E suite`.

hello1101n (approved):

Issue #2213, opened 2026-08-05
[epic] Semantic Layer


cyberantonz:
# Semantic Layer

Carved out of the presentation-layer split (#1803) so that epic can finish the **division** (engineering contract vs. presentation, safety, tenant filter, FE experiment loop) without being blocked on a much larger metric-semantics rebuild.

## Why separate

The FE experiment loop and the division need **nothing** from the semantic layer: experiments run in the dev cluster against real data through the saved-query API and the existing metric endpoints (all shipped in #1803 Phase A). The semantic layer only changes *how metrics are authored* — a typed palette instead of hand-written SQL — and is an enhancement, not a prerequisite.

## What it is

Definitions-as-data with one compiler over datasets: datasets/measures/metrics authored as data (product-seeded, runtime-editable), a single compiler turning definition + request into warehouse SQL, capability as a projection of definitions (not stored rows), and server-injected tenancy/org-scope/cohort scopes at one choke point.

Governing specs: `docs/domain/semantic-layer/specs/` (DESIGN, PRD, REFERENCE, IMPLEMENTATION, FINDINGS, ADR-001 `cpt-semantic-layer-adr-adopt-compiler-over-datasets`).

## Phase 1 — Definition Core (#2208) slices

1. Schema (domain-shaped tables) — done (merged, PR #2210).
2. Field catalog — first attempt (PR #2211) reinvented a schema snapshot the repo already produces; **needs redesign** to source types/engine from the committed `connectors-ddl` DDL snapshot instead. Tracked as a subtask.
3. Validators (expression allowlist + filter-tree types).
4. Authoring + store wiring (product YAML seed → reconcile).
5. Extraction (transcribe measures onto datasets; verify roles).
6. Structural validation (CI vs the DDL artifact at the same commit; runtime probe vs live warehouse).

## Later phases

Phase 2 compiler + cutover (#1976), FE catalog/discovery rework (#1977 semantic half), promotion ladder (#1978 semantic form). #1974 (registry) and #1975 (passports) landed as the first definitions-as-data steps.

Prioritized **after** #1803's division work is finished.

Issue #1965, opened 2026-07-28
[pres] Saved-query CRUD (presentation.queries)


cyberantonz:
Phase A. `GET/POST/PUT/DELETE /v1/queries` + `/v1/queries/{id}/run`. Parent: #1803.

PR #2260, opened 2026-08-06
feat(analytics): saved-query export/import — promote experiment queries dev->prod (#2259)

### LINKED ISSUE #2259: [pres] Saved-query export/import — promote experiment queries dev->prod
## Why

Tier-3 experiments (the A/B loop under #1803) are built on a dev stand: a bespoke frontend plus new **saved queries** authored against that stand's data. The frontend promotes to production by PR into the monorepo (`src/frontend`), but the saved queries it depends on stay stranded in the dev stand's database — there is no way to carry them across stands.

Export/import lets those queries travel dev -> prod alongside the frontend, so a promoted experiment arrives with the queries it needs.

## Mechanism

Two thin endpoints over the existing saved-query CRUD (#1965):

- `GET /v1/queries/export` -> JSON dump of the calling tenant's saved queries as `[{ name, description, sql }]`.
- `POST /v1/queries/import` -> bulk-create from that JSON. Each `sql` is **re-validated** through the single-SELECT gate (`validate_single_select`) on import.

Portable by construction, no SQL rewriting:

- SQL is contract-relative (`silver.*` / gold), identical across stands.
- Tenant is session-injected at run time (the `{tenant}` parameter), never stored — so import re-homes each query to the importing session's tenant. Import drops any source id/tenant, generates a fresh id, and binds the importing session's tenant. No cross-tenant leakage.
- Name-collision policy chosen and documented in the implementation.

No schema change — reuses the existing `saved_queries` table.

## Out of scope

- Curated/product queries auto-seeded to every stand. That is definitions-as-data and belongs to the semantic epic #2213, not a bespoke query seeder.
- A/B rules, traffic routing, variant assignment.

Part of #1803.

cyberantonz:
## What

Adds saved-query **export/import** to the analytics service — two thin endpoints over the existing saved-query CRUD (#1965):

- `GET /v1/queries/export` -> the calling tenant's saved queries as a portable JSON document (`{ queries: [{ name, description, sql }] }`), carrying no id, tenant, or timestamps.
- `POST /v1/queries/import` -> bulk-creates from that document; returns `{ imported, skipped }`.

## Why

A tier-3 experiment is built on a dev stand: a bespoke frontend plus new saved queries authored against that stand's data. The frontend promotes to production by PR into `src/frontend`, but the saved queries it depends on are otherwise stranded on the dev stand's database. Export/import lets those queries travel dev -> prod alongside the frontend.

## Mechanism

Portable by construction — no SQL rewriting:

- SQL is contract-relative (`silver.*` / gold), identical across stands.
- Tenant is session-injected at run time (the `{tenant}` parameter), never stored. Import drops any source id/tenant, generates a fresh id, and binds the importing session's tenant, so an import can never write cross-tenant.
- Import **re-validates** every SQL through the single-SELECT gate (`validate_single_select`) — one bad statement rejects the whole document as a 400.
- **Collision policy:** a row whose `name` already exists for the tenant is **skipped**, never overwritten, so a re-import is idempotent.

No schema change — reuses the existing `saved_queries` table.

## Out of scope

- Curated/product queries auto-seeded to every stand (definitions-as-data, semantic epic #2213), not a bespoke query seeder.
- A/B rules, traffic routing, variant assignment.

## Specs & tests

- Extends the presentation PRD/DESIGN (export/import FR + the two endpoints under API Contracts); `cfs validate` is 0 errors per artifact.
- Regenerates the analytics OpenAPI contract; drift gate green.
- Rust unit tests for the collision/dedup policy; stand endpoint-contract tests for export 200, import 201 + idempotent re-import, and import re-gating a non-read SQL, plus the coverage-gate wiring for the two new operations.

Closes #2259
Part of #1803

## Summary by CodeRabbit

* **New Features**
  * Added authenticated endpoints to export saved queries as portable JSON and import them into another tenant.
  * Imported queries receive fresh identifiers and are revalidated for safe SQL.
  * Duplicate or same-name queries are skipped without creating conflicts.
  * Import responses report the number of imported and skipped queries.
* **Documentation**
  * Updated API documentation, requirements, and acceptance criteria for saved-query portability.


cyberantonz:
Thanks @coderabbitai — went through all six. Addressed in a006d64f:

**Applied**

- **Bound import size** (major): import now rejects a document over `MAX_IMPORT_QUERIES` (1000) with a 400 on the `queries` field, at the route boundary before any load or write. Export is bounded by construction (only the tenant's own analyst-authored queries; never truncated), so I did not cap it — silently dropping queries would defeat a promotion tool.
- **PRD envelope** (minor): PRD FR now uses `{ "queries": [...] }` to match the OpenAPI contract and DESIGN.

**Documented rather than changed**

- **Atomic collision handling** (major): the saved-query surface has no `(insight_tenant_id, name)` uniqueness invariant — `create` already permits duplicate names — and adding one is a schema change, which is explicitly out of scope for this PR. The name-skip is therefore best-effort: two concurrent imports of the same new name race exactly as two concurrent `create`s do, and it is never a cross-tenant concern. DESIGN now states this explicitly instead of implying atomicity.

**Skipped, with reasons**

- **Document 415/422 on import** (minor): `openapi.json` is generated from the handlers via `.standard_errors` and drift-gated in `.github/workflows/openapi-specs.yml`; hand-editing it fails the gate. The 415/422 under-declaration is a known, surface-wide condition (#1670) shared by every analytics route (e.g. `POST /v1/queries`), not specific to import — the stand suite already pins the real 415/422 behaviour. Fixing it on one route by hand-editing generated output would be both inconsistent and immediately reverted by the gate.
- **Move import policy to a domain module** (major): the sibling handlers in this file (`create`/`update`/`run`) all keep their validate-build-persist logic inline; the pure pieces (`select_new_queries`, `new_query_row`) are already extracted. A standalone domain module for `import` alone would make the file inconsistent for no correctness gain.
- **Remove commentary / `///` docs** (major): these match the file's established style — the pre-existing private helpers and tests in this same module (`parse_json_each_row`, `missing_param_name`, `classify_run_error`, and every test) already use `///` docs and explanatory `//` "why" comments. Selectively stripping only the new ones would make the module inconsistent.

aleksdotbar (approved):

cyberantonz:
Closing — the design moved. Migrating saved queries doesn't integrate anything into the product FE; the FE renders **metrics** (by `metric_key`), so the real path is query→**custom metric**. The schema already scaffolds it (`source_kind='custom_observation_sql'`, `origin='custom'`, and the YAML reconcile's disable-missing is scoped to `origin='builtin'` so custom rows survive). Reworking as custom-metric CRUD + export/import (keyed on `metric_key`) + a `/metrics` FE surface, with `custom_observation_sql` wired into the compiler so a metric can carry multi-table custom SQL. Continues in the same branch.

PR #2285, opened 2026-08-06
feat(identity-resolution): govern discovered person attributes

### LINKED ISSUE #2284: Discovered person attributes have no tenant-governed policy
## Problem

Attributes discovered in source systems arrive with no tenant-governed
meaning attached to them.

Nothing records which discovered fields exist, what a tenant calls them, or
whether a field may be used to group people, to compare a person against
peers, or neither. Nothing records that a field is sensitive. A field that
has been retired stays indistinguishable from one that is current.

Without that governance layer, either every discovered field is usable for
everything the moment it lands — including fields whose use in a peer
comparison would be a disclosure problem — or none is usable at all and each
new field waits for a code change.

The decisions are administrative and reversible, so they need the properties
configuration changes are held to: an actor recorded against every change, a
reason, and a history that later readers can inspect rather than a value that
was silently overwritten. Consumers evaluating a request need to read the
state that applied, not guess at it.

## Scope

The record of which person attributes exist per tenant and source, the
tenant-curated policy governing each one, and the administrative surface for
reading and revising that policy.

Publishing the policy to the analytical store, the attribute catalog, and the
grouping and comparison APIs are tracked separately under the parent epic.

aleksdotbar:
## Problem

Discovered person attributes have no governed home. Analytics needs
tenant-curated policy per attribute — label, sensitivity, whether it may
group or drive comparison — with an audit trail, and a registry that learns
new fields as connectors deliver them. Nothing owns that today.

## Change

Person-attribute registry in the identity-resolution service.

- `person_attribute_definitions` — one row per discovered
  (tenant, source type, source instance, field), keyed by the raw warehouse
  strings so the policy snapshot joins claims byte-equal.
- `person_attribute_policy_revisions` — append-only, actor and reason on
  every row; the revisions are the audit trail. Defaults: grouping allowed,
  comparison denied until an admin enables it.
- `reconcile-attributes` subcommand and CronJob register discovered fields
  (advisory lock, operations journal). Guards refuse — exit 3, journalled —
  when the claims relation is absent (deployable ahead of the ingestion
  release) or when a run would register nothing.
- Admin API: `GET /v1/person-attributes[/{id}]`, `PUT …/{id}/policy`
  (optimistic concurrency, stale revision → 409), journal GETs under
  `/v1/person-attributes-reconcile`. All behind the existing admin gate.
- `openapi` subcommand and CI drift job; the committed spec was
  hand-maintained and stale, now generated from the route table.

## Out of scope

Publishing policy and assignment snapshots to the analytical store (next PR),
the attribute catalog, grouping and comparison APIs, named groups,
manager-source authority.

## Validation

- 98 unit tests; clippy pedantic clean; helm render-contract 23 green.
- Migrations and the exact repository SQL exercised against MariaDB 11.4:
  the guarded self-referencing INSERT is legal, a stale revision inserts zero
  rows, a race loser hits the unique key, the CHECK fires, and a refresh keeps
  `first_observed_at` while advancing `last_observed_at`.
- OpenAPI drift gate passes on the regenerated spec.

## Notes

- `first_observed_at` records when a field was first registered, not the
  earliest claim ever observed for it.
- A warehouse tenant string that is not a canonical UUID registers normally
  but cannot be addressed through the admin API; the reconcile summary counts
  those so the mismatch is visible rather than silent.

Closes constructorfabric/insight#2284

### DIFF
--- a/.github/workflows/openapi-specs.yml
+++ b/.github/workflows/openapi-specs.yml
@@ -94,6 +94,44 @@ jobs:
           if [ "$rc" -ne 0 ]; then echo "::error::Authenticator OpenAPI spec drift — $AUTHENTICATOR_SPEC is stale. Regenerate: (cd src/backend && cargo run -p authenticator --bin authenticator -- openapi) > dump.json && python3 scripts/ci/openapi_spec.py update --file $AUTHENTICATOR_SPEC --live-file dump.json"; fi
           exit "$rc"
 
+  # ─── identity-resolution — offline emit via the `openapi` subcommand ──────
+  identity-resolution:
+    name: Identity-resolution OpenAPI spec drift gate
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+        with:
+          persist-credentials: false
+      - name: Install protoc
+        # protobuf-compiler: grpc-hub -> prost-build (same host gear set as
+        # the analytics service).
+        run: sudo apt-get update && sudo apt-get install -y protobuf-compiler
+      - uses: dtolnay/rust-toolchain@4cda84d5c5c54efe2404f9d843567869ab1699d4 # stable
+      - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
+        with:
+          workspaces: src/backend -> target
+          key: identity-resolution-openapi
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
+        with:
+          python-version: "3.12"
+      - name: Generate the OpenAPI doc offline and drift-check it
+        env:
+          IDENTITY_RESOLUTION_SPEC: docs/components/backend/identity-resolution/openapi.json
+        run: |
+          # `identity-resolution openapi` builds offline and prints the doc; a
+          # build failure here should abort loudly under the runner's `set -e`.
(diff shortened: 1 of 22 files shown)


dzarlax:
Building the consumer side of this in #2335 (the portal's cohort control), and
I would rather match your contract than guess at it.

Today the portal discovers which comparisons it can offer by walking the
viewer's roster and keeping attributes that take more than one value. That can
only work for a viewer who can see other people, and identity serves a viewer
their own subtree — so a viewer with no reports has a roster of one person, no
attribute qualifies, and no comparison can be offered at all. #2335 replaces
that with a catalog the server supplies, falling back to the roster until one
exists, so the change is additive on both sides.

`DESIGN.md` §3.3 says `GET /v1/metric-definitions` gains "the active
allowed-attribute catalog", but does not pin the shape, so the client reads an
optional `comparison_attributes: [{ id, label }]` as a placeholder — in one
function, so aligning it is a one-line change.

Three things would let me replace the guess:

1. **Field name and entry shape.** Is each entry the attribute `id` plus a
   resolved display label, or the whole `PersonAttributeResponse` this PR
   defines?
2. **The label.** `PolicyResponse.label_override` is optional. Does the catalog
   resolve the display name (falling back to `source_field_id`), or is that the
   client's job? I would rather not have two places deciding what an attribute
   is called.
3. **Filtering.** Is the catalog already limited to `comparison_enabled` and
   not `retired`, or does the client apply the policy itself? Enforcing it
   server-side and having the client render what it is given seems the safer
   split, given the same flag has to gate the query anyway.

No rush from my side — the portal behaves correctly with the field absent, so
nothing here blocks either PR.

PR #2287, opened 2026-08-06
feat(bamboohr): rewrite as CDK source collecting all employee fields

### LINKED ISSUE #2270: BambooHR: collect all custom fields without per-field configuration
## Problem

BambooHR extracts only the employee fields enumerated in the connection
configuration. Custom fields the customer defines in BambooHR are invisible
until an operator discovers their aliases and adds them to the configuration,
and the same list must be mirrored in the ingestion tracking configuration for
changes to be versioned and claimed. A new custom field therefore requires
two coordinated manual updates before it can appear anywhere downstream, and
a mismatch between the two lists silently produces attribute claims with
wrong effective dates or none at all.

The connector is a declarative manifest, and the employee request enumerates
fields from static configuration. BambooHR has no request form that returns
all fields; the field list must be built per sync from the field-metadata
endpoint, which a declarative manifest cannot express. Collecting everything
therefore requires reworking the connector as a Python CDK source, for which
the active-directory connector is the in-repo precedent.

## Outcome

- Every employee field, standard and custom, is extracted on every sync with
  no per-field configuration.
- The bronze contract is preserved: same table, same standard columns, same
  unique_key formula, full record in the raw payload, so existing staging,
  identity, and class models keep working unchanged.
- Change detection and attribute claims cover the full raw payload. These land
  together with fetch-all extraction as one unit: fetch-all without
  raw-payload change detection makes custom-field changes invisible, and
  dynamic claims without it stamp wrong effective dates.
- The custom-fields configuration knobs are retired.

## Notes

The pairing rule applies to every connector, not just this one: a
customer-defined field may only be consumed downstream once change detection
covers the whole payload. Consuming a field that change detection does not
cover is worse than ignoring it, because the value is then picked up carrying
the timestamp of some later change and appears to have taken effect at a time
it did not.

Workday needs no extraction change of its own (report-defined columns already
pass through) but does need the same versioning change before its custom
columns can be consumed. ms-entra and active-directory need both, as separate
work, when their extraction broadens.

No consumer of customer-defined fields exists today; #2262 deliberately
claims curated source fields only, for the reason above.

## Acceptance criteria

Proposed — this issue carried Problem, Outcome and Notes rather than criteria. One criterion per distinct oracle, cut from the Outcome bullets. The scenarios below cite them.

- **AC-1.** Every employee field the source offers, standard and customer-defined, is collected on every sync, with no per-field configuration anywhere: the field list is built per sync from the source's own field metadata, and the connector's configuration carries no field list.
- **AC-2.** A declared bronze column is always requested, so a gap in the source's field metadata cannot silently empty a column.
- **AC-3.** The bronze contract is preserved: the same table, the same standard columns, the same unique-key formula, and the whole collected record in the raw payload.
- **AC-4.** No sensitive standard field is requested or stored, under any vocabulary the source offers — alias, alias-less id, or indexed id form.
- **AC-5.** A field list too long for one request is split, and the split is invisible downstream: one record per employee, every batch merged into the payload.
- **AC-6.** Change detection covers the whole raw payload rather than a named field list: an edit to any collected field, including a customer-defined one, produces a new version, and an edit to nothing but the source's own change stamp produces none.
- **AC-7.** Field history covers every key in the raw payload, so a customer-defined field's changes are versioned with the effective dates of the change that carried them.
- **AC-8.** A declared column the source's answer omits lands as null and the sync continues, rather than failing the run or being read as data loss.

## Testing

The connector was rewritten from a declarative manifest to a Python CDK source so it can ask the source what fields exist and then request all of them. The risk is a silent one: a field that stops being collected, or a customer-defined field whose changes are versioned against the wrong date because change detection still reads a curated list. Reliability leads, and the sensitive-field guard is the Security half — a fetch-all connector is exactly where a never-collect field arrives by accident. Checks run in the connector's own pytest suite (132 tests, 98% coverage on the source package per PR #2287) and the dbt layer; the change-detection and history macros are exercised through the models this connector is the only user of.

Two facts are stated rather than tested around. Change detection covers the whole payload, but the attribute claims built on it still name a curated list of eight fields — so a customer-defined field is *versioned* and not yet *claimed*, which is #2262's scope and the Notes' own pairing rule. And the sensitive-field denylist covers standard aliases only; a customer-defined field named like a sensitive one is collected in full, by design.

All 8 acceptance criteria covered: AC-1 → 1, 2 · AC-2 → 3 · AC-3 → 4, 5 · AC-4 → 9, 10 · AC-5 → 6 · AC-6 → 7 · AC-7 → 8 · AC-8 → 11.

- [x] 1. **Field list built per sync** — Reliability · connector-tests · AC-1 — read the source's field metadata and build the request set, with fields carrying an alias, no alias, a deprecation flag, and an entry that is not an object → each is requested by alias or by id, deprecated and unusable entries are skipped, and metadata that is not a list is an error rather than an empty sync → `src/ingestion/connectors/hr-directory/bamboohr/tests/test_fields_discovery.py::TestRequestKeys`.
- [x] 2. **No per-field configuration** — Reliability · connector-tests · AC-1 — read the shipped connection specification → it carries the five connection properties only, and the retired custom-fields knob is absent → `.../tests/test_source.py::TestWiring::test_the_spec_is_the_shipped_connection_specification`.
- [x] 3. **Declared columns always requested** — Reliability · connector-tests · AC-2 — build the request set from metadata that omits a declared bronze column → every declared column is requested anyway, each field once however often it is named, and customer-defined fields follow the declared ones → `.../tests/test_fields_discovery.py::TestRequestedSet`.
- [x] 4. **Bronze record shape** — Reliability · connector-tests · AC-3 — emit a record from a report row carrying fields inside and outside the declared columns → the record carries exactly the declared columns, an undeclared field stays out of the top level, and every returned field is preserved in the raw payload with a key order that does not follow the API → `.../tests/test_streams.py::TestEmployeeRecords::test_the_record_carries_exactly_the_declared_columns`, `::test_a_field_outside_the_declared_columns_stays_out_of_the_top_level`, `::test_every_returned_field_is_preserved_in_the_raw_payload`, `::test_the_raw_payload_key_order_does_not_follow_the_api`.
- [x] 5. **Unique key unchanged** — Reliability · connector-tests · AC-3 — emit records for employees with and without a usable id → the key is tenant, source and employee id, and a row without a usable id is skipped rather than keyed wrongly → `.../tests/test_streams.py::TestEmployeeRecords::test_the_unique_key_is_tenant_source_and_employee_id`, `::test_a_row_without_a_usable_id_is_skipped`.
- [x] 6. **Split request invisible** — Reliability · connector-tests · AC-5 — request a field list within the limit and one over it → the first is one request, the second splits into accepted requests, and the split still yields one record per employee with every batch merged into the payload → `.../tests/test_streams.py::TestEmployeeReportRequest`.
- [ ] 7. **Version on a real edit only** — Reliability · dbt-tests · AC-6 — build the snapshot over an unchanged payload, a payload whose only difference is the source's change stamp, and a payload with one customer-defined field edited → no new version for the first two, exactly one for the third. **Not covered today**: `bamboohr__employees_snapshot.sql` is the only user of the whole-payload snapshot macro, and neither the macro nor the model has a test; PR #2287 verified this by hand on a local ClickHouse.
- [ ] 8. **History over every key** — Reliability · dbt-tests · AC-7 — build the field history over a payload gaining a new customer-defined field and changing an existing one → both appear as history rows dated by the change that carried them, and a key absent from one version reads as empty rather than dropping the row. **Not covered today**: the history macro has no test either, and the model was reshaped twice after this issue closed (#2495 for an out-of-memory failure on real data, #2512 for column drift).
- [x] 9. **Sensitive fields never requested** — Security · connector-tests · AC-4 — build the request set with the source offering every sensitive standard alias → none is requested, no declared bronze column is treated as sensitive, and a sensitive alias does not block discovery of the rest → `.../tests/test_fields_discovery.py::TestSensitiveFields`.
- [x] 10. **Sensitive fields never stored** — Security · connector-tests · AC-4 — return a sensitive field from the report anyway, under its alias, its alias-less id and its indexed id form → none is stored, in a named column or inside the raw payload → `.../tests/test_streams.py::TestEmployeeRecords::test_a_sensitive_field_the_report_returned_anyway_is_not_stored`, `::test_a_sensitive_field_volunteered_under_its_id_is_not_stored`.
- [x] 11. **Omission tolerated, not hidden** — Reliability · connector-tests · AC-8 — withhold a declared bronze column, withhold a custom field, answer one under an indexed key, and return a report declaring no columns at all → the first three land as null and the sync continues, the indexed answer is not counted an omission, and the empty report is not read as data loss → `.../tests/test_streams.py::TestSilentlyOmittedFields`, `::TestEmployeeRecords::test_a_declared_column_the_report_omitted_reads_as_null`.

Versatility — n/a: this is one source's extraction path. The same pairing rule applies to Workday, ms-entra and active-directory, but each is separate work with its own scope, as the Notes say.
Performance — n/a: a nightly batch collection adds no user-facing query path.
Efficiency — n/a: no run-cost lane is wired in this repo. Worth knowing that fetch-all did move the cost: the history model ran out of memory on real data after this closed (#2495), which a run-cost lane would have caught before the stand did.

aleksdotbar:
## Problem

BambooHR extracted only the fields listed in `bamboohr_employees_custom_fields`,
and the same list had to be mirrored in the dbt tracking config. A new custom
field needed two coordinated edits; a mismatch silently produced attribute claims
with wrong effective dates or none at all.

BambooHR has no request form that returns every field — the list must be built
per sync from `meta/fields`, which a declarative manifest cannot express.

## Fix

- BambooHR becomes a Python CDK source. Each sync reads `meta/fields` and requests
  every non-deprecated field; the declared bronze columns are always requested so
  a metadata gap cannot empty a column.
- The snapshot hashes the whole `raw_data` payload (order-insensitive, `lastChanged`
  excluded); a new `fields_history_raw` macro diffs it without naming fields. Both
  read the payload through one shared helper so they cannot disagree about a value.
- Sensitive standard fields are never requested and never stored.
- Both custom-field knobs retired.

`snapshot` gains two default-off arguments; existing callers compile byte-identical,
so no other connector re-hashes.

## Collection policy

`SENSITIVE_FIELDS` excludes government identifiers, protected demographics,
personal contact details, street address, photos, social profiles and compensation
amounts — the categories PRD §4.2 puts out of scope, enforced in the connector
because discovery is otherwise exhaustive.

Two things reviewers should weigh: the list contents are a judgment call, and it
covers **standard BambooHR aliases only**. A customer-defined field named e.g.
"Salary band" is still collected in full, because BambooHR's field metadata carries
no classification to filter on.

## Verification

- Bronze DDL byte-identical to the committed snapshot — verified by running the
  real gate path (build → `discover` → `destination-clickhouse write` →
  `bronze_promoted`) and diffing `SHOW CREATE TABLE` for all three tables.
- dbt models run against ClickHouse 25.7.5: a `lastChanged`-only edit produces no
  new version; a real edit produces one version plus history rows for the changed
  custom field and any newly appearing field.
- Hash-algorithm cutover simulated: one extra version, first-observation history
  rows only, steady state on the next run. No spurious change rows, so no wrong
  effective dates.
- 132 unit tests, 98% coverage on `source_bamboohr`.

## Notes

`images.cdk.image` is empty until the first main build pins it.

Workday keeps the equivalent `workday_custom_fields` var — out of scope here, and
its report columns already pass through as bronze columns.

Closes constructorfabric/insight#2270

## Summary by CodeRabbit

- **New Features**
  - Introduced a redesigned BambooHR connector with API-key authentication and configurable start dates.
  - Added employee, leave request, and field metadata synchronization.
  - Dynamically discovers employee fields, supports large reports, filters sensitive data, and preserves complete raw payloads.
  - Added comprehensive change tracking for dynamically discovered fields.
  - Added automatic migration support when connectors change from manifest-based to CDK-based definitions, preserving connection state.

- **Documentation**
  - Updated BambooHR setup, usage, deployment, field-handling, and connector classification guidance.

- **Refactor**
  - Migrated BambooHR from a manifest-based connector to a Python CDK connector.


cyberantonz (approved):

mitasovr (approved):

mitasovr (approved):

mitasovr (approved):
Looks ok

Issue #2224, opened 2026-08-05
Git proxy service: clone-based extraction for git connectors (nocode + git-proxy hybrid)


mitasovr:
# Concept

Our git connectors (github-v2, gitlab, bitbucket-cloud) are CDK-based and extract commit-level data via vendor REST/GraphQL APIs. The heaviest streams make **one API call per commit** (`file_changes`: GitHub `GET /repos/{o}/{r}/commits/{sha}`, Bitbucket `diffstat/{sha}`), so every backfill and every large sync runs into vendor rate limits (Bitbucket ~1000 req/h is the worst case: a 50k-commit repo ≈ 50 hours just for diffstats).

Everything those streams need is derivable from a git clone in minutes. The proposal: a single **git proxy service** shared by all git vendors + thin **nocode (declarative) connectors** per vendor.

## Architecture

**Per-vendor nocode connector** (replaces the CDK connector):
- `repositories` (and PR streams: `pull_requests`, comments, commits, reviewers) → vendor API directly, as today. `repositories` is incremental on the vendor's "last pushed" field (GitHub `pushed_at`, GitLab `last_activity_at`, Bitbucket `updated_on`).
- `commits`, `file_changes`, `branches` → the git proxy, with `repositories` as parent stream (`SubstreamPartitionRouter` + `incremental_dependency: true`, so only repos actually updated since the last sync are visited). Declarative CDK supports per-stream `url_base`, so mixing vendor API and proxy in one manifest is standard.

**Git proxy service** (one deployment, vendor-agnostic, likely Go/Rust/Python + git CLI):
- Stateless with respect to sync state and credentials: sync cursors live in Airbyte state; git credentials arrive per request in headers and are never persisted. Disk content is a rebuildable cache — the proxy may lose any repo at any time and re-clone on demand.
- Repos are **bare blobless clones** (`git clone --bare --filter=blob:none`): commits + trees only (~10–20% of full size). Blobs are batch-prefetched only for commits in the requested window (promisor remote fetch by OID / `git backfill`), used to compute `--numstat`/patches, then purged with `git repack -a -d --filter=blob:none`.
[...]

PR #2288, opened 2026-08-06
feat(git-cli-proxy): clone-based git extraction service

mitasovr:
Implements the design merged in #2237: a vendor-agnostic **git-cli-proxy** service that serves commit-level data from a bare blobless clone, replacing the one-API-call-per-commit pattern that drives the git connectors into vendor rate limits (worst case ~1000 req/h → a 50k-commit backfill takes days). Refs #2224.

The nocode connectors that consume it were split into #2366.

## What's here

**`src/backend/services/git-cli-proxy`** — gears-rust host, cluster-internal, consumed only by Airbyte connector pods:

- **API v1**: `/v1/commits`, `/v1/file-changes`, `/v1/branches` — cursor-paginated ascending, served from a per-generation index, pinned to a snapshot: a continuation never fetches, a superseded snapshot fails `409` instead of splicing histories.
- **Cache**: bare blobless clones keyed by `sha256(tenant, source, clone_url)`; repo names never build paths. Blobs are transient: prefetched per page window, repacked back to the skeleton after serving — opportunistically first, queued for the entry lock when continuous paging starves the probe. Origins that refuse promisor fetches get promoted to full clones and are never purged.
- **Freshness**: fetch-if-stale with single-flight; requests wait a bounded slice, then `429` + `Retry-After` while the work continues in the background.
- **Auth**: static bearer token on every `/v1` route; a warm read is served only to credentials that proved origin access (fingerprint on the entry). Tenant/source headers partition the cache, never authorize.
- **Credentials**: token lives only in the child process env — invisible in `ps`, absent from disk, rotation needs no re-clone.
- **Disk**: budget with watermarks and two-tier LRU reclaim (blob purge before whole-repo eviction), size accounting that survives continuous paging, per-repo cap → `413`. Origin allowlist only ever widens; in-cluster and file destinations are refused (file:// exists behind an explicit escape hatch for tests).

**Helm** — first PVC-carrying chart in the repo: RWO + `replicas: 1` + `Recreate` are one decision (single cache writer); NetworkPolicy denies ingress until a namespace is allow-listed; registered in the umbrella with `deploy: false`.

## Verification

- 183 unit + 22 integration tests (hermetic `file://` fixtures with partial clone), clippy pedantic clean, coverage above the 80% gate.
- 18 Helm render-contract assertions + `helm lint` in CI.
- Full 85k-commit backfill of a large public repository through the HTTP API: every commit exactly once in walk order, entry disk saw-tooths at its skeleton instead of growing unbounded, no 5xx under a 2 GiB memory cap.

## Not in this PR

- Nocode connectors and their silver models — #2366.
- Vendor-API streams (MRs/notes), rebase resilience beyond the lookback, and the identity fields git does not carry — tracked in the design's open questions.


.github/workflows/git-cli-proxy-helm.yml
```diff
@@ -0,0 +1,39 @@
...
+jobs:
+  helm-contract:
+    name: render contract
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      - uses: azure/setup-helm@fe7b79cd5ee1e45176fcad797de68ecaf3ca4814 # v4
+      - uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5
+        with:
+          python-version: "3.12"
```
cyberantonz, line 37: Why to use outdated 3.12 ?

aleksdotbar:
`/v1/commits` walks `git log --all --since=<value>` (`engine/read/commits.rs:67-72`), and connectors drive `since` from the last `committed_date` (`gitlab-nocode/connector.yaml:311-313`). That makes reachability incremental on committer date. Rewrites are fine — rebase stamps a fresh committer date, so the force-push claim in DESIGN.md §4.2 holds. Commits that become newly reachable while keeping old dates are not:

- `engine/read/commits.rs:67`: 🔴 bug: a restored or first-pushed branch carries commits older than the cursor. `--since` drops them, the `pushed_at` gate advances anyway, nothing revisits them.
- `api/data.rs:90`: 🔴 bug: `branch_membership` runs only on window shas, so `branch_names` never refreshes on stored commits. Commits merged into the default branch keep `["feature/x"]` forever — "commits on default branch" undercounts.

Fix belongs here rather than per-manifest: persist branch heads per snapshot and widen the walk when a head is new or moved non-fast-forward (`git log --all --not <seen-heads>`), or expose a reachability delta.

❓ q: is `--since` a correctness boundary or a walk-cost optimization? If the latter, a head-set delta has the same cost without the gap.

charts/insight/templates/secrets.yaml
```diff
@@ -282,3 +282,28 @@ stringData:
...
+# upgrade (`lookup`), so rotating it is a deliberate act: the connectors carry
+# the same value in their Airbyte source config, and a silent rotation would
+# fail every git sync until they are updated.
+{{- $existingProxy := (lookup "v1" "Secret" .Release.Namespace "insight-git-cli-proxy-config") }}
+{{- $proxyToken := "" }}
+{{- if $existingProxy }}
+{{- $proxyToken = (index $existingProxy.data "APP__gears__git_cli_proxy__config__proxy_token" | b64dec) }}
+{{- else }}
+{{- $proxyToken = (default (randAlphaNum 40) .Values.gitCliProxy.proxyToken) }}
+{{- end }}
```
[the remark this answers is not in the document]
cyberantonz: Use sealed secrets. You generate secret, put it to gitops and then you do not need rotate each render.

docs/components/connectors/git/git-cli-proxy/specs/DESIGN.md
```diff
@@ -243,6 +267,12 @@ nowhere else:
   supplies the username; the proxy does not hardcode vendor rules.
 - Access-log middleware redacts `Authorization`, `X-Git-Token`, and any
   header matched by a deny-list.
+- **`X-Tenant-Id` / `X-Source-Id` are cache partition inputs, not
```
cyberantonz: Tenants can be spoofed, no any checks is it real or not. It is better to create an issue for this tech debt. We need to use proper interservice JWT tokens here with proper gear auth + api gateway.

src/backend/services/git-cli-proxy/helm/values.yaml
```diff
@@ -0,0 +1,73 @@
...
+  size: 50Gi
+  storageClass: ""
+  # Fixed mount path; the config's data_dir points here.
+  mountPath: /app/data/repos
+
+cache:
+  # Cache budget in bytes. Keep 10-15% below persistence.size — headroom for
+  # transient packs and git temp files (an ENOSPC mid-pack-write is a failed
+  # sync, not a graceful reclaim). 50Gi volume -> ~43Gi budget.
+  diskBudgetBytes: 46170898432
```
cyberantonz: Please use variable from values.yaml for disk sizes

src/backend/services/git-cli-proxy/src/api/error.rs
```diff
@@ -0,0 +1,235 @@
...
+        response
+    }
+}
+
+impl ApiError {
+    /// Wire mapping the connector's declarative error handler depends on:
+    /// `429` is retried with `Retry-After`, `409`/`4xx` are not.
+    fn classify(&self) -> (StatusCode, &'static str, Option<u64>) {
+        match self {
+            Self::BadRequest(_) => (StatusCode::BAD_REQUEST, "bad_request", None),
```
cyberantonz: IMO it is quick win to use common errors from gears-rust

src/backend/services/git-cli-proxy/src/api/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+pub struct AppState {
+    pub store: Arc<RepoStore>,
+    pub config: GearConfig,
+}
+
+/// Mount the git-cli-proxy routes onto the host's router.
+///
+/// Our routes live on a fresh sub-router so the bearer middleware and state
+/// scope to `/v1` only; `/healthz` belongs to the `api-gateway` host gear.
+pub fn register_routes(host_router: Router, state: Arc<AppState>) -> Router {
```
cyberantonz: Why not OperationBuilder?

src/backend/services/git-cli-proxy/Dockerfile
```diff
@@ -0,0 +1,94 @@
+# Multi-stage build for Insight Git CLI Proxy (clone-based git data extraction
+# for the nocode git connectors; design: docs/components/connectors/git/git-cli-proxy/).
+#
+# Build context: src/backend/
+# Usage: cd src/backend && docker build -f services/git-cli-proxy/Dockerfile -t insight-git-cli-proxy:dev .
+
+# Stage 1: Builder
+FROM rust:1.95-bookworm AS builder
```
cyberantonz, line 8: 1.97 is current version

cyberantonz (dismissed):

src/backend/services/git-cli-proxy/Dockerfile
aleksdotbar: 1.95 is used in every service

hello1101n (approved):

ktursunov (approved):

ktursunov (approved):

hello1101n (approved):

hello1101n (approved):

PR #2302, opened 2026-08-06
fix(frontend): the person page reports one fact once

dzarlax:
## What

The person page, gone over surface by surface after a design review. Fourteen
commits, each one thing; the diff is frontend-only.

The through-line is not "less red" — it is **one fact, one place**. A screen
that shows a finding twice does not report two facts, it reports one and looks
twice as alarming; a reader counts marks, not findings.

## The bug that started it

`/v1/metric-results` resolves a request all-or-nothing: one key it cannot serve
and the whole call is rejected with 400 UNAVAILABLE, so a single missing metric
blanks a zone instead of its own tile. The frontend's metric keys are compiled
in while the catalog is per installation — seeded from a registry that moves,
and a tenant on an older backend or a different connector set simply has fewer
keys. One key missing is enough for "At a glance" to render "Unable to load"
while its other four requests returned 200.

Requests are now filtered through the installation's own catalog. They wait for
it rather than firing unfiltered "for now" — that would be the same 400
followed by a recovery on a second round-trip — and a collection filtered down
to nothing sends no request at all, since an empty `metrics: []` is itself a
400. If the catalog cannot be read, nothing is dropped: an unknown catalog must
not silently shrink a request.

This is latent wherever the catalog and the frontend agree, and it fires
wherever they do not.

## What the page shows now

**Three blocks, three questions, no overlap.**

| block | answers | lives |
|---|---|---|
| At a glance | what the numbers are | page |
| Needs attention | what stands out | page |
| section marks | which section to open | nav |

The section cards are gone. Every section has its own screen, listed in the nav
to the left; cards restated that list a second time while carrying three
metrics each, dotted individually, so a row of them put a dozen-odd status
marks on a screen whose job was to point somewhere. The mark moved to the nav
item — one dot per section, grey when the section has no data this period. It
costs no request: `usePersonSectionStandings` runs the section screens' own
queries, which react-query serves from cache, and injects the same cohort peers
so the mark and the screen it points at cannot disagree.

**"Needs attention" reports changes, not standings.** It used to be everything
in the bottom quartile of the cohort. For anyone whose job differs from the
people they are measured against — a lead against the developers reporting to
them — that is a permanent state, not an event, and a block repeating it every
period teaches the reader to skip it. An item now has to be below the cohort
AND to have moved against itself by at least a tenth since the previous period.
A structural gap is flat and stays quiet; a regression comes through. Without a
previous period the block makes no claim at all.

**The page names the pool it compares against.** Every "vs median" on screen was
computed against people the reader could not see or count — which is also what
hid the case worth seeing, a lead measured against a whole department. Pools
differ per metric, so the most common size is reported rather than the largest:
the typical comparison, not the most flattering one. The control is called
**Cohort** now, with the word outside the dropdown and a tooltip beside it.

## Honesty rules, applied consistently

- **A dash is not a zero.** A null means no connector feeds that metric for that
  person; a measured zero prints 0 and stays, because a developer who merged
  nothing this period is a finding. Rows and tiles with no value are dropped,
  everywhere, by the same rule.
- **The headline row shows what a person is measured on.** `KPI_ROW` became an
  ordered candidate list; the row renders the first few the person is observed
  for. The tail is a fallback, not a demotion — a developer still leads with
  tasks and PRs because those come first and they have them.
- **One finding, one mark.** The value carries the peer verdict; the gap line
  under it explains rather than judging a second time. Row values in a card lost
  their colour to the dot that already carried the same standing, and the
  coloured left stripe went with them.
- **A badge states its own scope.** "5 of 12 behind peers" — it counts the whole
  section while the rows below show fewer, and a bare count read as a claim
  about those rows and contradicted them.

## Access

The header always offered a button to the person's supervisor and fetched that
supervisor's record for the siblings dropdown. Identity serves a viewer their
own subtree, so anyone above them is not theirs to open — and the analytics read
path does not enforce this on its own (#1995), which makes the link itself the
problem. Both the button and the fetch are gated on the supervisor being inside
the viewer's subtree.

## Legibility

Full metric names that wrap rather than truncate — "Pull requests mer…" is not a
shorter name, it is a missing one. The period range names its year when the
range crosses one or sits outside the current year. Attention rows are a grid:
label, digits right-aligned, unit left-aligned, comparison — so numbers line up
on their last figure and units start together. Every openable surface carries a
standing chevron instead of a hover-only tint. Sections with no data collapse
into the nav's grey mark instead of full-height "No data" cards.

## What this cannot fix

The cohort is `org_unit`: a lead is compared with everyone in their department,
including people doing a different job. Any threshold or presentation built on
top of that answers a question nobody asked, and that is a backend concern —
`metric_entity_cohorts_current` and `cohort_key`. Naming the pool on the page at
least makes it visible to whoever opens the screen.

Two more are registry matters, listed for the record: `collab.focus_time_pct`
counts time outside meetings, which is not focus time, and
`ai.accepted_lines` carries `higher_is_better`, so a larger volume of
AI-generated code reads as a better outcome.

## Verification

`pnpm typecheck` clean, `eslint --max-warnings 0` clean, 848 unit tests pass.
Each change was checked in a running app against a real dataset before being
committed; one direction was tried, seen to concentrate the problem rather than
remove it, and dropped before this branch was opened.

## Summary by CodeRabbit

* **New Features**
  * Added cohort comparison guidance and clearer cohort labeling.
  * Added section standing indicators, peer comparisons, and data-availability messages.
  * Added metric help tooltips and navigation chevrons across dashboard cards and attention items.
  * Expanded KPI coverage while hiding unmeasured metrics and preserving measured zero values.
* **Bug Fixes**
  * Restricted supervisor links to permitted reporting scopes.
  * Excluded unavailable metrics from requests and dashboard displays.
  * Improved custom date-range labels and year handling.
* **Style**
  * Improved value alignment, label wrapping, status presentation, and comparison text.
  * Clarified peer comparisons with expanded “of total” wording.


dzarlax:
All six taken, in `a7602fad`.

**Headline exclusion was too wide.** `KPI_ROW` is a candidate list longer than
the row it fills, so excluding all of it hid a bottom-quartile metric that
never got a slot — visible neither above nor in the attention list.
`metricAttentionItems` now takes the keys the row actually rendered.

**Percent lost its unit.** The list renders the number and the unit in
separate columns, and the percent sign lived in neither, so 50% arrived as
50 — a different number, not a shorter one. `splitMetricValue` derives the
unit from the format.

**The comparison period escaped the gates.** It sits behind the new
"below cohort AND worsening" rule, so a failed comparison rendered as
"nothing needs attention" — the most reassuring possible way to fail. It now
participates in loading, error and retry.

**The legacy dashboard fetched no comparison at all**, which the same rule
would have silently emptied. It now fetches its own.

**The tooltip trigger** is a `<button type="button">`.

**Tests.** `use-person-sections.test.tsx` covers cohort injection, a measured
zero counting as data, no-data, pending, and full section coverage; the nav
mock in `context-pane.test.tsx` is configurable and the three rendered states
are asserted. 858 tests pass, `tsc -b` clean.

aleksdotbar (approved):

dzarlax:
Three more commits, from a fresh look at the person screen after the review
round. All frontend, no API change.

**`e739db99` — the attention block names a standing, not only a change.**
The block only reported metrics that moved against their own past, which hid
its worst case: a metric that fell to almost nothing long ago and has been
flat since. No period shows a change, so the block says nothing, and an empty
block reads as "all clear". There are two ways in now, named apart on the row
— a fall (moved adversely by at least a quarter of the cohort's spread) and a
standing (at a quarter of what a typical peer does, or less). Where there is
no earlier period the row says so rather than staying silent.

Three corrections travel with it: the verdict is the shared standing
derivation rather than a raw quartile test, so the block obeys the same
eligibility guards as the tile beside it; the move threshold is measured in
cohort spreads instead of as a share of the person's own previous value, which
was easiest to clear exactly where the evidence is thinnest; and for metrics
counted in whole events a standing needs the typical peer to be doing at least
a handful, because a share of a tiny number is always large.

One fact also reaches the reader once now: metrics that contain others, and
metrics that restate others in different units, collapse to the more specific
one. The relationships come from the catalog's own definitions — a pair that
merely moves together is not a pair where one contains the other.

**`8c1d23f8` — the value states a quantity, the line states the verdict.**
A tile put two different questions in one red/green channel: the number was
coloured by standing against the cohort, the badge beside it by which way the
number moved against its own past. A red value touching a green badge reads as
a contradiction rather than as two facts. The value is plain ink now and the
standing moved to the line that states the comparison, marked by a dot as well
as a colour — the badge already had an arrow to fall back on, the value had
nothing.

**`f144bda6` — a row of tiles fills the row it is in.**
auto-fit packed in as many tiles as the width allowed and left the remainder
stranded beside a hole, which reads as something that failed to load. The
column count follows the tile count now. The header also names what each
comparison compares, once, instead of leaving a lone percentage ambiguous.

Verified against a local install at three widths; unit and story suites green,
`tsc -b` clean, and each commit builds on its own.

dzarlax:
All four taken, plus the nitpick, in `56e83cdf`.

**Gate attention on comparison-query state** (`screens/dashboard.tsx`) — valid.
The screen fetched the comparison period but `isLoading` ignored it, and a
failure had no surface at all. It now waits on it and renders a retryable error
in the block's place; an empty attention block is the most reassuring possible
way to fail, so it must not be what a failed comparison looks like.

**Include cohort query state before rendering cohort comparisons**
(`metric-groups-view.tsx`) — valid, and the same mistake one screen over.
`cohortKpi` / `cohortGroup` are what every comparison is drawn against, and
`injectCohortPeer` only copies peer values that have arrived, so the page could
render a person measured against a cohort that was not there yet — under a
heading naming its size. Both now participate in loading, error and retry.

**Render explanation-only metric help** (`members-grid.tsx`) — valid. Fixed as
suggested: the explanation follows the description, and a column whose catalog
entry carries only the explanation is no longer a silent header.

**Make `MetricName` help available to keyboard users** — taken, with a
different remedy than the first option offered. Making that trigger focusable
is not available to us: these names sit inside cards that are themselves
buttons, and nesting a control in a control is invalid markup with broken
semantics. So the second option — expose the help as accessible text — is what
landed: the catalog's words are rendered for assistive technology in the
reading order they belong to, with a test for both the populated and the empty
case. A sighted keyboard user still reaches the same words by opening the card.

**Assert the complete configured section set** — done; the test now compares
against `GROUPS` rather than sampling two ids. A section dropped from the list
would otherwise lose its mark in the nav unnoticed.

942 tests pass, `tsc -b` clean.

aleksdotbar (approved):

Issue #2301, opened 2026-08-06
ci(security): the secrets gate cannot tell a credential from ordinary code, including for services we integrate with


Gregory91G:
## Summary

The secrets gate cannot tell a credential from an ordinary fragment of code, and #2300 did not change that — it only stopped the resulting failure landing on an unrelated author. A full scan of every ref with the pinned image (`trufflehog 3.96.0`, all result kinds, 47.6 MB / 37 783 chunks) returns **272 findings from 9 detectors, and not one of them is a credential**. The 69 findings marked `verified` are all Lob: its sandbox accepts any `test_`-prefixed string, so a Python test function name comes back "verified". The verified/unverified split therefore carries no signal today.

`--exclude-detectors lob`, merged in #2300, is a stopgap and is explicitly not the answer: it works only because Lob is a service we do not use. It does nothing for the detectors that cover services Insight does integrate with, and those are the ones that matter.

## Measured inventory (2026-08-06, all refs, trufflehog 3.96.0)

| Detector | Findings | verified | What the match actually is |
|---|---|---|---|
| JiraToken | 164 | 0 | 3 distinct fixture values repeated across `src/ingestion/tests/e2e/metrics/*.test.yaml` |
| Lob | 69 | 69 | 54 distinct test function names — `test_` + 35 chars is also the shape of a Lob key |
| Gitlab | 19 | 0 | identifiers and constant names, e.g. `CURSOR_OVERLAP_MINUTES` |
| HubSpotApiKey | 6 | 0 | UUIDs, including synthetic `00000000-…-0010` |
| URI | 6 | 0 | local dev connection strings, `http://…@localhost:9000` |
| SQLServer | 4 | 0 | `CHANGE_ME` placeholders in `.env.local.example`, a MariaDB test factory |
| Wit | 2 | 0 | a zero-padded hex id in a Rust migration |
| Atlassian | 1 | 0 | an example account id quoted from Atlassian's own documentation |
| CloudflareApiToken | 1 | 0 | a prose fragment in `STORAGE_TECHNOLOGY_EVALUATION.md` |

Every row is a code identifier, a fixture, a placeholder or prose. The pattern is the same in each case: the detector matches a *shape*, and ordinary source code produces that shape.

## Why excluding detectors does not generalise

[...]

PR #2308, opened 2026-08-07
ci: name the GHCR packages explicitly instead of globbing

### LINKED ISSUE #2307: ci: GHCR cleanup fails on every branch deletion — a glob in image-names needs a PAT
## Summary

`GHCR cleanup` fails on every run. The last 12 runs are failures, all on `event=delete`, i.e. one red run per deleted branch — the retention policy has not run successfully at all.

## The failure

```
thread 'main' panicked at src/client/client.rs:46:17:
Restrictions in the Github API prevent us from listing packages when using a
$GITHUB_TOKEN token. Because of this, filtering with '!' and '*' are not
supported for this token type.
```

Both jobs pass `image-names: "insight-*"` with `token: ${{ secrets.GITHUB_TOKEN }}`. `snok/container-retention-policy` resolves a glob by calling the list-packages endpoint, which a `GITHUB_TOKEN` cannot reach; with an explicit name it calls get-package instead, which works. The action documents this: the `*` and `!` operators "are only available for personal- and GitHub app-tokens".

## Impact

Bounded but not zero. Both jobs are still `dry-run: true`, so nothing was being deleted and no image was lost. What we do have is a permanently red workflow that fires on every branch deletion, and a retention policy that has never actually been exercised — so the day someone flips `dry-run: false`, the first real run will be the first run.

## Fix

Replace the glob with the explicit, space-separated list of the images `build-images.yml` publishes. No new secret, no new permission.

Refs #1994.

Gregory91G:
Closes #2307.

`GHCR cleanup` has failed on every run — the last twelve, all on `event=delete`, so one red run per deleted branch. The action panics before it lists a single version:

```
Restrictions in the Github API prevent us from listing packages when using a
$GITHUB_TOKEN token. Because of this, filtering with '!' and '*' are not
supported for this token type.
```

Both jobs asked for `image-names: "insight-*"` while authenticating with `secrets.GITHUB_TOKEN`. A glob is resolved through the list-packages endpoint, which that token cannot reach; an explicit name goes through get-package, which it can. The action documents the split — `*` and `!` "are only available for personal- and GitHub app-tokens". The alternative fix, a PAT with `read:packages`, would mean a long-lived org-wide token in secrets; naming the packages needs neither a new secret nor a new permission.

**Coverage grows rather than shrinks.** The list is every image `build-images.yml` publishes — 15, including the eight connector images the old glob never reached, because they do not start with `insight-`. The dry-run below finds real candidates in them: 10 in `source-gitlab-insight`, 8 in `insight-jira-enrich`, and more.

Packages the glob used to touch and this list omits, with where they belong:

| Package | Why it is not here |
|---|---|
| `insight-front` | published by `constructorfabric/insight-front`; retention for it belongs in that repository |
| `insight-lite` | published by `constructorfabric/insight-lite`, same |
| `insight-identity`, `insight-analytics-api`, `insight-api-gateway` | legacy names, superseded by `insight-identity-resolution`, `insight-analytics` and `insight-gateway`; nothing in this repository publishes them any more, so no new tags accrue. They need a one-off decision — delete or keep — not a retention rule. |
| `charts/insight` | the chart package, deliberately never in scope |

Nothing else moves. `image-tags`, `cut-off`, `keep-n-most-recent` and `dry-run: true` are untouched — tag globs keep working, because they filter versions of an already-resolved package rather than resolving the package itself.

The cost is a maintenance step: a new image in `build-images.yml` needs a line here too. That is one line in the header comment; the reasoning lives in this pull request and #2307 rather than in the file.

## Why the list lives in `env`, and the one `trufflehog:ignore`

The names are declared once at workflow level and both jobs read them, so the list cannot drift between the trunk and branch policies.

The connector line carries a trailing `# trufflehog:ignore`. Without it the secrets gate fails this pull request: the GitLab detector takes the `gitlab` in one connector name as its keyword and then matches a neighbouring name as a token, reporting `source-hubspot-insight` as a credential. The marker suppresses findings **on that line only** — the detector stays enabled everywhere else, and a real GitLab token in this file or any other still blocks. Disabling the detector was not an option: GitLab is a service Insight integrates with. This is the false-positive class tracked in #2301.

The file carries no comment about any of this, deliberately — the reasoning belongs here and in the issues. Two consequences a reader of the diff alone cannot see, and which therefore live here:

- `SERVICE_IMAGES` holds the seven service images, `insight-analytics` first; `CONNECTOR_IMAGES` holds the eight connector ones. Both are passed to `image-names`, so all fifteen are in scope. The split exists only so the marker covers the connector line alone.
- Adding an image to `build-images.yml` means adding it to one of these two variables. Nothing enforces that yet; generating the list from the same source the build uses would, and is worth a follow-up.

## Test plan

- [x] `actionlint .github/workflows/ghcr-cleanup.yml` — exit 0
- [x] The folded scalar parses to one space-separated string of 15 names in both jobs
- [x] Dispatched from this branch ([run 31139161264](https://github.com/constructorfabric/insight/actions/runs/31139161264)): `trunk` **succeeded**, zero panics, zero errors, and all 15 packages were enumerated — 509 candidates in `insight-toolbox`, 24 in `insight-analytics`, 10 in `source-gitlab-insight`, 8 in `insight-jira-enrich`, down to 2 each in the smaller connectors. Everything reported as `dry-run: Would have deleted …`; nothing was deleted.
- [x] An earlier dispatch of the same list ([run 31139064023](https://github.com/constructorfabric/insight/actions/runs/31139064023)) died on `{"message": "We couldn't respond to your request in time…"}` — a GitHub API timeout on one package's version listing, which the action does not retry. It passed on the next attempt with no change. Worth knowing: a longer list means more paginated calls, so this run can fail transiently and needs a re-run rather than a fix.
- [ ] `branch` job green on the next branch deletion after merge — it takes the same input, but only a `delete` event exercises it

## Summary by CodeRabbit

* **Chores**
  * Updated automated container image cleanup to cover explicitly defined service and connector images for both trunk and branch builds.
  * Preserved existing retention periods, tag matching, and cleanup behavior.


ktursunov (approved):

.github/workflows/ghcr-cleanup.yml
```diff
@@ -43,7 +49,7 @@ jobs:
         with:
           account: constructorfabric
           token: ${{ secrets.GITHUB_TOKEN }}
-          image-names: "insight-*"
+          image-names: ${{ env.SERVICE_IMAGES }} ${{ env.CONNECTOR_IMAGES }}
```
cyberantonz, line 48: Why only connector images? What about insight-analytics, for example?

cyberantonz (changes requested):

.github/workflows/ghcr-cleanup.yml
Gregory91G: the list is bigger - as you may see in string 38-39, it includes `insight-analytics` and other non-connector images

cyberantonz: i see, thanks

cyberantonz (approved):

PR #2322, opened 2026-08-07
feat(gitops): provision grafana_ro with Grafana, not with ingestion

mitasovr:
Grafana's ClickHouse datasource needs a ClickHouse login to chart connector load volume out of `bronze_*` — the connector dashboards can say whether a sync ran and what it cost, but nothing except the warehouse can say whether it moved any data. That login must not be the admin user, which holds `CREATE USER` / `DROP TABLE` / `INSERT`.

> **Rewritten.** The first version of this PR created the user from the umbrella's clickhouse-migrate Hook, next to `presentation` (#1963/#1964) — less code, wrong dependency direction. Ingestion ships to every instance; Grafana is optional (`inventory.system.grafana`). A cluster without Grafana should carry no `grafana_ro`, and nothing under `src/ingestion` should exist because of a dashboard. This version touches neither `src/ingestion` nor `charts/`.

## Shape

Grafana owns its own user. `make system-grafana` gains one step:

```
make system-grafana
  ├── apply environments/<env>/sealed-secrets/insight-infra/grafana*-sealedsecret.yaml
  ├── system/grafana/provision-clickhouse-user.sh   ← role + user in ClickHouse
  └── helm upgrade --install grafana
```

| File | Role |
|---|---|
| `system/grafana/clickhouse-role.sql` | role + grants |
| `system/grafana/provision-clickhouse-user.sh` | the grant-less user |
| `system/grafana/SECRETS.md` | operator contract |
| `system/grafana/tests/test_clickhouse_role.py` | grant-matrix pin |

The ordering is deliberate: the Secret lands first (the script reads the password from it), the user exists before helm, so the datasource authenticates on first start rather than after a restart.

## Everything degrades, nothing fails

- no `grafana-clickhouse` Secret → role, no user
- external / managed ClickHouse → skips with a note
- admin without `access_management` → skips with a warning

In each case Grafana installs and only its ClickHouse datasource is unhealthy; the Loki dashboards are untouched. Re-running converges. Both `envValueFrom` keys are `optional`, so a missing Secret is a degraded datasource rather than a pod that will not start.

## Why the role grants `SELECT ON *.*`

Unlike `presentation_ro`, which enumerates its contract databases, this one takes a wildcard: every connector onboarding creates a new `bronze_<name>`, and a fixed list would silently drop that connector out of the dashboards until someone noticed. Read-only-by-construction still holds — `SELECT` and `SHOW` are the only privileges, and the wildcard only discloses data the dashboards exist to display. On top of the role the user carries `readonly = 2` and `max_execution_time = 60`, matching the datasource's `queryTimeout`.

## Verification

The script has two modes so it stays testable: cluster mode resolves credentials from Secrets and reaches ClickHouse over a temporary port-forward; setting `CLICKHOUSE_URL` bypasses kubectl entirely. `tests/test_clickhouse_role.py` drives the second — 10 tests, green against ClickHouse 25.7.5:

- every database is readable
- **a database created *after* the grant is readable** — the point of the wildcard
- every write and DDL refused, in every database
- self-escalation refused, including `SET readonly = 0`
- a re-run converges an existing user
- the role-only path when no password is supplied

Cluster mode was run against a live stand, and it earned its keep: the port-forward leaked. Backgrounding a shell function makes `$!` the subshell's pid, so the real `kubectl` survived as its child and the tunnel outlived the script. It now invokes `kubectl` directly. That path is not covered by the tests — only the live run found it.

## Operator side

insight-gitops!67 carries the same four files for the internal deployment, plus the dashboards and panels that use the datasource.

Not ported here: the five extra dashboards and the loaded-volume panels live only in the internal copy for now. This PR adds the mechanism (plugin, datasource, user), not the content.

### DIFF
--- a/deploy/gitops/Makefile
+++ b/deploy/gitops/Makefile
@@ -612,10 +612,20 @@ system-alloy: vpn-up kube-ctx
 		$(call _system_values_args,alloy) \
 		--wait --timeout 5m
 
+# Grafana's ClickHouse datasource needs a ClickHouse login, and this is the
+# only place that knows Grafana is being installed at all — the umbrella must
+# not provision it, because ingestion ships to every instance and Grafana does
+# not. Runs after the sealed secrets land (it reads the password from one) and
+# before helm, so the datasource authenticates on first start. Idempotent, and
+# it degrades to a warning rather than failing the install when ClickHouse is
+# external or the password has not been sealed yet.
 .PHONY: system-grafana
 system-grafana: vpn-up kube-ctx
 	@$(call _helm_repo_add,grafana,https://grafana.github.io/helm-charts)
 	$(call _apply_optional_system_secrets,grafana)
+	@echo "$(C_GRN)→$(C_RST) provisioning grafana_ro in ClickHouse"
+	@NAMESPACE=$(NS_INFRA) KUBE_CTX=$(KUBE_CTX) \
+		bash system/grafana/provision-clickhouse-user.sh
 	helm upgrade --install $(GRAFANA_RELEASE) $(GRAFANA_CHART) \
 		--namespace $(NS_INFRA) --create-namespace \
 		--version $(GRAFANA_VERSION) $(HELM_EXTRA_FLAGS) \
--- a/deploy/gitops/system/grafana/SECRETS.md
+++ b/deploy/gitops/system/grafana/SECRETS.md
@@ -1,0 +1,114 @@
+# Grafana — Secrets
+
+Grafana itself needs no credentials to install. This file covers the one
+optional Secret it does read: the ClickHouse login behind the `ClickHouse`
+datasource in `values.yaml`.
+
+## `grafana-clickhouse` — the ClickHouse datasource login
+
+Grafana charts warehouse volume straight out of `bronze_*`, which needs a
+ClickHouse user. It must not be the admin (`clickhouse.auth.username`, which
+holds `CREATE USER` / `DROP TABLE` / `INSERT`) — a dashboard has no business
+carrying any of that. So `make system-grafana` provisions a dedicated
+`grafana_ro`:
+
+```
+make system-grafana
+  ├── apply environments/<env>/sealed-secrets/insight-infra/grafana*-sealedsecret.yaml
+  ├── system/grafana/provision-clickhouse-user.sh   ← role + user in ClickHouse
+  └── helm upgrade --install grafana
+```
+
+**Why here and not in the umbrella chart.** The umbrella already provisions
+ClickHouse users at deploy time (the `presentation` user for analytics), and
+putting `grafana_ro` there would have been less code. It would also have been
+backwards: ingestion ships to every instance, Grafana is optional
+(`inventory.system.grafana`). A cluster without Grafana should carry no
+`grafana_ro`, and the thing that installs Grafana is the thing that should
+create its user.
+
+### Required keys
+
+| Key | Used for |
+|---|---|
+| `username` | the ClickHouse user Grafana connects as — always `grafana_ro` |
+| `password` | its password; the provisioning script sets the same value in ClickHouse |
+
+### Everything is optional, and degrades rather than fails
+
+- no `grafana-clickhouse` Secret → the script provisions the role and no
+  user, Grafana installs, and only the ClickHouse datasource fails its health
+  check. The Loki dashboards are unaffected.
+- ClickHouse not in this namespace (external / managed) → the script says so
+  and skips; provision `grafana_ro` out-of-band with `clickhouse-role.sql`
+  plus a `CREATE USER`, then seal the Secret.
+- admin without `access_management` → same, warns and skips.
+
+So the order of operations does not matter: install Grafana first and seal the
+password later, or the reverse. Re-running `make system-grafana` converges.
+
+### Adding it to an environment
+
+```bash
+# 1. Generate. Alphanumeric on purpose: the script refuses a password with a
+#    quote, backslash or semicolon, because the value rides the SQL body of
+#    the CREATE USER statement.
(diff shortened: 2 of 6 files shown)


ktursunov (approved):

cyberantonz (approved):

aleksdotbar (commented):
Design is right (mirror #1963/#1964, static role DDL + guarded password script, optional key). Three blockers below, rest is polish.

---

`tests/test_grafana_role.py:111`: 🔴 `DROP DATABASE IF EXISTS` over `SEEDED_DBS` drops **`insight` and `silver`** on whatever server `GRAFANA_ROLE_TEST_CH_URL` points at — one mis-set env var destroys a warehouse. `test_presentation_role.py` deliberately tears down tables only. Seed `{db}.probe` and drop only those tables; for the wildcard proof use collision-proof names (`bronze_grafana_role_probe` / `_late`) and drop just those two DBs.

`bootstrap-db/provision-grafana-access.sh:43`: 🔴 `readonly = 2` permits `SET`, so `max_execution_time`/`max_result_rows`/`result_overflow_mode` are defaults, not ceilings — the user can `SET max_execution_time = 0`. Only `readonly` itself is unliftable, so the "runaway query dies in ClickHouse" guarantee in the comment and PR body does not hold. Constrain: `max_execution_time = 60 MAX 60, max_result_rows = 2000000 MAX 2000000, result_overflow_mode = 'throw' CONST`. Also no memory bound — add `max_memory_usage ... MAX` for a wildcard-SELECT user scanning raw bronze.

`bootstrap-db/grafana-role.sql:15`: 🟡 both arguments for `SELECT ON *.*` need reworking before this lands.

Bronze databases are not out-of-band: they are committed DDL in `connectors-ddl/*.sql`, applied by `create-bronze-placeholders.sh` in this same Job on every upgrade. A grant loop derived from `system.databases WHERE name LIKE 'bronze%'`, run after the placeholder step, refreshes itself exactly when a new connector's DDL lands — same convergence, no wildcard. Would need the provisioning call moved after `create-bronze-placeholders.sh` in `apply-ch-migrations.sh:49`.

And the stated blast radius is wrong: a datasource is not a panel. Any Grafana user with Explore gets `SELECT` on `identity.*` and `person.*` (PII, per `presentation_ro`'s own contract), `presentation.*`, and `system.query_log` (every query text every other user ran). The dashboard reads `_airbyte_extracted_at`; the credential reads the whole warehouse.

Keeping the wildcard is defensible — but then state that blast radius and pair it with the Grafana-side control (datasource restricted to viewers, Explore off). The current comment reads as if exposure is bounded to the panels.

---

`deploy/gitops/environments/local/sealed-secrets/insight/insight-db-creds-sealedsecret.yaml.template`: 🟡 shape doc still lists only `clickhouse-presentation-password`; #1964 added its key here for this reason. Add `clickhouse-grafana-password`. (Checked other consumers — `functional-k3s.yml:171` correctly needs nothing, no Grafana in CI.)

`charts/insight/templates/clickhouse-migrate-job.yaml:115`: 🟡 "the gitops repo's clickhouse SECRETS.md" — `deploy/gitops/system/clickhouse/SECRETS.md` is in *this* repo and documents `clickhouse-creds`/`admin-password`, not this key. Fix the pointer or add the section.

`bootstrap-db/provision-grafana-access.sh:46-49`: 🟡 user and role both named `grafana_ro` → `GRANT grafana_ro TO grafana_ro`, ambiguous `SHOW GRANTS FOR grafana_ro`, and `DROP USER` vs `DROP ROLE` as a live incident hazard. Precedent splits them (`presentation` / `presentation_ro`). Name the user `grafana`.

`tests/test_grafana_role.py:149`: 🟡 `test_rerun_converges_password_and_settings` asserts `SELECT 1`, nothing about settings. Assert as the grafana user: `SELECT name, value FROM system.settings WHERE name IN ('readonly','max_execution_time','max_result_rows')`. That assertion fails today under the `readonly = 2` finding — which is the point.

`tests/test_grafana_role.py:141`: 🟡 escalation test checks only `not ok`; a typo passes. Assert `ACCESS_DENIED` for the DDL/GRANT cases, `READONLY` for `SET readonly = 0` (precedent asserts codes).

`tests/test_grafana_role.py:164`: 🟡 `test_role_only_when_password_unset` takes no fixture but drops the user the module fixture owns, then the role — passes only in definition order, breaks under `-k`, `--lf`, xdist. Fixture also never drops the role (leak). Give it its own fixture.

`bootstrap-db/grafana-role.sql:16`: 🔵 `GRANT SHOW ON *.*` is dead — CH grants `SHOW` implicitly on any object holding another privilege. Drop it.

`bootstrap-db/provision-grafana-access.sh:43`: 🔵 `max_result_rows = 2000000` and `result_overflow_mode = 'throw'` documented nowhere — comment and PR body describe two settings, the list has four.

`charts/insight/values.yaml:285`: 🔵 comment sits between `initDatabases` and `runMigrations`, so it annotates `runMigrations`. It is about password keys — move under `passwordSecret`.

`bootstrap-db/provision-grafana-access.sh:27-33`: 🔵 guard rejects `'`, `\`, `;` but not newline. `run_ch` strips full-line `--` comments, so a password with a line starting `--` is silently mangled → CH password diverges from the sealed secret → Grafana auth fails with no signal. Pre-existing in `presentation` too; one `*$'\n'*` arm fixes both.

PR body: 🔵 names `dev-vhc` and the gitops MR. Deployed-instance and deploy-ops specifics stay out of this repo.

---

Checked and clean: password interpolation is injection-safe (bash does not re-expand a variable's value; the guard closes the quote-break path); `optional: true` secretKeyRef matches presentation exactly; `CREATE USER IF NOT EXISTS` + unconditional `ALTER USER` is correct rotation convergence; provisioning order vs `create-bronze-placeholders.sh` is harmless *while* the grant stays a wildcard.

aleksdotbar (changes requested):

hello1101n:
will create another one based on it

Issue #2312, opened 2026-08-07
Attribute policy and account assignment are not readable from the analytical store


aleksdotbar:
## Problem

The query path cannot see either half of what it needs to answer a grouped or
comparative request.

Attribute policy is administrative state: it is edited transactionally, it is
audited, and it lives where those properties hold. The query path is
elsewhere. Reading policy at request time would put an administrative service
on the critical path of every analytical query, so a request either consults
something it should not depend on, or proceeds without knowing whether an
attribute may be grouped, may drive a comparison, or is sensitive.

Account-to-person assignment has the same shape from the other side. Facts
discovered in source systems are recorded against the accounts that produced
them, while measurements are attributed to people. Nothing in the analytical
store states, for a given account, which person it currently resolves to, so
a request cannot turn account-scoped evidence into person-scoped results
without leaving the store.

Both are read on every request that groups or compares, so both must be
available where those requests are evaluated, and both must be pinnable: a
result has to be explainable in terms of the exact policy and the exact
assignment state that produced it, not whatever the state happened to be by
the time someone asked.

## Scope

Making the current attribute policy and the current account-to-person
assignment readable from the analytical store, with a stated freshness and
something a request can pin them by.

The registry those policies are edited in, the attribute catalog, and the
grouping and comparison APIs are tracked separately under the parent epic.

Issue #2028, opened 2026-07-29
GRAPH 7: [EPIC] Cohorting is a GROUP BY over identity attributes


dzarlax:
## Goal

> **A cohort is a `GROUP BY` over identity attributes — for the period being measured, with a rule for people who have several values, and a minimum group size.**

That is the whole idea. Group people by their attributes, hide a group that is too small, and say what the group was made of. Attributes come from identity: some are ingested from source systems, some we derive ourselves. Nothing else in the system keeps its own idea of what a "group" is.

Each of the three conditions is in that sentence because of something the system gets wrong today.

- **For the period being measured.** A person's group is taken as it is *right now*. Someone who moved from one division to another in June has their May numbers compared against the division they were not in yet, and nothing on the screen says so.
- **A rule for people with several values.** If a person belongs to two groups — say they hold two functions — a plain `GROUP BY` puts them in both. The group they are compared against then contains that person twice, at double weight.
- **A minimum group size.** `GROUP BY` will return a group of one person and hand back that person's own number as the group median. "Below your peers" can then mean "below yourself".

The code behind each of these is in **Why now**, further down.

---

## How a user uses this

On a person's or a team's screen there is a **"compare with"** control. It offers the attributes identity holds about people — division, department, job title, manager, office — and later the derived ones, such as function.

1. The user picks **one or several** of them. Picking *job title* and *office* means "people with the same title, in the same office".
2. Every number on the screen that supports comparison now shows three things: the person's own value, the group's median, and **how many people are in the group**.
3. The group is always described in words next to the numbers: *"compared with: same title, same office — 14 people"*. A comparison is never shown without saying who it is against.
4. If someone has defined a **named group** ("Frontend in Cloud"), the user can pick that instead. The difference matters: "people like this person" changes with whoever you are looking at, a named group does not, so two users looking at the same named group see the same numbers.
5. The selection applies **for the period on screen**. If a person changed division inside that period, the grouping used is the one valid then, and the screen says so.

### What the user sees when a comparison is not possible

Refusals are part of the feature, not error states. Each one says what happened and what to change.

| Situation | What the user sees |
|---|---|
| Fewer people match than the minimum | *"Only 3 people match this selection — too few to compare."* Plus the option to drop one condition |
| The attribute may not be used for comparison (gender, age band, pay band) | Grouping still works; the comparison says *"this attribute can't be used as a comparison group"* |
[...]

Issue #1983, opened 2026-07-28
metric-results: peer cohorts only support org_unit, so slice-based peer comparison is computed client-side


dzarlax:
## Summary

`/v1/metric-results` computes peer stats server-side, but always against **one** cohort: `metric_definitions.peer_cohort_key`, which is `org_unit` for all 59 definitions, and `insight.metric_entity_cohorts_current` only carries `cohort_key = 'org_unit'`. Any UI that lets the reader re-group people by another attribute (division, title, manager, …) therefore cannot use the server's peer view for that grouping.

The portal (constructorfabric/insight-front#221) has a global "slice" control that re-cohorts every surface at once, so it currently recomputes quartiles **client-side** (`src/lib/insight/within-team-peer.ts`) over the fetched member set. That works, but it means:

- two implementations of the same statistic (server `quantilesExactIf` vs client `quantile`) that can drift;
- the client can only rank people it has already fetched, so a cohort is silently truncated to the current page/scope;
- the small-cohort suppression threshold is duplicated (server `min_peer_n`, client `MIN_COHORT = 4`).

## What exists today (verified)

- `compile_peer_batch_query` (`src/backend/services/analytics/src/domain/metric_results/compiler.rs`) already takes a `cohort_key` parameter and returns p25/median/p75/min/max/n with a `min_peer_n` pool gate.
- The request DTO already accepts it: `MetricViewRequest::Peer { cohort_key: Option<String> }` (`domain/metric_results/dto.rs`).
- `SELECT DISTINCT cohort_key FROM insight.metric_entity_cohorts_current` → `org_unit` only.
- The FE type carries `cohort_key?: string` but the collection layer never sends one (`src/lib/metrics/collection.ts` returns `{ view: "peer" }`), so the definition default always wins.

So the plumbing is in place end to end; what is missing is **cohort rows for more than one attribute**.

## Ask

1. Emit additional cohort keys into `metric_entity_cohorts_current` from the identity attributes that already exist on the person profile: at minimum `division`, `department`, `job_title`, `manager` (keyed on the supervisor's stable id/email, not the display name — two managers can share a name).
2. Document the available cohort keys somewhere the client can read them (a small `GET` or a field on the metric-definition response), so the UI can offer exactly the groupings the data supports instead of guessing.
3. Keep the existing `min_peer_n` suppression semantics — the UI relies on "no peer data" being an honest state rather than a comparison against one or two people.

## Why this unblocks UI work

[...]

PR #2335, opened 2026-08-07
feat(frontend): the cohort choice follows the catalog, not the roster

dzarlax:
## Problem

The portal decides which comparisons it can offer by walking the viewer's
roster and keeping the attributes that take more than one value
(`collectRosterAttrs` + `availableSlices`). That only works for a viewer who
can see other people, and identity serves a viewer their own subtree
(`use-org-scope.ts`, "permission boundary"). A viewer with no direct reports
therefore has a roster of one person, no attribute reaches the two-distinct
-values threshold, and no re-grouping is possible at all. That is most readers,
not an edge case.

The comparison itself was never the problem — the peer view computes it
server-side within the person's organization unit. What is missing is the
choice. Meanwhile the control still rendered a single option, which reads as a
setting the reader chose rather than the only thing available.

## Change

`cohortOptions` becomes the one place that decides, from two sources in order:

1. **the server's catalog** of attributes a comparison may be built on — the
   shape epic #2028 introduces, decided from governed policy (#2284, #2312)
   rather than from whoever happens to be visible;
2. **the roster walk**, until that lands, so nothing regresses for a viewer who
   has reports today.

They are not merged. A locally derived attribute standing beside a governed one
is indistinguishable to the reader, who cannot tell which of the two they
picked. When neither source yields anything the control states the fallback
instead of offering one option.

## Working now, and working later

The catalog is read in a single function against an **optional** field, so an
installation on either side of the change behaves correctly: absent today,
followed once present, with no second code path to keep in step.

One thing needs confirming: `docs/domain/person-attributes/specs/DESIGN.md`
§3.3 says `GET /v1/metric-definitions` gains "the active allowed-attribute
catalog", but does not pin the field name or entry shape. This reads
`comparison_attributes: [{ id, label }]` as a placeholder, in one function, so
agreeing on the real name is a one-line change here rather than a search
through the portal. Happy to align it with whatever #2285 / #2313 settle on.

## Notes

- constructorfabric/insight#1983 asked for the opposite shape — more
  `cohort_key` rows and a document listing them. Closed as superseded: the
  design answers both asks differently, and extending `cohort_key` would build
  on the path being retired.
- No API call changes in this PR; only which options the control offers and
  what it says when it has none.

## Verification

937 tests pass, `tsc -b` clean, eslint clean. New tests cover the catalog path,
the roster fallback, the empty case, a response without the field, and entries
missing an id or label. Checked against a running stand at three widths.

## Summary by CodeRabbit

- **New Features**
  - Cohort dimensions now use available catalog attributes, with roster data as a fallback.
  - Added clearer handling when no cohort dimensions are available.
  - The cohort selector is disabled and explains why selection is unavailable when no options exist.

- **Bug Fixes**
  - Improved consistency of cohort options across portal views.

- **Tests**
  - Added coverage for catalog, roster fallback, empty states, and disabled selector behavior.


dzarlax:
Taken, in `79254f99` — and it was a real disagreement, not just a stale value.

`SliceSelect` falls back to "Team (all)" when its value is not among the
offered dimensions, but `usePersonCohort` read the stored slice directly and
never consulted them. So with a link carrying a dimension the catalog does not
offer, the control said "Team (all)" while the comparison was still being built
from the old attribute. The screen said one thing and did another — worse than
a stale selection, because nothing on it was wrong-looking.

Fixed one level up rather than by clearing state in an effect.
`useCohortOptions` now owns both the options and which selection is in effect,
and the control and the cohort builder read the same answer. Deriving it once
removes the possibility instead of correcting it afterwards, and avoids an
effect that writes state back while the sources are still resolving — which is
also why a stored value is kept until they have: a shared link carrying
`?slice=` must survive the moment before the options arrive.

The catalog-transition test you asked for is in the cohort hook's own suite,
and it runs the real query against a real query client with only the request
stubbed — so a cohort built from an attribute the catalog does not offer fails
there rather than passing behind a stubbed hook. Verified by mutation:
restoring the direct read of the stored value fails exactly that case and
leaves the other sixteen passing.

941 tests pass, `tsc -b` and eslint clean.

hello1101n (approved):

PR #2340, opened 2026-08-07
feat(frontend): the person page is the person's own

dzarlax:
## Whose page this is

The person page is the person's own — for most readers it is the first screen
they meet, and they did not ask to be measured. That answers a question the
page had been getting wrong: what leads, what recedes, and what it owes the
reader before any number.

## What changed

**It names its own limits first.** The page opened with four numbers and a red
list, and read as a complete picture of a person. For someone whose work leaves
few traces in the connected systems, that picture is mostly their chat and
their calendar. It now says which sections have no reading at all, above the
numbers — because the first question on your own page is what this thing knows
about you, and an answer at the bottom arrives after the page has already been
read as complete. "No data reaches us", not "no activity": only one of those is
something we know. No extra request — the standings are the ones the navigation
already asked for.

**The tile leads with the comparison the reader owns.** It carried two and
named neither: a badge against the person's own last period, a line against the
cohort median, so a bare "-13%" was readable as either. The reader did not
choose the cohort, cannot see who is in it, and cannot decide its median is the
right target — that is context. Their own last period is the one comparison
they can act on. Value in plain ink; under it, named and coloured, the change;
under that, quietly, the cohort. The badge is gone, and with it the header
clause that existed to explain it.

**Each headline number carries where it has been.** A single period plus two
comparisons cannot say whether something is new, recovering, or how it has
always been — three findings that call for three different conversations. A
line now sits beside each value, in the unit the reader picked.

**Whose page it is decides what comes first.** A standing is the larger claim
and leads by default, which is what a manager scanning someone else needs; on
your own page what moved this period leads, because that is what you can still
act on. Neither order hides anything.

## Three deliberate refusals in the line

- The **running period is not drawn**. A month four days old holds four days of
  work; beside finished months it invents a collapse, and nothing on a tile
  could tell the reader that the last point means something else.
- **Fewer than three readings draw nothing.** Two points always make a straight
  line, and a straight line reads as a steady trend — but two is also exactly
  what a metric has when it has only just started being measured.
- A **missing reading breaks the line** rather than dropping it to the floor: a
  period with no data must not be readable as a period with none of the thing.

The line is scaled to its own range and says only which way and how steadily —
the figures beside it carry the magnitude. It has a rule beneath it so it reads
as a span rather than floating, and the span is named once above the row, since
a tile has no room for an axis. Its request sits outside every loading and
error gate: the tiles render from the current period and the line appears when
it arrives, so the first screen never waits on it.

## One thing removed

`showSections` gated the per-group cards **and** the drilldown sheets. The cards
moved to the navigation long ago, so what survived was a flag whose only
remaining effect was to make a click land nowhere on any screen that set it.
Removed; the sheets now follow the only question that matters — whether the
caller routes clicks elsewhere.

## Not in this PR

Which metrics the headline row shows is unchanged: a fixed priority order,
filtered to what the person is observed for. For a role the order was not
written around, that shows what happens to be instrumented rather than what the
work is — which needs role expectations, not a layout change.

## Verification

943 tests pass across both projects, `tsc -b` and eslint clean. Checked against
a running stand at several widths and periods.

## Summary by CodeRabbit

- **New Features**
  - KPI tiles now show trend charts, comparison periods, directional changes, and peer benchmarks.
  - Added messaging identifying sections without measured activity.
  - Added trend-range legends and clearer handling when comparison data is unavailable.
- **Improvements**
  - Attention items prioritize recently changed metrics on a person’s own page while preserving existing ordering elsewhere.
  - Drilldown panels now open automatically unless navigation is handled directly.
- **Bug Fixes**
  - Improved neutral, missing-data, and edge-case KPI display behavior.


dzarlax:
Both findings were right — neither path had a test. Addressed in `ebb1b22` and `374f257`.

**Trend path.** Added `person-trend.test.ts` (11 cases) and `sparkline.test.tsx` (5), plus two cases on the tile. What they pin is deliberately the *refusals*, because each one looks like a missing feature from the outside and would otherwise be "fixed" by deleting it: the running period is not drawn, fewer than three readings draw nothing, a gap breaks the line rather than dropping it to the floor, and a gap cannot buy its way past the minimum. The tile is covered for rendering its numbers before any readings arrive and for taking a null trend as "no line" rather than substituting one of its own. The sparkline now carries a `data-slot` so a test can tell it from the arrow icon in the delta.

**PersonCoverage.** Tested, and testing it turned up a defect worth more than the missing coverage. The component said "no data reaches us" of every section with no reading, but two different things arrive that way: a source nothing is wired to, and a source that works and holds none of this person's work. The second reader was being told their tracker is broken while their colleagues' numbers for that same section sit one screen away.

The signal was already in the response — a peer view counts the entities that had a reading, so an empty pool across every metric of a section means nobody is measured there, and a pool that reads means the measurement works and this person is absent from it. Two sentences now, never the wrong one. It is not proof (a wholly idle cohort looks the same) but it fails toward the weaker claim, which is the safe direction.

966 tests pass across both projects, `tsc -b` and eslint clean.

hello1101n (approved):

PR #2345, opened 2026-08-07
feat(frontend): a section shows the work, not a second ranking

dzarlax:
## What a section is for

The section screen and the drilldown sheet were the same component — the sheet
lifted onto a page with a heading added. A modal answering a click and a
destination reached from the navigation are not the same job, and the seam
showed.

The overview had already said which sections were worth opening. The section
then repeated that judgment, bigger and in red: a hero card labelled TOP ISSUE
above three coloured outliers, most of which the overview had just named. A
reader who came in because Focus Time needed attention was shown Focus Time
needing attention. Every comparison on the screen was against the cohort, so
the screen one level deeper had dropped the only comparison the reader owns.
And most of the class was behind a fold titled by its statistical standing, so
finding a metric meant knowing in advance how it had ranked.

A section is where a number is explained, and the explanation a reader wants is
the work behind it.

## The rule the body follows

Every metric declares, in the results response, how closely it can be read:
the things themselves, the daily readings a ratio was built from, or a daily
counter. Nothing on screen used that. So the section asks each metric and
renders what it gets — commits listed as commits, counters drawn as days, a
share of the day drawn with the denominator it was taken out of, and a plain
statement where a metric offers nothing.

Nothing here is configured per group. A metric that gains detail later gains it
on screen without a layout change.

## What that produced

**The work behind these numbers.** Per headline metric, at its own grain. The
day strip walks the calendar rather than the rows, so quiet days keep their
place — drawn from rows alone, four busy days in a month pack together and read
as a month that was busy throughout. A day with no reading stays a gap and a
measured zero stays a zero; they draw alike and mean opposite things. Hover
reads out in the caption already under the strip, rather than in a native title
on each of thirty-one bars.

**Also measured here.** The rest of the class, alphabetical, in one column.
Three aligned columns read as a table — the eye takes the row first — so an
alphabet running down each column is invisible and the list still looks like a
heap. Rows that hold nothing for this person are left out: a column of dashes
is the fastest way to teach someone to stop reading a list.

**A standing mark on one shared axis.** Distance in cohort spreads —
(value − median) / IQR — puts hours, messages and percentages on one axis. The
axis line is drawn once behind every row, not per row, so everything ordinary
settles onto it and blends. That blending is the point: "I am like the others
here" should cost no attention, and what the eye is left with is the few marks
standing off the line.

**From <systems>.** What the section was watching, assembled from breakdowns it
had already fetched. A low number means one thing when the tool everyone uses is
connected and another when it is not, and nothing on the screen let a reader
tell those apart.

**Composition, and no verdict.** The headline cards keep their totals and their
breakdowns and lose their status stripe and coloured value. In a section that
withholds judgment, a coloured card makes the judgment anyway — louder than any
sentence beside it, because colour is read before words are.

## Deliberate refusals

- The axis is **never flipped by direction**. Lower left, higher right. Orienting
  it so that left means worse would put the verdict back in through the
  geometry, and a metric with no better direction has no worse to point at.
- **Colour only where the shared standing already calls a reading bottom of the
  pack.** Standing out is not being wrong: eleven spreads above the middle is
  remarkable on any metric and a problem on almost none. Two coloured marks in a
  list of nineteen is a finding; nineteen is a scoreboard.
- **An extreme reading pins to the edge in a different shape** rather than
  setting the scale for every other row.
- **Where nothing is comparable there is no mark**, though the element still
  renders — so an absent dot cannot be misread as a dot at the middle.
- **No mark in the headline blocks.** It reads against the line the list draws
  behind it; three rows that already state both comparisons in words are not the
  problem it solves.
- **A metric a chart already draws over time gets no strip beside it** — the
  chart carries the split a strip cannot. Event-grade metrics are exempt: a list
  of the things themselves is a different answer, not a finer drawing.

## Empty sections

A class with nothing recorded said so once per metric, in four wordings, down a
full page of placeholders. It now says it once, and skips composition entirely.

## Verification

966 tests across both projects, `tsc -b` and eslint clean. Checked against a
running stand on every section, for people with data in most classes and for
people with none.

## Summary by CodeRabbit

- **New Features**
  - Added detailed metric activity views with daily trends, event evidence, comparison text, and empty-state messaging.
  - Added peer-comparison markers showing median, spread, and relative standing.
  - Added an index of additional measured metrics with values and peer comparisons.
  - Added source attribution and “work behind these numbers” activity sections.
  - Added clearer period, loading, error, and no-data states.
- **Improvements**
  - Refined drilldowns to support explanatory content and more relevant metric filtering.
  - Removed status-based coloring from metric summary cards for a more consistent presentation.

### DIFF
--- a/src/frontend/src/components/portal/single-group-view.test.tsx
+++ b/src/frontend/src/components/portal/single-group-view.test.tsx
@@ -1,0 +1,148 @@
+// @vitest-environment jsdom
+/**
+ * One activity class, composed.
+ *
+ * The choices this file makes are all about what NOT to draw: which metrics
+ * earn a detail block, when a whole class collapses to a sentence, and what the
+ * index is allowed to repeat. Each of those is invisible in the output when it
+ * works, which is why they are pinned here.
+ */
+import { render, screen } from "@testing-library/react";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+import type { MetricResult } from "@/api/metric-results-client";
+import { normalizeMetricResults } from "@/lib/metrics/collection";
+
+const mocks = vi.hoisted(() => ({
+  byKey: new Map<string, unknown>(),
+  isPending: false,
+  isError: false,
+  cohort: [] as string[],
+}));
+
+vi.mock("@/queries/metric-results", () => ({
+  useMetricCollection: () => ({
+    byKey: mocks.byKey,
+    previousByKey: null,
+    isPending: mocks.isPending,
+    isFetching: false,
+    isError: mocks.isError,
+    refetch: vi.fn(),
+  }),
+}));
+vi.mock("@/lib/portal/use-person-cohort", () => ({
+  usePersonCohort: () => mocks.cohort,
+}));
+vi.mock("@/hooks/use-portal-period", () => ({
+  usePortalPeriod: () => ({
+    period: "month",
+    dateRange: { from: "2026-03-01", to: "2026-03-31" },
+  }),
+}));
+// The composition blocks are someone else's tests; stubbing them keeps this
+// file about which children get built, not how they paint.
+vi.mock("@/components/widgets/metric-views/collection-drilldown", () => ({
+  CollectionDrilldown: ({ children }: { children?: React.ReactNode }) => (
+    <div data-testid="composition">{children}</div>
+  ),
+}));
+vi.mock("@/components/widgets/metric-views/metric-activity", () => ({
+  MetricActivity: ({ metric }: { metric: { label: string } }) => (
+    <div data-testid="activity">{metric.label}</div>
+  ),
+}));
+
+import { SingleGroupView } from "./single-group-view";
+
+const ME = "019e27bc-dec0-7626-81a9-c5524662a6a9";
+
+function metric(key: string, label: string, value: number | null) {
+  return {
+    metric_key: key,
+    label,
+    unit: null,
+    format: "integer",
+    computation: "sum",
+    direction: "higher_is_better",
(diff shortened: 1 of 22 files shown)


dzarlax:
Two of the three findings were right and are fixed; the third I am declining, with a reason.

**The strip took its peak from every reading handed in, not from the days it draws.** Correct, and worse than it looks: a reading outside the window set the height of bars it is not drawn beside, so *every* visible bar read shorter than it is. The peak now comes from the calendar the strip walks, with a test carrying an out-of-range reading larger than anything inside.

**The strip answered a pointer and nothing else.** Also correct. It now carries its own description — the span, the busiest day, and how many days hold no reading. I did not make the days focusable, which was the suggestion: thirty-one tab stops per strip and three strips to a section is a worse answer for a keyboard reader than one sentence, and the per-day readout is an enhancement over content the header and caption already state. The description says that shape out loud rather than making it navigable.

**Declining the median-unit change.** `kpi-row.ts` builds its median label with the same conditional — plain number for a non-percent metric, the unit only where the format carries it. Both consumers of this string render it beside the value, which does carry the unit: an index row reads `26 days · median 25`, and the activity header reads `424 messages` above `team median 512`. Adding the unit here and not to the tiles would make two surfaces state the same metric differently, which costs more than the ambiguity it removes. Worth changing in both places or neither — and that is a separate change to the shared convention, not this PR.

Also covered the parts of this PR that had tests for their arithmetic and none for their decisions: the query hook's refusals, each grain rendering as itself, and the composition collapsing a silent class to one sentence rather than four placeholders.

979 tests pass, `tsc -b` and eslint clean.

hello1101n (approved):

PR #2347, opened 2026-08-09
ci: deploy + seed a test stand after the umbrella chart publish on main

### LINKED ISSUE #2244: ci: deploy + seed a test stand after the umbrella chart publish on main
## Goal

Validation of a merge to `main` currently ends at "images built, chart published" (`build-images.yml`). Add the next step: install the just-published chart on a test stand, seed it, and prove a user can log in and see data — catching broken-but-installable releases before anyone deploys them.

## Scope

- New workflow (or a job after the chart publish in `build-images.yml`), on merge to `main`: `helm upgrade --install` the just-published chart version against a dedicated test cluster/namespace.
- Run the seed Job from the toolbox image (depends on #2243) with a stand-scoped `TENANT_DEFAULT_ID` and `DEV_USER_EMAIL`.
- Smoke gate: `/auth/login` redirects to the IdP; scripted password login as `DEV_USER_EMAIL` succeeds; `/auth/me` returns 200 with that email; at least one metric returns data for the seeded window.
- Failure of any step fails the workflow visibly on the merge commit.

## Acceptance criteria

- [ ] Merge to `main` → chart publish → automated deploy of that exact chart version to the test stand.
- [ ] Seed + login + data smoke checks run and gate the workflow result.
- [ ] Test-stand kubeconfig/credentials live in GitHub environments/secrets, not the repo.

ktursunov:
Closes #2244.

Merge-to-`main` validation currently stops at "images built, chart published". This adds the next step: install that exact chart on the shared test stand, seed it, and prove a person can sign in and see data.

## What this is

An **after-merge alarm, not a gate** — and it cannot be anything else. The artefact under test does not exist until `publish-chart` has run, and that runs on push to `main`. Merge blocking stays with the existing merge-queue lanes; this guards the claim that what those lanes approved actually installs and works. A red run means `main` is broken: the merge author owns it, fixes forward if the cause is small, and reverts if it is not.

## Shape

`build-images.yml` grows an `outputs:` block on `publish-chart` and one calling job that hands `chart_version` to a reusable `deploy-test-stand.yml`, `main` only. That workflow runs three separately-named stages with their own timeouts, so a red run says which one died without opening a log:

| stage | what it does |
| --- | --- |
| deploy | `helm upgrade --install` at that exact version through `deploy/gitops/environments/test-stand/`, `--wait`, **no `--atomic`** |
| seed | `src/ingestion/tools/seed/seed-stand.sh` verbatim, 730-day window |
| smoke | `tests/stand/smoke` against the public URL — IdP redirect, persona login, `/auth/me`, one metric with data |

Smoke never runs after a failed seed. Deploys coalesce (`cancel-in-progress: false`) so a running upgrade is never killed mid-flight.

## Design notes worth a reviewer's attention

**No `--atomic`.** A rollback destroys the evidence and leaves the stand silently running an older chart than `main`, so the next merge's smoke passes against stale code. A failed upgrade is left failed; recovery is the next merge or a deliberate `make rollback`.

**Not `make deploy`.** That target chains `apply-app-secrets`, which hard-requires a sealed manifest this cluster has no controller for, then overwrites the chart-owned `*-config` Secrets from a key name this stand does not use — writing a blank OIDC client secret and breaking every login while the release still reports `deployed`. It also hardcodes `--atomic` where no variable can clear it. `make diff`, `make status` and `make rollback` do work for this environment.

**Umbrella-only gitops environment.** `bootstrap.*` and `system.*` are all false: the stand's datastores are operator-managed in their own namespaces and its edge is Envoy Gateway with HTTPRoutes the chart does not render. The values keep `credentials.deploymentMode: helm` / `autoGenerate: true` to match the live release, so the chart's `lookup` finds the pre-created, label-less `insight-db-creds` and stands down rather than rotating live datastore passwords.

**Public logs.** This repo is public, so run logs are public forever. All cluster and seeder output goes through `redact-stand-log.py`; failure diagnostics are an allowlist (`stand-diagnostics.sh`) — release status, pod table, warning reasons without messages, tail of failed containers only. No describes, no environment dumps, no uploaded artefacts. The seed manifest names every persona, which is also why the three stages are steps in one job rather than three jobs passing an artifact.

## Evidence

Rehearsed read-only against the stand with `deploy/gitops/scripts/emulate-ci-deploy.sh`, which runs these same commands from a laptop and prints a parity report against the workflow file:

- cluster guard, namespace reachability and an RBAC probe (get secret / get configmap / patch deployment / create job / get httproute) all pass
- `helm template` with the new values renders 41 objects at the current chart version
- both edge HTTPRoutes report `Accepted=True`; the OIDC client Secret is present (presence only — never read or printed)
- the seeder's own `--dry-run` resolves the tenant, the datastore coordinates, the seeder image and the IdP source type **from the cluster**, so no tenant is supplied from outside
- **a server-side dry run renders exactly the live release's object set — nothing added, nothing removed**, confirming the values file cannot damage the stand

## Not in scope (follow-ups)

No alerting on a red run, no rollback, no infrastructure bootstrap, no `rollout restart` after the upgrade, no exact-value assertions, no second environment. Each is a deliberate omission with reasoning in the workflow header, not an oversight.

## Before this can go green

- [ ] A repo admin creates the `insight-test-stand` GitHub environment restricted to `main` and sets its secrets — I have `push` but not `admin` on this repo. Runbook: `docs/components/deployment/specs/sop/credentials-runbook.md`
- [ ] Run `deploy/gitops/scripts/provision-ci-deployer.sh --apply` to mint the namespace-scoped ServiceAccount credential (dry-run by default; asserts its own containment)
- [ ] **Open design question:** the stand's authenticator uses the `insight-broker` realm with `sourceType: github`, which federates login to GitHub — so there is no local password for a scripted multi-persona login. Options are laid out in the smoke suite's README; this needs a decision before the smoke stage can pass.


.github/workflows/scripts/extract-seed-manifest.py
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env python3
```
ktursunov, line 1: is that script is really needed?

.github/workflows/scripts/redact-stand-log.py
```diff
@@ -0,0 +1,225 @@
+#!/usr/bin/env python3
```
ktursunov, line 1: Where does this credentials appear - let's use filtering in tests

.github/workflows/scripts/stand-diagnostics.sh
```diff
@@ -0,0 +1,236 @@
+#!/usr/bin/env bash
```
ktursunov, line 1: can it be part of helm deploy validation?

ktursunov:
Superseded by #2404, which is raised from a branch in this repository rather than from a fork.

The reason is mechanical: GitHub withholds environment secrets from fork pull requests, so the `insight-test-stand` environment could never be read by a run on this PR and the deploy stage could not be exercised here. Same branch, same commits, same head name.

Closing in favour of #2404.

Issue #1669, opened 2026-07-07
backend(analytics): OpenAPI spec over-declares boilerplate error codes and under-declares real ones (.standard_errors)


ktursunov:
## Summary

The analytics OpenAPI registrations in `src/backend/services/analytics/src/api/mod.rs`
build every route with `OperationBuilder::...standard_errors(openapi)`, which stamps a
uniform error set — `{400, 401, 403, 404, 409, 429, 500}` — onto **all 20 operations**.
The committed spec (`docs/components/backend/analytics/openapi.json`, generated by the
`analytics openapi` subcommand and drift-gated by `openapi-specs.yml`) therefore does not
describe what the handlers actually return: it over-declares codes no handler can produce
and omits codes handlers do produce.

## Over-declared (declared, but the handler can never return them)

- **401** — the gateway runs with authn disabled in every environment we ship; no analytics
  route returns 401.
- **429** — there is no rate limiter in front of these handlers; no route returns 429.
- **403** — only reachable on the admin-threshold write routes (lock / cross-tenant
  ownership); declared on all 20.
- **409** — only reachable on `POST /v1/admin/metric-thresholds` (duplicate) — and that path
  currently 500s instead, see #1664; declared on all 20.
- **404** — declared on collection / param-less routes (`GET /v1/metrics`, `POST /v1/metrics`,
  `GET /v1/columns`, `POST /v1/catalog/get_metrics`, …) that resolve no entity and cannot 404.
- **400** — declared on routes with no validated input (`GET /v1/metrics`, `GET /v1/columns`).

## Under-declared (returned by the handler, absent from the spec)

- **415** — every body endpoint returns 415 on a wrong/missing `Content-Type` (both the
  `CanonicalJson` and the legacy `axum::Json` extractors).
- **422** — the six legacy body endpoints return 422 on a well-formed-but-off-schema body
  (tracked separately as the canonicalization bug; the *intended* code there is 400).
- **503** — `POST` / `PUT /v1/admin/metric-thresholds` return 503 when the lock-audit sink is
  unavailable.
- Per-route **403 / 409** on the admin routes where they *are* reachable are folded into the
  boilerplate rather than declared deliberately.

## Impact

The published contract misinforms consumers, and it defeats per-status-code coverage
checking: a black-box e2e suite cannot exercise 401/403/429 (unreachable), so the endpoint
coverage gate (`src/ingestion/tests/e2e/lib/api_coverage.py`) has to carry a large `BLOCKED`
[...]

PR #2349, opened 2026-08-09
feat(identity): publish the identity-resolution OpenAPI contract and gate it

mozhaev-dev:
## Why

The committed `docs/components/backend/identity-resolution/openapi.json` was still the retired .NET contract: it declared routes the Rust service answers 404 for, omitted ones it serves, described none of the operator correction surface, and nothing in CI noticed. The service had no way to publish its own contract, so there was no path back to accuracy.

## What

**The service emits its own document.** New `openapi` subcommand and `api::openapi_document()`, mirroring analytics and authenticator — offline, no backends, built from the very `build_operations` route table the live gear serves, so the document and the router cannot diverge. It is genuinely offline: the config is now loaded per command, so the emit does not depend on a file it never reads.

**The contract is accurate again.** 15 paths became 21. The seven operator correction routes and both persons-sync operations appear for the first time; `/v1/persons/{email}` and `POST /v1/persons-seed` are gone, the service having retired them; the subchart parameter is spelled `{person_id}`, as the route actually is. The internal S2S resolvers stay out, registered raw exactly so — a test holds that.

**Parameters are described, not just paths.** Every templated route named its parameters in the path and described none of them, which is not a valid document: a generated client has nothing to fill `{source_id}` from. All path parameters are declared now, along with the filters the list handlers accept (`status`/`limit` on both journals, `person`/`role`/`active`/`limit` on role assignments, `viewer`/`viewed`/`active`/`limit` on visibility grants, `depth`/`valid_at` on both subchart reads, `limit` on the review queue) and the optional revoke-reason body both DELETE routes accept. A test holds the templated-path invariant for the whole table.

**It cannot go stale again.** Drift gate added to `openapi-specs.yml` beside the analytics and authenticator ones.

**The stand models are generated from it.** `generate_schemas.py` carried a tripwire for exactly this moment — identity was listed as `Untrusted` because its document was the .NET one, and `--check` fails that entry the moment it stops being true. It now does, so the entry is promoted to `Generated`. The models stop being a transcription of the Rust DTOs and become a contract test: a validation failure now says the service and its published document disagree.

Two consequences handled here:

- The `/internal/persons/*` resolvers are excluded from the document by design, so `IdentityValue` has no generated counterpart and moves to `identity_internal.py`, still hand-written.
- The hand-written models carried behaviour as well as fields (`walk()`, `emails()`, `in_force`). Generated models carry the contract and nothing else, so that behaviour moves to free functions in `stand/api/identity/views.py`, beside the tests that use it. Deliberately not wrapper subclasses: those must re-declare the fields they keep, and a hand-maintained field list beside a generated one is the drift the generator exists to prevent.

The generated names are the contract's (`SubchartResponse`, not `Subchart`); the schemas package re-exports them under the names the suite already uses, so the rename stops at that package.

## Verification

| Check | Result |
|---|---|
| OpenAPI drift gate | matches the live spec |
| Generated models gate | rc=0 |
| ruff over `tests/` | passed |
| rustfmt / clippy | clean |
| identity-resolution unit tests | 128 passed |
| stand suite collection | 293 collected |
| helpers against the generated models | walk / emails / in_force exercised on instances |

`--config <missing> openapi` now succeeds and emits the document; commands that do need a config still validate the path exactly as before.

**Not verified locally:** the four affected stand suites need a running stand, so they did not execute here — the layer that changed (models plus helpers) was verified by instantiating them. CI covers the rest.

## Notes for review

- `Untrusted` in `generate_schemas.py` now has no entries. Kept rather than deleted: it just did its job, and the next service arriving with a foreign contract is its user. Happy to remove it if reviewers prefer.
- The coverage gate still does not check identity against its spec. The old reason (a .NET document) is gone; what remains is that `.standard_errors` stamps the full error set onto every operation, so gating would demand codes nothing observes. The stale justification in `coverage.py` is corrected; turning the gate on is a change to the gate (#1669), not to this PR.
- `main.rs` now loads config per command, which diverges from analytics and authenticator. The divergence is deliberate — they share the same wart, and pulling them along belongs in its own change.

## Summary by CodeRabbit

* **New Features**
  * Added an offline command to generate the identity-resolution OpenAPI document as formatted JSON.
  * Expanded API documentation for identity resolution, roles, visibility, journals, and subcharts, including request and response details.
  * Added generated, strongly typed API schemas covering identity operations, profiles, roles, visibility, accounts, and errors.

* **Documentation**
  * Added identity-resolution to automated OpenAPI drift checks, with stale-spec reporting and results in workflow summaries.

* **Tests**
  * Updated identity API tests and shared helpers to reflect the documented API contract.

### DIFF
--- a/.github/workflows/openapi-specs.yml
+++ b/.github/workflows/openapi-specs.yml
@@ -10,6 +10,9 @@ name: OpenAPI Specs
 #     no HTTP listener; see api::openapi_document).
 #   • authenticator — the offline `authenticator openapi` subcommand (no Redis,
 #     no IdP; see api::openapi_document), same shape as analytics.
+#   • identity-resolution — the offline `identity-resolution openapi`
+#     subcommand (no MariaDB, no ClickHouse; see api::openapi_document),
+#     same shape as analytics.
 
 on:
   # Deliberately NO `paths:` filter — this is meant to be a required status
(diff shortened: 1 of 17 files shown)


mozhaev-dev:
Addressed in `f90dee3f`. Verified the clamp constants rather than assuming them: all four list routes share `LIST_DEFAULT_LIMIT = 50` and `LIST_MAX_LIMIT = 500` (`clamp_limit` in `person_roles.rs` / `visibility.rs`, and the same pair in `seed.rs` / `sync.rs`), so the descriptions now read `(1..=500, default 50)` — not the `default 100` the suggestion proposed. The review queue keeps its own `(1..=1000, default 100)`, which matches `DEFAULT_QUEUE_LIMIT` / `MAX_QUEUE_LIMIT`.

cyberantonz (approved):

PR #2387, opened 2026-08-10
fix(reconcile): convert Secret strings to spec-declared types

### LINKED ISSUE #2381: github-v2 source creation is always rejected for the documented Secret, and reconcile reports success
## Summary
A github-v2 Secret written exactly as the connector README documents can never be provisioned: Airbyte rejects the source creation, and the reconcile loop still finishes with "0 change(s), 0 error(s)", so the connector silently never comes into existence.

## Steps to Reproduce
1. Create the connector Secret following `src/ingestion/connectors/git/github-v2/README.md` verbatim (`github_skip_archived: "true"`, `github_skip_forks: "true"`, as in the example).
2. Run the reconcile loop (scheduled tick or a manual run) and watch its log.
3. The source-create call is rejected with HTTP 422, the log prints `CHANGE github-v2: source  created` with an empty source id, then `WARN github-v2: source not yet created (will be on real run) — skipping connection setup`, and the run ends `reconcile finished: 0 change(s), 0 error(s)`.

**Expected:** the source, connection and sync CronWorkflow are created, or the run fails loudly naming the rejected field.
**Actual:** no source, no connection, no CronWorkflow — and the run reports success, every 15 minutes, indefinitely.

Reproduces on a fresh install with a single github-v2 Secret and no other connector state.

## Additional information
- Airbyte's response to `POST /api/v1/sources/create`, verbatim:
  ```
  422 The provided configuration does not fulfill the specification. Errors: json schema
  validation failed when comparing the data to the json schema.
  Errors: $.github_skip_forks: string found, boolean expected,
          $.github_skip_archived: string found, boolean expected
  ```
- The same create with `github_skip_archived`/`github_skip_forks` as JSON booleans (all other fields identical, token replaced by a dummy — sources/create only schema-validates) returns 200 and creates the source.
- `source_github_v2/spec.json` declares both fields `"type": "boolean"`; the README example and `src/ingestion/secrets/connectors/github-v2.yaml.example` both write them as the strings `"true"` — a Kubernetes Secret can only carry string values.
- `descriptor.yaml` lists both fields in `secret.required_fields`, so omitting them from the Secret makes the reconcile skip the connector at validation instead (`MISSING: github_skip_archived`).
- The github-v2 definition itself is registered correctly (`source_definitions/list_for_workspace` shows it at the descriptor's pinned image tag); only the source creation fails.
- The reconcile treats the failed create as non-fatal: the curl error surfaces as one `curl: (22) ... error: 422` line, the empty source id takes the `WARN ... skipping connection setup` branch, and the run's summary counts 0 errors.

## Notes
related to #2279

aleksdotbar:
## Problem

A connector Secret can only hold strings, while connector specs declare boolean, integer and array fields. Reconcile posted the raw strings, so `sources/create` was rejected with 422 and the connector was never provisioned. The create result went unchecked, so the tick still ended `0 change(s), 0 error(s)` — every 15 minutes, indefinitely.

Affects every connector with a typed field: `bitbucket_skip_forks`, `bitbucket_concurrency`, `bitbucket_workspaces`, `gitlab_groups`, `gitlab_projects`, `gitlab_max_workers`, `ad_use_ssl`, `ad_port`, `ad_page_size`, `metrics_lookback_days`, `github_organizations`, `github_skip_archived`, `github_skip_forks` and the `*_page_size` fields on zoom, youtrack, figma, confluence and outline.

## Fix

- Secret values are converted to the type the connector's spec declares (`source_*/spec.json` for cdk, `connector.yaml` for declarative). A value that cannot be converted fails the run naming its field instead of returning an opaque 422.
- Source create, update and recreate, and the recreate path's connection create, now check the API result and the returned id. Previously an empty connection id skipped state restore and still logged `state preserved`, so sync cursors were lost silently.
- Docs: list fields are written as JSON arrays, plus a value-format rule and a troubleshooting row in `CONNECTORS.md`.

Connector-side config changes for github-v2 are out of scope here; this PR is the reconcile fix only.

## Validation

`shellcheck` clean on `reconcile.sh`, `ruff` clean on both new modules, and the conversion is exercised against the real github-v2, gitlab and github-directory specs: typed strings, pre-parsed arrays, bad int, bad bool, empty value, unknown key, missing spec, non-object stdin, a number that overflows to infinity, and a spec type outside the JSON Schema set. No connector spec in the tree declares a type outside that set.

Closes constructorfabric/insight#2381

## Summary by CodeRabbit

* **Improvements**
  * Improved connector reconciliation with clearer API error messages and safer recovery of existing connection state.
  * Enhanced source configuration to apply injected settings and convert secret values according to connector requirements.
  * Improved dry-run behavior and connection setup safeguards, including validation of failed operations and missing identifiers.

* **Documentation**
  * Added guidance for quoting secret, numeric, and boolean values.
  * Documented JSON-array syntax, default handling, empty-value rejection, and troubleshooting steps.
  * Updated GitHub and Bitbucket examples to use JSON arrays.


src/ingestion/reconcile-connectors/python/extract_api_error.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python3
+"""Print the human-readable message from an Airbyte API error body.
```
cyberantonz: Why you need this script at all? Shell `jq` works pretty well

src/ingestion/reconcile-connectors/python/build_source_config.py
```diff
@@ -0,0 +1,205 @@
...
+        manifest = yaml.safe_load(manifest_path.read_text(encoding="utf-8")) or {}
+    except (OSError, yaml.YAMLError) as e:
+        raise SpecUnavailable(f"cannot read {manifest_path}: {e}") from e
+    return (manifest.get("spec") or {}).get("connection_specification") or {}
+
+
+def _parse_scalar(raw: str, declared: str) -> JsonValue:
+    text = raw.strip()
+    if declared == "boolean":
+        if text.lower() in ("true", "false"):
```
cyberantonz: text.lower() called twice without a reason

src/ingestion/reconcile-connectors/python/build_source_config.py
```diff
@@ -0,0 +1,205 @@
...
+            return text.lower() == "true"
+        raise ValueError('expected "true" or "false"')
+    if declared == "integer":
+        if not _INT_RE.match(text):
```
cyberantonz: The `int(str)` conversion fires an exception if conversion failed, the regex seems not needed

src/ingestion/reconcile-connectors/python/build_source_config.py
```diff
@@ -0,0 +1,205 @@
...
+            raise ValueError("expected an integer")
+        return int(text)
+    if declared == "number":
+        if not _NUMBER_RE.match(text):
```
cyberantonz: Same here, the regex is not needed

src/ingestion/reconcile-connectors/python/build_source_config.py
```diff
@@ -0,0 +1,205 @@
...
+        key: coerce_field(key, value, (properties.get(key) or {}).get("type")) for key, value in secret_data.items()
+    }
+    config["insight_tenant_id"] = tenant_id
+    config["insight_source_id"] = source_id
+    return config
+
+
+def _fail(reason: str, code: int) -> int:
+    sys.stderr.write(f"build_source_config: {reason}\n")
+    sys.stdout.write(reason)
```
cyberantonz, line 211: Why to write error reason twice to stdout & stderr?

cyberantonz (changes requested):

PR #2390, opened 2026-08-10
ci(security): raise an issue for what the nightly scan reports, and run in the merge queue

Gregory91G:
Refs #2301. Builds on #2351, now merged.

A finding that reaches `main` is announced exactly once, in the pull request that introduced it, and never again: the blocking job only ever scans a pull request's own commits, so later pull requests never see it. The nightly job does see it, every night, for as long as it is there — but it reports and exits zero, and nobody opens a run summary. A missed secret therefore looks the same as a clean repository.

Two changes.

## An issue per finding, assigned to whoever brought it

`trufflehog_gate.py` already computes the findings that are not in the baseline; it now also writes them to a JSON file when a sixth argument is given. Values are not included — the file carries the detector, the path, the line, the commit and the fingerprint.

`trufflehog_notify.py` reads that file, resolves each commit to a GitHub login through the commits API, and opens one issue per fingerprint, assigned to that person. One issue per fingerprint rather than a digest, because they close independently and each belongs to a different author; the fingerprint sits in the title, so an already-open issue suppresses the next night's duplicate.

The assignee is resolved in two steps. GitHub can name a commit's author only when the address inside the commit is attached to an account, and a work address often is not — a quarter of this repository's commits resolve to nobody that way. The pull request a commit arrived with names an account either way, so it is the second route, and it covers every commit that reaches `main` through review.

Neither the issue nor the scripts name an address. The body links the GitHub profile of whoever is resolved and the commit itself; when nothing resolves — a commit pushed straight to a branch — it says so and links the commit, which is where the author is visible anyway. An earlier revision carried a file mapping work addresses to logins; it is gone, because a single harvestable list of live addresses in a public repository is worth more than the historical commits it covered.

The fix for an address that resolves to nobody belongs with its owner: adding it under [Settings → Emails](https://github.com/settings/emails) attributes it, past commits included, and needs no table here. Attribution works from any verified address on an account, not only the primary one.

## The blocking job runs in the merge queue

A required status check has to produce a result inside the merge queue, where there is no `pull_request` event. The job now also triggers on `merge_group`, and takes its scan range from `github.event.merge_group.base_sha` rather than `HEAD^1` — on a queue branch the first parent is the previous entry, not the base.

The two nightly jobs are pinned to `schedule` and `workflow_dispatch`. They used to run on anything that was not a pull request, which would have meant a full-history scan and a GitHub API sweep on every entry the queue builds.

This does not make the check required — that is a ruleset change, and it belongs after this has run in the queue at least once.

## Test plan

- [x] The gate writes the JSON only when asked, and its verdict is unchanged: 272 known findings against the seeded baseline → exit 0, empty list; the same corpus against an empty baseline → 272 entries, keys `detector, verified, commit, file, line, email, fp`
- [x] `trufflehog_notify.py --dry-run` over real findings assigns through both routes: `aleksdotbar` by commit author, `ktursunov` by pull request author, and a commit with neither is left unassigned
- [x] Author resolution measured over all 53 distinct commits behind the findings: 35 by commit author, 10 by pull request author, 8 by neither
- [x] No address appears in the issue body, the scripts or the JSON the gate writes
- [x] The commit id reaches the API whole; only the summary table and the issue body shorten it
- [x] `python -m py_compile` on both scripts; the workflow parses and keeps its five, four and two steps per job
- [x] `actionlint` 1.7.12 on this workflow and on every workflow in the repository — exit 0
- [ ] `secrets (diff)` green on this pull request
- [ ] The job produces a result on a `merge_group` event — only observable once this is on the default branch
- [ ] A nightly run with a finding outside the baseline opens an assigned issue

## Summary by CodeRabbit

* **New Features**
  * Added optional redacted JSON output for newly detected secrets.
  * Added automated issue reports for unaddressed findings, with dry-run previews and author attribution.
  * Added merge queue scanning support.
  * Added duplicate-finding detection to prevent repeated issue reports.

* **Bug Fixes**
  * Limited historical and API scans to scheduled or manually triggered runs.
  * Improved tracking of findings against an established baseline.
  * Preserved complete commit identifiers in generated finding data.


cyberantonz (approved):

cyberantonz (changes requested):
You provide real email addresses of real people. I think it is not the right way

.github/scripts/trufflehog_notify.py
```diff
@@ -0,0 +1,129 @@
...
+        "",
+        "- **It is a credential.** Revoke and rotate it. Removing the commit is not remediation —"
+        " assume it was compromised the moment it was pushed.",
+        "- **It is a false positive.** Add the fingerprint to `.github/trufflehog-allowlist.txt`"
+        " with a reason, in a pull request, so the claim is reviewed.",
+    ])
+
+
+def main() -> int:
+    new_json, repo, author_map = sys.argv[1], sys.argv[2], sys.argv[3]
```
cyberantonz: There is argparse library for this

.github/scripts/trufflehog_notify.py
```diff
@@ -0,0 +1,129 @@
...
+            print(f"would open: {title}  -> assignee: {login or '(none)'}")
+            created += 1
+            continue
+        args = ["issue", "create", "--repo", repo, "--title", title, "--body", body_for(f, login)]
+        if login:
+            args += ["--assignee", login]
+        print(gh(*args) or f"failed to open an issue for {f['fp']}")
+        created += 1
+
+    print(f"{created} issue(s) {'would be ' if dry else ''}opened, {skipped} already tracked")
```
cyberantonz: No pre-commit installed?

Gregory91G:
Emails are removed, now attributions of issues are based only on emails registered in the accounts

.github/scripts/trufflehog_notify.py
Gregory91G: thanks, fixed

.github/scripts/trufflehog_notify.py
Gregory91G: thanks, fixed

cyberantonz (approved):

Issue #2266, opened 2026-08-06
Image CVE gate fatals on attestation digests and blocks merge jobs — temporarily disabled, must be fixed and re-enabled


cyberantonz:
## What happened

The image CVE gate (`.github/actions/image-cve-gate`) fails merge jobs in **Build & Push Container Images** with a trivy fatal error, not with a real vulnerability finding:

```
FATAL Fatal error run error: image scan error: ... unable to find the specified image
"ghcr.io/constructorfabric/insight-authenticator@sha256:818f5340..." in ["docker" "containerd" "podman" "remote"]:
* remote error: no child with platform linux/amd64 in index ghcr.io/constructorfabric/insight-authenticator@sha256:818f5340...
```

Example failing run: https://github.com/constructorfabric/insight/actions/runs/31075495062/job/92536985267 (`merge-authenticator`; the first digest scanned clean with 0 findings, the second digest is not a scannable image).

## Root cause

The gate scans **every** file in `/tmp/digests`, but the build jobs (buildx with provenance/SBOM enabled) also push digests that are attestation/provenance manifests rather than per-arch container images. Trivy cannot scan those and exits 1, which fails the merge job even when the actual images are clean.

## Temporary mitigation

The gate has been made **non-blocking** (scan errors and findings are reported as warnings but no longer fail the job) so merges are not blocked. This means fixable CRITICAL CVEs currently do NOT block image tagging — this issue tracks restoring the enforcement.

## Proper fix

- In `image-cve-gate`, only scan digests that are real per-arch images: inspect each digest's media type / platform (e.g. `docker buildx imagetools inspect` or the registry manifest media type) and skip attestation/provenance manifests (`application/vnd.in-toto+json` layers, `vnd.docker.reference.type=attestation-manifest`), or have the build step export only image digests into the digests dir.
- Re-enable `--exit-code 1` blocking behavior once the scan no longer fatals on non-image digests.

Gregory91G:
To Verify:
https://github.com/constructorfabric/insight/actions/runs/31489589971, job `merge-authenticator` has passed
gate is no longer failed and stays where it should be

PR #2396, opened 2026-08-10
ci(security): give the image scans their own code-scanning tool (#2266)

Gregory91G:
Splits the noisy half of the `Trivy` check away from the useful half, without giving up the image findings. Follow-up to #2266.

## The mechanism

Code scanning creates one check **per tool**, not per category, and reports that tool's missing configurations. Today the `Trivy` tool owns nineteen categories:

| category | produced by | present on a PR |
|---|---|---|
| `trivy-fs` | `trivy.yml` (pull_request, push, schedule) | yes |
| `trivy-helm` | `trivy.yml` | yes |
| `trivy-image:*` × 17 | `trivy-images.yml` (**schedule only**) | never |

`trivy-images.yml` scans *published* images, which a pull request does not have, so those seventeen can never be produced anywhere except the nightly cron. The `Trivy` check therefore reports them missing on every analysis this repository uploads — on PRs and on pushes to `main` alike:

```
#2371  Trivy  failure  17 configurations not found
#2367  Trivy  neutral  17 configurations not found
main   Trivy  neutral  17 configurations not found
```

`Semgrep OSS` sits beside it reporting `success — No new alerts in code changed by this pull request`, for one reason only: its single category is produced everywhere.

The statement is accurate and permanently unactionable, and it costs more than noise — the check is genuinely useful for `trivy-fs`, `trivy-helm` and `semgrep`, so if one of those ever stopped uploading, this is what would say so, under seventeen bullet points nobody reads.

## What this PR does

One step, before the upload, rewrites `runs[].tool.driver.name` in the image SARIF from `Trivy` to `Trivy Images` with jq. That is the field code scanning keys the tool on, so the two sets separate:

- **`Trivy`** keeps `trivy-fs` and `trivy-helm` — both produced by every PR, so its check goes clean, like Semgrep's.
- **`Trivy Images`** owns the seventeen image categories, produced by the nightly.

## What stays exactly as it is

The scan, the per-image summary table, `Fail on a fixable critical` (#2238), the per-image `category:`, the `.trivyignore` handling, and the alerts themselves. The image findings remain in the Security tab — only the tool they are filed under changes. `+14` lines, nothing removed.

## What to look at in the nightly

<https://github.com/constructorfabric/insight/actions/workflows/trivy-images.yml> — filter *Event: schedule*. One job per image; the run page's **Summary** block carries a table per image, grouped by package:

| Ecosystem | Package | Max | CVEs | Fixed in |
|---|---|---|---|---|

Latest run for reference: <https://github.com/constructorfabric/insight/actions/runs/31359228269>, and its `insight-toolbox` job <https://github.com/constructorfabric/insight/actions/runs/31359228269/job/93364623071> — that image is where 10 of the 11 open image alerts sit.

A red job in that workflow means a fixable **critical** appeared in a published image. Everything else is reported and not gated there; the gate that blocks a bad build lives in `.github/actions/image-cve-gate`.

Alerts, after this lands, filtered by the new tool:
<https://github.com/constructorfabric/insight/security/code-scanning?query=is%3Aopen+tool%3A%22Trivy+Images%22>

## Test plan

- [x] The rewrite runs against a real Trivy image SARIF: `driver.name` `Trivy` → `Trivy Images`, every other driver field identical, 25 results and 24 rules preserved, still valid JSON.
- [x] jq assignment creates the intermediate objects: a SARIF with no `tool` at all comes out as `{"driver":{"name":"Trivy Images"}}`, so a malformed input is renamed rather than silently skipped.
- [x] Output goes through a temporary file and `mv`; redirecting onto the input would truncate it before jq reads it.
- [x] `.github/workflows/trivy-images.yml` parses; the `scan` job now has six steps in the order Checkout, scan, summary, rename, upload, gate.
- [x] `actionlint`: 8 findings on `main`, 8 on the branch, identical set; none in `trivy-images.yml`.
- [ ] After merge, first nightly: alerts appear under tool `Trivy Images`, the summary tables are unchanged.
- [ ] First PR after that nightly: the `Trivy` check reports `success` instead of listing configurations.

## Known unknown

Whether a check appears for `Trivy Images` on pull requests, where it uploads nothing. Both existing tools always upload, so there is no way to observe this before merging. If one does appear and is red, the fallback is to drop the upload entirely — that was this branch's previous revision and it is one revert away.

## After merge, one admin action

The seventeen categories already recorded under the **old** tool do not move; the `Trivy` check keeps listing them until their analyses are deleted. That is 261 analyses, and `DELETE /repos/.../code-scanning/analyses/{id}` answers `403 You are not authorized to delete analyses` for a `write` account, so it needs someone in @constructorfabric/admins. Order matters: this PR first, deletion after, otherwise the nightly re-registers them under `Trivy` the same night.

## Summary by CodeRabbit

* **Chores**
  * Improved security scan reporting by labeling container image findings as “Trivy Images” in uploaded results.
  * SARIF uploads continue only when a report is available.


ktursunov (approved):

.github/workflows/trivy-images.yml
```diff
@@ -135,6 +135,20 @@ jobs:
           out.close()
           PY
 
+      - name: Give the image scans their own tool name
+        # Code scanning reports missing configurations per tool, and only this nightly
+        # can produce these categories — keep them off the Trivy check PRs see (#2266).
+        if: ${{ always() && hashFiles('trivy-image.sarif') != '' }}
+        run: |
+          python3 - <<'PY'
```
cyberantonz: Why not to use jq?

cyberantonz (changes requested):

.github/workflows/trivy-images.yml
Gregory91G: thanks. fixed

ktursunov (approved):

cyberantonz (approved):

Issue #2344, opened 2026-08-07
Metric definitions carry no subject, so a family's metrics cannot be grouped


dzarlax:
## Summary

A metric definition carries no subject or category, so anything presenting more than a handful of metrics has no ordering to offer beyond the alphabet.

## Steps to Reproduce

1. `GET /v1/metric-definitions` and read a metric's fields: `metric_key`, `label`, `short_label`, `description`, `explanation`, `unit`, `format`, `direction`, `dimensions`, `is_enabled`, `schema_status`, `schema_error_code`, `last_observed_date`, `drilldown`.
2. Look for an attribute that would group the metrics of one family into what a reader would call topics — meetings, messages, email, documents.
3. There is none, in the response or in the `metric_definitions` table behind it.

**Expected:** a surface listing a family's metrics can group them by subject.
**Actual:** it can only sort them by name.

## Additional information

- What exists nearby does not substitute for it:
  - `metric_sources.source_key` names the family the metric belongs to (`collab`, `git`, `ai_usage`, `task`, `wiki`) — the grouping a screen already has from its own layout, not a division within it. It also cannot serve as one in principle: `metric_definition_inputs` binds each input to its own `source_measure_id`, so a metric's numerator and denominator may come from different sources. A grouping derived from the source is therefore not a partition — one metric can fall into two groups at once — even before asking whether the groups mean anything to a reader. A source is in turn a union of connectors, which appear as a dimension value rather than as a property of the metric, so descending to the connector does not recover a partition either.
  - `metric_source_dimensions` declares the dimensions a source can break down by. Within one family this is frequently the same key for nearly every metric, so grouping by it produces one bucket.
  - `unit` splits a family into plausible-looking groups that cut across the subject: an hours-valued metric about meetings lands apart from a count-valued metric about the same meetings, purely because one counts hours.
  - `metric_catalog.source_tags` holds the connector a metric came from. That is provenance, and a screen wanting it can already state it.
- Deriving the grouping in a client means a hardcoded map from metric key to topic. It would be a second definition of the metrics living outside the catalog, and it would silently drop any metric added later into an "other" bucket, with no signal that anything was missed.

## Notes

Blocks grouping in any surface that lists a whole family. Where a fixed set is presented, the alphabet is at least honest.

dzarlax:
Added to *Additional information* after a review point worth recording: a metric is not bound to a single source.

[...]

PR #2401, opened 2026-08-10
feat(analytics): carry subject and tags through custom-metric CRUD

cyberantonz:
> **Stacked on #2397** — builds on `fix/2344-metric-subject-tags`. Review/merge after #2397; until it lands, the diff here also shows #2397's commit. Refs #2344.

## Summary

#2397 gave **builtin** metrics `subject` (grouping partition) and `tags` (cross-cutting filters) via the registry. This PR closes the matching gap for **custom** (tenant-authored) metrics: the CRUD could neither accept nor return them, so a custom metric was stuck with `subject = NULL` and no tags, unreachable from the API.

Both are now threaded through the full custom-metric path (`domain/metric_crud.rs`):

- **DTO:** `subject` (optional slug) and `tags` (slug list) on `CustomMetric`; `subject` on `CustomMetricSummary` so the management list can group by topic like the definitions listing.
- **Validation:** subject shape, tag shape + per-metric uniqueness, a `MAX_TAGS` bound — mirroring the existing measures/dimensions rules.
- **Create / import:** `subject` on the `metric_definitions` insert; a `metric_definition_tags` insert loop mirroring the dimensions loop.
- **Read / export:** `subject` on the definition read and summary; new `fetch_tag_keys`; `into_graph` carries subject + tags, so **export/import round-trips** them.

## Not included (follow-up)

- **Frontend display** of subject/tags (the `insight-front` side) — will be a separate change on this branch or its own PR.

## Testing

- `cargo test -p analytics` → 416 passed (incl. new `rejects_malformed_subject_and_tags`, `accepts_absent_subject_and_empty_tags`, and round-trip assertions in the ignored live tests), 0 failed.
- `cargo clippy -p analytics --tests` clean.
- Regenerated `openapi.json` (drift-checked) and the stand `analytics.py` schema.

## Summary by CodeRabbit

* **New Features**
  * Custom metrics can now include an optional subject and descriptive tags.
  * Tags support up to 64 unique values and require valid lowercase snake-case formatting.
  * Subject and tag metadata is preserved when metrics are saved, retrieved, listed, and exported.
  * Metric summaries now display the associated subject when available.

### DIFF
--- a/tests/stand/api/schemas/analytics.py
+++ b/tests/stand/api/schemas/analytics.py
@@ -617,6 +617,7 @@ class CustomMetricSummary(BaseModel):
     entity_type: str
     label: str
     metric_key: str
+    subject: str | None = Field(None, description='Grouping subject, so the management list can partition custom metrics\nby topic like the definitions listing; absent when none is declared.')
 
 
 class MetricDefinitionView(BaseModel):
@@ -779,6 +780,8 @@ class CustomMetric(BaseModel):
     scale: float | None = None
     short_label: str | None = None
     source_key: str
+    subject: str | None = Field(None, description='The single topic this metric groups under within its family; a\nlowercase snake-case slug. Optional for custom metrics.')
+    tags: list[str] | None = Field(None, description='Cross-cutting filter labels; lowercase snake-case slugs, unique per\nmetric. Optional — defaults to empty.')
     transform: ValueTransform | None = None
     unit: str | None = None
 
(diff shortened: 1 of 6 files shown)


tests/stand/api/schemas/analytics.py
```diff
@@ -779,6 +780,8 @@ class CustomMetric(BaseModel):
     scale: float | None = None
     short_label: str | None = None
     source_key: str
+    subject: str | None = Field(None, description='The single topic this metric groups under within its family; a\nlowercase snake-case slug. Optional for custom metrics.')
```
ktursunov, line 783: Can we add or extend test for this endpoints with new fields?

cyberantonz: Sure.

ktursunov (approved):

PR #2404, opened 2026-08-10
ci: deploy + seed a test stand after the umbrella chart publish on main

### LINKED ISSUE #2244: ci: deploy + seed a test stand after the umbrella chart publish on main

### LINKED ISSUE #2447: All person metrics return null on a freshly seeded install
## Summary
On a freshly installed and seeded instance, `POST /api/analytics/v1/metric-results` returns 200 with `value: null` for every metric key and every seeded person over the seeded window.

## Steps to Reproduce
1. Fresh install (umbrella 0.5.126, seeder at current `main`), then seed with `seed-stand.sh --days 365` — all steps complete; 26 personas, 10 fixtures in the manifest.
2. Run `insight-identity-resolution-sync` (completes and reports rows synced), then rebuild gold with `seed-stand.sh --step gold` (dbt PASS=19).
3. Query any definition listed by `/v1/metric-definitions` (~60 total) for any seeded person over the seeded window.

**Expected:** metric values computed from the seeded data.
**Actual:** 200 with a non-empty values array in which every `value` is null.

## Additional information
- Verified exhaustively 2026-08-12: all ~60 metric definitions × every seeded persona — every value null.
- Login and `/auth/me` resolve every persona correctly; the identity API returns 200s, and each metric-results response carries the persona's canonical UUID as `entity_id`.
- Seeded bronze and silver rows exist; the gold rebuild completes without errors.
- Reconciliation-style tests pass against this state because both serving relations they compare are empty (122/123 analytics API tests green). The one value-level assertion, `tests/stand/api/analytics/test_results.py::test_metric_results_200`, is marked non-strict `xfail` referencing this issue (PR #2404).
- The compose e2e lane has shown the same symptom since 2026-08-07: its readiness gate times out waiting for gold observations.
- Seed manifest and full probe output available on request (they carry persona data — not attached to this public issue).

ktursunov:
Closes #2244.

Merge-to-`main` validation stops at "images built, chart published". This adds the next step: install that exact chart on the shared test stand, seed it, and prove a person can sign in and see data.

## What this is

An **after-merge alarm, not a gate**, and it cannot be anything else: the artefact under test does not exist until `publish-chart` has run, and that runs on push to `main`. Merge blocking stays with the merge-queue lanes; this guards the claim that what those lanes approved actually installs and works. A red run means `main` is broken — the merge author owns it, fixes forward if the cause is small, reverts if not.

## Shape

`build-images.yml` gains an `outputs:` block on `publish-chart` and one calling job handing `chart_version` to a reusable `deploy-test-stand.yml`, `main` only. Named stages with their own timeouts, so a red run says which one died without opening a log:

| stage | what it does |
| --- | --- |
| deploy | `make -C deploy/gitops deploy` (the same target an operator runs) at that version, `--wait`, **no `--atomic`**; a downgrade guard refuses installing an older chart than what is deployed |
| edge (black-box) | unauthenticated DNS/TLS/route/auth probe against the public URL, before the seed spends its budget on an unreachable stand |
| seed | `src/ingestion/tools/seed/seed-stand.sh`, 365-day window (the analytics API refuses ≥400 days); an `all`/`silver` run then forces the identity projection (persons-seed + persons-sync) and a gold rebuild, so the stand serves resolved metrics rather than a null for every person |
| smoke | `pytest tests/stand -m stand_smoke` — the deployed-stand suite's own login/`/auth/me`/metric checks, not a separate package |

Smoke never runs after a failed seed. Deploys coalesce, so a running upgrade is never killed mid-flight. Chart version is resolved from the OCI registry's own latest tag (or `publish-chart`'s output on the `workflow_call` path), falling back to `.insight-version` on the default branch via the GitHub API only if the registry is unreachable — never from a local checkout, which is as old as its branch.

## What has actually been proven

Not "implemented" — run, against the real stand:

- **A full rebuild works from genuinely empty infrastructure**, through `deploy/gitops/scripts/recreate-test-stand.sh`: drops the application databases, reinstalls through `make deploy`, verifies.
- **Seed works without `--force`** on a freshly wiped tenant.
- **Login and per-persona identity resolution work** — `/auth/me` and the session-identity checks pass, now consolidated into `tests/stand/api` under the `stand_smoke` marker instead of a parallel `tests/stand/smoke` package.
- **Person metrics resolve to a real number** — `POST /v1/metric-results` for a seeded person's canonical UUID returns a non-null value, proven by `test_metric_results_200` under `stand_smoke`. This closes constructorfabric/insight#2447: the seed had left the stand with the persons-seed CronJob disabled, so connector accounts were never linked to the seeded person and gold served null for every person metric. The seed now forces the projection (persons-seed links each connector account to the seeded person by e-mail, so the UUID a login resolves to is the UUID gold attributes activity to) and rebuilds gold over it.
- **No restart step**: the unconditional `kubectl rollout restart` of three hardcoded deployments is gone. This stand's config Secrets are static by policy; a deliberate config-values change requires a manual rollout restart of the three readers — documented as an accepted gap in `environments/test-stand/INFRA.md` §25.

## Script and workflow inventory (post-restructure)

- `deploy/gitops/scripts/recreate-test-stand.sh` — every `make deploy` precondition (version resolution, sync-clean, kube-ctx, chart availability) now runs before the first destructive action, not between the wipe and the reinstall.
- `deploy/gitops/scripts/lib.sh` (new) — the shared guard/helper library (`cluster_guard`, `namespace_assert`, `oidc_secret_nonempty`, colour/`hdr`/`note`/`die`) `recreate-test-stand.sh` sources, replacing four-plus divergent inline copies.
- `deploy/gitops/scripts/provision-ci-deployer.sh` — now a 50-line CLI wrapper around three new Makefile targets (`provision-ci`, `verify-ci-credential`, `revoke-ci`); the RBAC objects it used to render from a heredoc are a committed, static manifest at `deploy/gitops/environments/test-stand/ci-deployer-rbac.yaml`.
- `.github/workflows/scripts/extract-seed-manifest.py` — deleted. The seeder emits one `SEED_MANIFEST_JSON:` sentinel line (`insight_seed/manifest.py`), captured with `grep`/`cut` instead of a parser.
- `.github/workflows/scripts/redact-stand-log.py` — the long-blob rule now also masks base64url (Keycloak client secrets, PKCE verifiers), and the final verification pass re-checks it.

## Design notes for a reviewer

**`make deploy` is the deploy path.** CI runs the official target, not a second helm procedure. Deploy shape is declared in `environments/test-stand/inventory.yaml` and `values.yaml`.

**No `--atomic`, and the stand is disposable.** A rollback reinstalls the *previous* chart onto a stand whose job is to show what the newest one does. Recovery is fix-forward or `recreate-test-stand.sh`, never a downgrade — a stuck-release guard (Makefile) and a downgrade guard (workflow) both hold this; see `INFRA.md`, "Incidents that shaped this environment", for the failure modes these close.

**No cluster diagnostics are published.** A red run prints which stage died and the black-box edge result; helm history and pod logs are read operator-side. `redact-stand-log.py` filters everything else that reaches the console.

**Least-privilege credential.** `make provision-ci` applies a committed, namespace-scoped RBAC manifest and asserts its scope with a `kubectl auth can-i` matrix (`make verify-ci-credential`) before any secret is handed to CI.

**Incident log.** Six failure modes this environment's guards close are documented once, in `deploy/gitops/environments/test-stand/INFRA.md` ("Incidents that shaped this environment"), rather than as narrative comments scattered across the workflow and scripts.

## Known follow-up (not this PR)

- `seed-stand.sh` still discovers the dev-lead's realm-config-map name by sed-scraping `insight_seed/profiles.py`'s `DEV_LEAD_UUID` constant and, when `keycloakConfig.realmsConfigMap` is unset, guessing the chart's Helm-fullname convention. Both are read-only fallbacks, not a correctness bug, but they couple this script to another module's internals. Follow-up: declare the realm ConfigMap's name (or the dev-lead's address directly) as a `values.yaml` key so the realm and the seeder read one committed value instead of one script inferring the other's naming convention.
- A config-Secret content change does not roll the three `envFrom` readers (env vars are read once at container start, and the pod spec doesn't change). Accepted for now: this stand's config values are pinned and rotations are rare manual events; the manual-restart procedure is in `environments/test-stand/INFRA.md` §25. Permanent fix when it matters: hash the Secret's data into each subchart's `checksum/config` from the same values it is templated from.

## Setup — confirmed by the deploy → seed → smoke runs

- [x] The `insight-test-stand` environment exists with its secrets set (`TEST_STAND_KUBECONFIG`, `TEST_STAND_PERSONA_PASSWORD`) — the runs authenticate with the namespace-scoped `ci-deployer` credential and log the seeded personas in. Runbook: `deploy/gitops/environments/test-stand/credentials-runbook.md`
- [x] `make -C deploy/gitops provision-ci ENV=test-stand` (or the `provision-ci-deployer.sh` wrapper) minted the credential; it self-verifies with `verify-ci-credential`.
- [x] Persona password set — a constant in this public repo applied to every seeded user on an internet-reachable stand.
- [x] **Admin, before merge:** lock the environment's deployment-branch policy to `main` (it currently also allows this branch, which was needed for pre-merge validation).
- [ ] A known flake to watch, not to paper over: an occasional single-run timeout on one drilldown case (`#2463`, merged to `main`, closed the related drilldown coverage gaps). This workflow is an after-merge alarm, so a rare timeout will eventually page somebody about a healthy stand.

## Summary by CodeRabbit

* **New Features**
  * Added automated and manually triggered test-stand deployment with configurable deployment and seeding stages.
  * Added environment configuration for routing, authentication, data services, and seeded test data.
  * Added guarded test-stand recreation and CI deployment credential management.
  * Added configurable persona passwords and gold-data rebuilding.
* **Security**
  * Strengthened credential handling, access controls, log redaction, and authenticated endpoint validation.
* **Documentation**
  * Added deployment, recovery, infrastructure, and credential-management guidance.
* **Bug Fixes**
  * Improved anonymous access checks and seed discovery and execution reliability.


.github/workflows/deploy-test-stand.yml
```diff
@@ -0,0 +1,1058 @@
...
+          # holding the reason are still there for the diagnostics step. Recovery
+          # is the next merge, or a human running `make rollback ENV=test-stand`.
+          #
+          # Deliberately NO --create-namespace: the namespace exists, and the CI
+          # credential is namespace-scoped and could not create one anyway —
+          # asking would turn a working deploy into an RBAC failure.
+          #
+          # --timeout 10m rather than the gitops Makefile's 30m default: a deploy
+          # that is going to fail should say so inside the CI budget.
+          helm upgrade --install "$STAND_RELEASE" "$CHART_REF" \
```
ktursunov: Please keep CI on the official GitOps deployment path rather than spelling out a second Helm procedure here. The one-time credential bootstrap is separate, but application deploys should call the same supported entry point operators use. If make deploy cannot represent this stand, adjust the official target or inputs so it can; otherwise this workflow and the documented deployment procedure will drift.

.github/workflows/deploy-test-stand.yml
```diff
@@ -0,0 +1,1058 @@
...
+        required: false
+        default: all
+        type: string
+  # The human path: redeploy a specific version by hand, e.g. after fixing the
+  # stand by hand and wanting CI's view of it back. Same input, same code path —
+  # there is no manual-only branch anywhere below.
+  workflow_dispatch:
+    inputs:
+      chart_version:
+        description: "Umbrella chart version to install (e.g. 0.5.101). Leave empty to take the newest one published from main."
```
ktursunov: Please remove support for selecting or deploying arbitrary older chart versions. On main, deploy the exact already-built version emitted by publish-chart. This test stand is disposable: recovery from a broken release is to wipe and recreate the stand, then install the current published chart, not downgrade or roll back in place. Please also include or point to an executable, idempotent infrastructure create/recreate script; INFRA.md alone is not enough to prove we can rebuild on the same or a replacement cluster.

.github/workflows/deploy-test-stand.yml
```diff
@@ -0,0 +1,1058 @@
...
+          # resolve a variable to learn which tests gate a deployment.
+          uv run --project tests --frozen \
+            pytest tests/stand/smoke -ra --stand-manifest "$INSIGHT_STAND_MANIFEST" \
+            2>&1 | python3 "$REDACT"
+
+      # ── Diagnostics ───────────────────────────────────────────────────────
+      # One step, only on failure, and everything it can print is decided by a
+      # committed script rather than by this file — so widening what a red run
+      # publishes is a reviewable diff and not an edit to a `run:` block.
+      - name: Failure diagnostics (curated, redacted)
```
ktursunov: Please remove the public cluster-diagnostics step and script. After deploy, run a shallow black-box smoke first—DNS, TLS, the public route, and a cheap auth endpoint—before seeding and the full stand checks. If that fails, keep deeper Helm and Kubernetes diagnostics operator-side instead of publishing cluster output in CI.

deploy/gitops/scripts/provision-ci-deployer.sh
```diff
@@ -0,0 +1,1027 @@
...
+#
+# The \`secrets\` rule below looks like a widening and is the opposite: it is
+# restricted by \`resourceNames\` to the exact Secret the chart's own Role names,
+# and it exists only because RBAC escalation prevention forbids creating a Role
+# that grants a permission the creator does not hold. Removing it does not make
+# the credential smaller; it makes the deploy fail.
+#
+# Nothing else in this namespace is reachable: no pods, no other Secrets, no
+# workloads, no exec.
+apiVersion: rbac.authorization.k8s.io/v1
```
ktursunov: The insight-airbyte-auth-reader Role and RoleBinding are chart-owned and should continue to follow the chart lifecycle. The extra CI-deployer authorization to mutate the Airbyte namespace is infrastructure/bootstrap policy and should not be installed by this application-repository helper. Please provision and use the official GitOps deployment identity/path rather than changing deployment permissions here.

.github/workflows/deploy-test-stand.yml
```diff
@@ -0,0 +1,1058 @@
...
+        run: |
+          set -euo pipefail
+          {
+            echo "### Test stand — \`$CHART_VERSION\`"
+            echo ""
+            echo "| | |"
+            echo "|---|---|"
+            echo "| chart | \`insight-$CHART_VERSION\` |"
+            echo "| environment | \`$ENV_NAME\` (namespace \`$STAND_NAMESPACE\`, release \`$STAND_RELEASE\`) |"
+            echo "| stages | \`${STAGES:-all}\` — deploy ✓, seed $([ "$RUN_SEED" = "true" ] && echo '✓' || echo '— skipped'), smoke $([ "$RUN_SMOKE" = "true" ] && echo '✓' || echo '— skipped') |"
```
ktursunov: These checkmarks represent requested stages, not completed ones. A deploy failure currently prints seed ✓ and smoke ✓ even though both were skipped. Please derive each status from actual step outcomes, or label this row as requested and avoid success marks.

ktursunov (commented):
Follow-up review focused on simplifying the disposable test-stand flow and keeping it on the official deployment path.

.github/workflows/deploy-test-stand.yml
ktursunov: Resolved in 3e7662b9: the workflow's only deploy invocation is `make -C deploy/gitops deploy` — the same target an operator runs — with the stand's shape declared in `environments/test-stand/inventory.yaml`/`values.yaml`. The workflow's duplicated preflight guards (context, reachability, chart-present, OIDC secret) were folded into the make targets too, so there is no second procedure left to drift.

.github/workflows/deploy-test-stand.yml
ktursunov: Resolved in 3e7662b9 + 9cf6e3ec: a `workflow_call` installs exactly the version publish-chart just published (required output, never `.insight-version`); a manual dispatch takes the registry's newest published version by default; and the downgrade guard now refuses an older chart **unconditionally** — the override input was removed. Going backwards means wiping and recreating, for which the executable, idempotent rebuild now exists: `deploy/gitops/scripts/recreate-test-stand.sh`, which resolves every precondition (version, tree state, context, chart, protected secrets) before its first destructive step.

.github/workflows/deploy-test-stand.yml
ktursunov: Resolved in 3e7662b9: a shallow unauthenticated edge probe (route status for every parent, `/`, `/auth/me`, OIDC discovery issuer) runs immediately after deploy and before the seed spends its budget; the public cluster-diagnostics output is gone — pod logs no longer stream to the job log (summary + allowlisted failure diagnostics only, full logs kept runner-side), and deeper helm/kubectl diagnosis is documented operator-side in `environments/test-stand/INFRA.md`.

deploy/gitops/scripts/provision-ci-deployer.sh
ktursunov: Resolved in 3e7662b9: the provisioning helper no longer touches the Airbyte namespace at all. The committed `environments/test-stand/ci-deployer-rbac.yaml` installs only app-namespace objects (ServiceAccount, namespace-scoped `admin` RoleBinding, a CRD-supplement Role/Binding, token Secret); `make verify-ci-credential` merely *asserts* the externally provisioned identity holds the expected grants — it installs nothing. The chart-owned Role/RoleBinding stay on the chart lifecycle.

.github/workflows/deploy-test-stand.yml
ktursunov: Resolved in 3e7662b9: every summary row is derived from its step's actual outcome (`steps.<id>.outcome || 'skipped'`), rendered worst-first (failure > cancelled > skipped > success). A failed deploy renders seed and smoke as `skipped`, never as ✓.

deploy/gitops/environments/test-stand/ci-deployer-rbac.yaml
```diff
@@ -0,0 +1,103 @@
...
+# RoleBinding, not ClusterRoleBinding: roleRef names cluster-scoped `admin`, but the
+# binding itself is namespaced, so the grant stops at this namespace's edge.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: ci-deployer-admin
+  namespace: insight
+  labels:
+    app.kubernetes.io/name: ci-deployer
+    app.kubernetes.io/component: ci-credential
```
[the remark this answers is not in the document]
ktursunov: By design, with containment asserted rather than assumed: this binding grants the built-in `admin` ClusterRole via a namespace-scoped RoleBinding only — the credential is the CI deployer for this one namespace (helm release storage, the seed Job, pod logs), and `make verify-ci-credential` proves the negative space with cluster-scoped `kubectl auth can-i` checks that must answer no. The token is revocable in one command (`make revoke-ci`). A narrower hand-rolled Role would need most of the admin verb surface anyway (helm owns arbitrary namespaced kinds) while adding a second RBAC definition to maintain. Suggest dismissing the alert as won't-fix / by-design with this rationale.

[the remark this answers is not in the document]
ktursunov: Superseded — fixed for real in 04d8bac8d rather than argued: the `admin` ClusterRole binding is gone, replaced by an enumerated `ci-deployer-core` Role listing exactly the kinds the umbrella chart manages in this namespace plus read-only diagnostics. A new chart kind fails the deploy closed with a helm forbidden naming the missing resource. Verified live: full `make verify-ci-credential` matrix green and a real `make deploy` through the credential after the admin binding was deleted from the cluster. KSV-0111 no longer matches by construction.

deploy/gitops/scripts/provision-ci-deployer.sh
ktursunov: Update to the resolution above: the manifest no longer contains an `admin` RoleBinding at all — since 04d8bac8d the grants are an enumerated `ci-deployer-core` Role (chart-managed kinds + read-only diagnostics), alongside the unchanged CRD supplement. The Airbyte statement stands: nothing in this repo installs grants in that namespace; `verify-ci-credential` only asserts the externally provisioned ones exist.

mozhaev-dev (approved):

PR #2420, opened 2026-08-11
test(stand): assert metric-definitions listing exposes subject and tags

cyberantonz:
## Summary

Follow-up to #2397 (merged), which added `subject` and `tags` to `GET /v1/metric-definitions`. That PR shipped Rust unit coverage (registry invariants + row→view mapping) and regenerated the stand schema, but no **deployed-stand** test asserted the fields are actually *populated* end-to-end. This adds it.

`tests/stand/api/analytics/test_results.py::test_metric_definitions_expose_subject_and_tags`:

- Every **builtin** definition carries a non-null `subject` that is a `^[a-z][a-z0-9_]*$` slug ≤64 chars (mirrors the `VARCHAR(64)` column and the registry invariant) — so a dropped subject on the reconcile write or the listing `SELECT` fails here.
- `tags` on every definition are shaped, ≤64, and unique per metric.
- At least one builtin carries a non-empty `tags` list — so an empty `metric_definition_tags` join (which the `extra="forbid"` schema alone would not catch, since `tags` may legitimately be empty) fails the contract.

Also exports `MetricOrigin` from `tests/stand/api/schemas` so the test can scope the subject assertion to builtins.

Refs #2344.

## Testing

- `ruff check` clean; `pytest --collect-only` discovers the case. Runs against a seeded stand in CI like the rest of `tests/stand/api/`.


cyberantonz:
Folded into #2401 (rebased onto main, test commit cherry-picked on top of the feature commit) — the stand assertions depend on the schema fields #2401 introduces, so they ship together.

Issue #1607, opened 2026-07-03
ANALYZE 3: Development cost analytics (AI)


mitasovr:
## Goal

Quantify the cost of AI-assisted development and attribute it to person / team / role / period. Layer 2 of #1461's three-layer model (*where AI was used → what it cost → what changed*); the cost input #1609 needs. **Bounded by the attribution ceiling in #1660: AI cost is honest at person × tool × period (L2); there is no honest per-PR cost (L7).**

## Where cost comes from (verified in ingestion code + live cluster, 2026-07-06)

- **Claude Team** — `total_cost` / `avg_cost_per_day` (per-seat overage), person×day. *[present in data]*
- **Claude Admin** — `claude_admin_cost_report` (token-metered). *[code path exists; empty on cluster]*
- **OpenAI** — `bronze_openai.costs` → `to_ai_cost.sql`, grain **`project_id × line_item × day` (not per-person)**. *[code path; empty locally]*
- **Cursor** — per-event `chargedCents`. *[0 rows locally]*
- **Claude Team seat overage** — `silver.class_ai_overage` (person × month): `credit_limit_cents`, `used_amount_cents`, `overage_cents`, `is_over_limit`, `seat_tier`. Already has a gold view + catalog metric (`m20260618`). *[293 rows on cluster]*
- Implemented gold view: `ai_cost_person_period` = `sum(cost_cents) GROUP BY person_id, tool`.

## Core principle

- **Two billing models, never mixed** — token-metered (Claude Admin, OpenAI) vs per-seat (Claude Team overage, Cursor). No blended figure.
- **Cost is honest only at person / team / role × period (L2).**
- **Per-seat economics are a first-class cost dimension, not just a total.** For per-seat plans (Claude Team), separate: (a) **base seat cost**, (b) **overage** beyond the included monthly credit limit, and (c) **seat utilization** = `used_amount / credit_limit`. Under-utilized seats (paying for unused budget) and over-limit seats (paying overage) are both cost-control signals. This keeps the methodology valid for both billing models a tenant may run (per-seat fixed + overage, and token-metered).
- **No `$/PR` as fact (L7).** Per-PR cost may appear ONLY as, and explicitly labelled:
  - **allocated** — a person's period AI-cost spread across their PRs (spreads all spend over all PRs, AI or not); or
  - **cohort-ratio** — `Σ cohort AI-cost / Σ cohort PRs` (group efficiency, not per-PR truth).
[...]

PR #2432, opened 2026-08-11
feat(ai-cost): ingest Claude Team vendor invoices

### LINKED ISSUE #2429: AI cost: Claude Team vendor invoices
Sub-task of #1607. Depends on the seat-economics sub-task for its specifications.

Delivers the invoiced layer for Claude Team: what Anthropic actually billed, and the only source of a seat's price.

## Why this is its own connector

The invoice list is declarative-shaped. Its line items are not: the claude.ai wrapper carries no invoice id and no lines, only a `hosted_invoice_url`, and reaching the lines takes a three-hop chain across two hosts where each hop's credential comes out of the previous hop's response body. Same reasoning as github-copilot ADR-0001. The existing claude-team connector stays declarative and untouched.

## Scope

- `claude-team-invoices` — a Python CDK source with one stream, `claude_team_invoice_lines`.
- `silver.class_ai_invoice` and its first contributor, at invoice-line grain.
- A data-quality check on invoices whose enrichment did not complete.

## What the measurement established

Probed against live data before any code was written:

- The proxy already forwards `/api/stripe/*` through a wildcard route, and the installed session key answers **200** on the invoice list — no proxy change, no new credential.
- Both Stripe hosts are reachable from the ingestion namespace.
- The seat price is the line's `hosted_invoice_unit_amount`. It is **not** the wrapper's `num_seats`, which is populated on 9 of 108 invoices and there names one line's quantity while the invoice covers several tiers. It is **not** `amount / quantity` either: a mid-period seat change emits proration lines whose amounts cover part of a period.
- One tenant runs several tiers at once — Standard at $25.00 and Premium at $125.00 on one invoice — so a seat price binds to a tier, not to an organisation.

## Known risk

The bootstrap host is undocumented: it appears in no Stripe documentation and no public discussion, carries no version and no contract. Stripe expires hosted invoice URLs 30 days after the due date and never later than 120 days; what keeps history reachable is that the wrapper re-issues a fresh URL on every list call, so the URL is followed inside the run that fetched it and never stored.

There is no supported alternative — the official Invoice API authenticates as the merchant, and for these invoices the merchant is Anthropic. The exposure is accepted and covered by the degradation rules rather than avoided.

Gregory91G:
Closes #2429. Sub-task of #1607. **Stacked on #2431** — review that one first; this branch contains its commit.

Ingests the invoiced layer for Claude Team: what Anthropic actually billed, and the only source of a seat's price.

## Why a CDK connector

The invoice list is declarative-shaped. Its line items are not: the claude.ai wrapper carries no invoice id and no lines, only a `hosted_invoice_url`, and the lines sit behind a three-hop chain across two hosts in which each hop's credential comes out of the previous hop's response body. Same reasoning as github-copilot ADR-0001. The existing declarative `claude-team` connector is untouched.

```
GET {proxy}/api/stripe/{org}/invoices            -> wrapper rows, hosted_invoice_url
parse https://invoice.stripe.com/i/{acct}/{tok}  -> (acct, token)
GET invoicedata.stripe.com/hosted_invoice_page/… -> invoice_id + ephemeral key
GET api.stripe.com/v1/invoices/{id}/lines        -> line items
```

## What lands

- **`claude-team-invoices`** — one stream, `claude_team_invoice_lines`: one record per invoice carrying that invoice's money, plus one per line. Its own bronze namespace, because dbt rejects two `schema.yml` files declaring one source.
- **`silver.class_ai_invoice`** and its first contributor, at invoice and line grain — aggregating earlier would make the per-tier seat price unrecoverable, and keeping the invoice total off the lines means summing it needs no dedup.
- **Data quality** — `assert_ai_invoice_lines_enriched`, bounded to the last three billing months because an invoice that never enriched does not self-heal, and excluding drafts, which legitimately carry no lines yet.

## The rules, and why they are what they are

**The seat price is the line's `hosted_invoice_unit_amount`.** Not the wrapper's `num_seats`, which may be absent and, where present, names one line's quantity while an invoice can price several tiers. Not `amount / quantity` either: a mid-period seat change emits proration lines whose amounts cover part of a period.

**A tier is part of the grain.** An invoice can price several tiers on separate lines in one period, so a price binds to a tier, not to an organisation.

**Every invoice carries its own row.** That row holds the invoice's money and how far its chain got; lines are added beside it only when the chain completed, and they carry only their own money. So an invoice's money sits on exactly one row, and a sync that enriches an invoice replaces the row an earlier unenriched sync wrote instead of adding a second one beside it. `chain_status` tells the four outcomes apart — `ok`, `failed`, `unparsable_url`, and `no_hosted_url` for an invoice the vendor offered no URL for, as a draft legitimately does. Only URLs that were offered count towards drift: more than half of them failing to parse fails the run instead, because a set of unpriced rows would read as the vendor having stopped charging for seats.

**The ephemeral key never lands.** Not in a record, a state message or a log line — which is why the chain lives inside one stream rather than a parent/child pair, whose parent records would persist it.

**Hosted URLs expire.** Stripe expires them 30 days after the due date and never later than 120. History stays reachable only because the wrapper re-issues a fresh URL on every list call, so a URL is followed inside the run that fetched it and is never stored.

## Known risk

The bootstrap host is undocumented — no Stripe documentation, no public discussion, no version, no contract. There is no supported alternative: the official Invoice API authenticates as the *merchant*, and here the merchant is Anthropic. The exposure is accepted and covered by the degradation rules rather than avoided.

## Test plan

Automated, run:

- [x] 62 connector unit tests — the chain's parsing and classification, and every degradation rule end to end against a stubbed fetcher, including an assertion that no emitted record carries the ephemeral key.
- [x] `dbt parse` — clean; `tag:claude-team-invoices+` selects the promotion model, the staging model, the class and its generated tests.
- [x] `metrics/test_ai_invoice_silver.py` — 11 tests through bronze → dbt → ClickHouse, including a recovery reached inside one build and one reached across two, which is the case an append-only staging model gets wrong.
- [x] seed-tool suite — 45 tests, including the invariant that every registered reset target is actually cleared by a generator.
- [x] `scripts/ci/connector_wiring.py` — passes. It caught a missing `bootstrap-db` entry during development, without which the bronze database is never created and the dbt models fail with `UNKNOWN_DATABASE`.

Chain verified end to end against the vendor before any code was written:

- [x] The proxy forwards `/api/stripe/*` through a wildcard route and answers 200 with an installed session key — no proxy change, no new credential.
- [x] Every hop of the chain returned 200 across a set of invoices spanning several billing months, and `Stripe-Version: 2026-06-24.dahlia` was accepted.
- [x] Both Stripe hosts are reachable from the namespace the connector runs in; where egress is governed by a `NetworkPolicy` it has to admit them.

Manual, needs a stand — please tick these yourself:

- [ ] Merge so `build-images.yml` builds the image and `bump-descriptors` patches `images.cdk.image` (empty here, as expected for a new connector).
- [ ] Create the Secret from `src/ingestion/secrets/connectors/claude-team-invoices.yaml.example`. Its three values are the same ones the `claude-team` connector already uses; no new credential is needed.
- [ ] After the reconcile loop picks it up, confirm `bronze_claude_team_invoices.claude_team_invoice_lines` holds rows with a non-empty `seat_unit_amount`, and that `silver.class_ai_invoice` carries one row per line.
- [ ] `e2e-bronze-to-api` covering `metrics/test_ai_invoice_silver.py`.

## Summary by CodeRabbit

- **New Features**
  - Added Claude Team invoice ingestion for subscription, proration, seat, and extra-usage line items.
  - Added unified AI invoice reporting with billing periods, currencies, amounts, categories, and enrichment status.
  - Added scheduled synchronization and configuration for billing and proxy credentials.

- **Bug Fixes**
  - Improved recovery from failed invoice retrievals while preserving valid data and preventing duplicate records.

- **Documentation**
  - Added setup, credential, network, pricing, and troubleshooting guidance.

- **Tests**
  - Added end-to-end coverage for invoice transformations and billing scenarios.


cyberantonz (dismissed):

hello1101n (dismissed):

mozhaev-dev (dismissed):

hello1101n (approved):

mitasovr:
@Gregory91G Review of this PR (8 independent finder angles, each candidate adversarially verified against the sources). Overall: the task is solved well at its core — the chain/degradation design, the seat-price rules, the grain, and the ephemeral-key handling all match the issue's specification, and the I/O-shell vs pure-logic split in `stripe_chain.py` is clean. But two findings are blockers and three more are worth fixing in this PR.

### Blockers

**1. A stale `failed` gap row survives invoice recovery — the double-count the code says it prevents.** `claude_team__ai_invoice.sql` — the `superseded` CTE only drops the gap row *within one build*. The model is `incremental`/`append` with a strict `_airbyte_extracted_at > max(collected_at)` watermark, so on the run where an invoice recovers, the `failed` row appended by the previous run is neither re-selected (watermark) nor deleted (append strategy; and its wrapper-derived `unique_key` differs from the `ok` rows' Stripe-id keys, so neither RMT collapse nor silver's `delete+insert` removes it). Result: `silver.class_ai_invoice` permanently holds both the gap row and the lines, `invoice_net_cents` is double-counted at invoice level, and `assert_ai_invoice_lines_enriched` stays red after recovery — exactly the two consequences the CTE comment claims to prevent.

**1a. The recovery e2e test passes vacuously and cannot catch this.** In `test_ai_invoice_silver.py`, the `failed` row in `RECOVERED_ROWS` carries the same `_airbyte_extracted_at` as every row of the earlier fixture, `DbtRunner.build()` never passes `--full-refresh`, and nothing truncates `staging.claude_team__ai_invoice` between fixtures (it is also absent from the conftest session-start truncate list, unlike its two `ai` siblings). The strict `>` watermark excludes the gap row before `superseded` ever runs, so `assert [...] == ["ok"]` holds even with the `superseded` JOIN deleted. Fix: advance the `RECOVERED_ROWS` timestamps past the first fixture's watermark and truncate staging between fixtures — then the test reproduces finding 1.

**2. Production-derived information in committed files** (root `AGENTS.md`, "Never expose production-derived information"). Observed-data claims in `README.md` (~lines 46–53: the `num_seats` frequency claim and the multi-tier-tenant claim) and in `stripe_chain.py` (~lines 188–190, the same frequency claim as a code comment) should be rewritten as capability statements ("`num_seats` may be absent; when present it names one line's quantity"). The fixture in `test_ai_invoice_silver.py` (~lines 89–127) uses real vendor tier names with specific realistic prices, seat counts, and totals — every other fixture in this PR is correctly synthetic ("Example plan"); this one should be re-cut to match. The repo is public, so this content is published and survives in forks.

### Should fix in this PR

**3. `descriptor.yaml` is the only descriptor (of 21) without `secret.required_fields`.** `validate_secret.py` returns `required=[]` for an absent block, so any Secret validates and the reconciler's "required field missing — skipping" branch is unreachable; a misconfigured Secret gets scheduled and the pod fails at CDK config validation on every run instead. The sibling `ai/claude-team/descriptor.yaml` lists exactly the three keys this connector also needs.

**4. Re-seeding a stand leaves phantom invoice months.** `seed_claude_team_invoices_bronze` truncates only bronze; `staging.claude_team__ai_invoice` and `silver.class_ai_invoice` are in no `RESET_TARGETS` and are never truncated or full-refreshed. With a pinned `SEED_ANCHOR_DATE`, a re-seed with a shorter window inserts zero staging rows (identical timestamps vs a strict `>` watermark) while the months that dropped out of the window persist in staging and silver — the seed is not idempotent, and the placeholder-drop hook can't help (the table is no longer marked as a placeholder after the first build).

**5. The drift guard counts a legitimately absent `hosted_invoice_url` as "unparsable".** `parse_hosted_invoice_url(None)` returns `None` and joins the drift count, so a small invoice set where one or two invoices carry a null URL (draft/void/zero-amount) aborts the entire sync with a misleading "format changed" error and zero rows written. Consider a separate `chain_status` for URL-less invoices and counting only non-empty, non-matching URLs toward drift.

### Worth a follow-up

**6. Silent unpricing on a type change.** `seat_unit_amount` returns `None` for any non-`int` (`isinstance(amount, int)` after the bool guard). If the hosted-invoice surface ever serialises the unit amount as a float or digit string, every seat price in the run becomes NULL while `chain_status` stays `ok` — invisible to the coverage test and indistinguishable from "the vendor prices no seats". Accepting int-valued floats/digit strings, or degrading the row to `failed` on an unrecognised type, would make it visible.

**7. Stateless daily full re-walk.** `read_records` ignores state, and the schedule is daily, so the two-hop chain re-runs for every invoice ever issued although finalised invoices' lines are immutable and already in bronze. URL expiry doesn't force this: the wrapper listing can still be walked in full while the chain runs only for invoices newer than a high-water mark plus previously `failed`/`unparsable_url` ones. That also minimises traffic against the undocumented host this PR itself flags as the known risk.

mitasovr (approved):

mitasovr:
@Gregory91G Re-checked the three fix commits — all five in-PR findings are resolved, and the invoice-row redesign is the right depth: one row per invoice carrying its money under a chain-outcome-free wrapper key, staging on `delete+insert`, the `superseded` CTE gone, and the recovery path now covered across two real builds. Nice work. From my side this is good to merge.

One residual edge worth a note (non-blocking, follow-up material): the invoice row's key includes `invoice_payment_intent` and `invoice_total`, so it is only stable while the wrapper reports those fields identically on every run. The draft lifecycle is the case where that likely breaks: a draft (now emitted as `no_hosted_url`) plausibly has no payment intent yet and a total that can still change; once the invoice is finalised and paid, the intent appears and the key changes — the old draft row then stays in the class beside the new one, carrying its money. `assert_ai_invoice_lines_enriched` deliberately doesn't report drafts, so nothing would surface it, but a sum over `invoice_net_cents` would count that invoice twice.

If drafts do come through the wrapper, the cheap guard is to keep a draft's money off its row (or skip drafts at emission entirely — a draft has no final money by definition), so a key change on finalisation can't duplicate ledger money. Fine as a follow-up alongside the `seat_unit_amount` type-strictness and the incremental-state one.

PR #2441, opened 2026-08-11
feat(frontend): read a chart day over its bar, and expand the zone rail on hover

dzarlax:
Two interaction changes on the person surface. Both came from watching someone use it rather than from reading the code.

## Reading a day out of a chart

The value appeared in the caption under the strip. That is a change in the middle of a chart, which is exactly where a reader looking at one bar is not looking: they hover, the number moves somewhere in their periphery, and they never learn it was there.

It now sits over the bar the pointer is on, and the caption stops changing at all — what is down there is constant again.

One positioned element rather than a tooltip per day: a month is thirty-one triggers, and thirty-one floating cards is both heavy and the wrong shape. It sits above the bars, with headroom reserved, so it covers neither the neighbours being compared against nor the header's own comparisons — which are exactly what a reader is hovering in order to reason against.

It anchors to its bar's edge and grows inwards rather than centring. Centring needs the rendered width in order to stay inside the strip, and it does not have it: the text runs from "3 messages" to "3.5 hours of 8", so any constant guarding the ends is a guess that the longer strings walk straight past. Growing inwards cannot overflow whatever the text says.

## The zone rail expands on hover

Ported from the lite product's rail. Four things make it work, and each is there because leaving it out broke something.

It opens on a timer, and that is the whole guard against opening over something a pointer was only passing. The first version delayed the fade in CSS instead and let the panel become clickable at once — which achieved the opposite, and is described under "what the first attempt got wrong" below.

The rail keeps its slot in the layout and the labels open OVER what is beside them: widening the element itself shoves the pane sideways every time a pointer crosses the rail on its way elsewhere. The buttons widen only while it is open — a label you can read but not click is a trap, because the pointer leaves the narrow column on its way to the word and the rail shuts before it arrives. People aim at what they can read. The buttons sit inside the hover target, so a pointer resting on one keeps it open rather than fighting what opened it. And a click collapses it until the pointer leaves: a click navigates, the pointer has not moved, and without this the rail reopens on top of the pane the click was aimed at.

Two things could not be copied over. The lite product keeps the post-click flag in `sessionStorage` because its click reloads the page; here navigation is client-side, so plain state carries it. And the open state is held in React rather than expressed as CSS variants — it is the product of two facts, one of which no selector can see. The variant version silently never matched and nothing said so, which is why the behaviour is now covered by tests rather than by eye.

## What the first attempt got wrong

The panel takes pointer events while open. With it inert, the gaps between buttons belong to whatever is underneath, so a pointer moving from an icon towards its label crossed bare panel, the rail counted that as having been left, and it slammed shut under the hand reaching for it.

Fixing that introduced the opposite fault, which review caught. `pointer-events` is not animatable, so it flipped the instant the state did while the opacity waited out its delay: for that window the panel was an invisible, fully clickable band over the pane. Clicks aimed at a row landed on it and vanished, and a pointer merely crossing the rail entered it, counted as still being inside, and got the rail opened over the row it was heading for — the exact thing the delay was written to prevent. Hence the timer: with it, open and visible are the same fact and no such window exists.

Three further ways the state could be stranded, all found in review: a keyboard activation set the "stay shut" flag that only a departing pointer could clear, so Enter left the rail dead to the mouse; a width change unmounted the rail under the pointer without a leave event, so it returned already open; and focus never opened it at all, leaving a sighted keyboard user with eight identical icons. Each is covered by a test now, and two of those tests fail on the commit before the fix.

It deliberately does not cover the pane beside it, and what it does not cover is faded rather than dimmed. Dimming was tried first and did nothing: the rows do not object to being darkened, they object to being CUT, and a hard edge through the middle of a word reads as a rendering fault whatever its brightness.

## Checks

`tsc -b` clean, eslint clean, 1081 unit tests pass, including eight covering the rail's open state. Verified in a browser: the click-then-still-hovering case, both ends of the chart strip measured against the strip and the card, and the labels measured against the box that used to clip them.

## Summary by CodeRabbit

* **New Features**
  * Lens navigation rail now expands on hover or keyboard focus, with smooth dismissal and responsive behavior.
  * Expanded navigation displays wider zone buttons and labels while preserving access to settings.
  * Day activity strips show hovered readings in a dedicated readout above each bar.
* **Bug Fixes**
  * Improved pointer and touch handling prevents accidental expansion during brief crossings.
  * Navigation state now resets reliably when the layout or rail is remounted.
  * Day captions remain consistent while inspecting individual readings.

### DIFF
--- a/src/frontend/src/components/portal/lens-rail.test.tsx
+++ b/src/frontend/src/components/portal/lens-rail.test.tsx
@@ -1,0 +1,111 @@
+// @vitest-environment jsdom
+/**
+ * The rail's open state.
+ *
+ * Every case here is about the interaction rather than the look, because the
+ * look is the easy half. The one that matters is the click: a click navigates
+ * and leaves the pointer sitting on the rail, so without an explicit dismissal
+ * the rail reopens on top of the pane the click was aimed at. That failed
+ * silently once already when the state was expressed as CSS variants — the
+ * rules simply never matched and nothing said so.
+ */
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const mocks = vi.hoisted(() => ({
+  layout: "desktop" as string,
+  selected: [] as string[],
+}));
+
+vi.mock("@/lib/portal/use-shell-layout", () => ({
+  useShellLayout: () => mocks.layout,
+}));
+vi.mock("@/lib/portal/use-zone-nav", () => ({
+  useZoneNav: () => ({
+    zones: [
+      { id: "overview", label: "Overview", icon: () => null },
+      { id: "people", label: "People", icon: () => null },
+    ],
+    activeZone: "overview",
+    selectZone: (z: { id: string }) => mocks.selected.push(z.id),
+  }),
+}));
+vi.mock("@/components/app-sidebar-footer", () => ({
+  AppSidebarFooter: () => null,
+}));
+
+import { SidebarProvider } from "@/components/ui/sidebar";
+import { LensRail } from "./lens-rail";
+
+const rail = () =>
+  render(
+    <SidebarProvider>
+      <LensRail />
+    </SidebarProvider>,
+  );
+
+/** The label is present either way; what changes is whether it can be seen. */
+const labelOf = (name: string) =>
+  screen.getByRole("button", { name }).querySelector("span:not(.sr-only)");
+
+beforeEach(() => {
+  mocks.layout = "desktop";
+  mocks.selected = [];
+  window.matchMedia ??= ((query: string) => ({
+    matches: false,
+    media: query,
+    onchange: null,
+    addEventListener: () => {},
+    removeEventListener: () => {},
+    addListener: () => {},
+    removeListener: () => {},
+    dispatchEvent: () => false,
+  })) as unknown as typeof window.matchMedia;
+});
+
+describe("LensRail", () => {
+  it("shows labels while the pointer is on it", async () => {
+    const user = userEvent.setup();
+    rail();
+    expect(labelOf("Overview")).toHaveClass("opacity-0");
+
+    await user.hover(screen.getByTestId("lens-rail"));
+    expect(labelOf("Overview")).toHaveClass("opacity-100");
+  });
+
+  it("collapses on a click and stays collapsed under the pointer", async () => {
+    // The whole reason this state exists. The click navigates; the pointer has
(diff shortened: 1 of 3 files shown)


dzarlax:
Pushed a review round. The body above has been updated to describe what the code now does rather than what the first attempt did — the two differed materially after this, and a description of a design that is no longer there is worse than none.

**The mechanism changed.** The rail opened via a CSS delay on the fade; it now opens on a timer. `pointer-events` is not animatable, so it flipped the instant the state did while the opacity waited out its delay — leaving an invisible, fully clickable band over the pane for that window. Clicks aimed at a row landed on it and disappeared, and a pointer merely crossing the rail entered it, counted as still being inside, and got the rail opened over the row it was heading for. That is precisely what the delay had been written to prevent, and the comment beside it said so. With a timer, open and visible are the same fact and the window does not exist.

**Three ways the state could be stranded**, none of which types or the original four tests would show:

- A keyboard activation set the "stay shut until the pointer leaves" flag. There is no pointer in that story, and nothing else cleared the flag, so pressing Enter left the rail dead to the mouse until the reader happened to move onto it and off again.
- A width change unmounted the rail under the pointer without a leave event, so it came back already open with the pointer nowhere near it.
- Focus never opened it, so a sighted keyboard user navigated eight identical icons with the labels at zero opacity. Screen readers were fine — the text survives `opacity: 0` — which is why this was invisible from the accessibility tree alone.

Each is covered now; two of the new tests fail on the previous commit.

**One fix broke another feature and had to be reconciled.** Removing the blanket child-overflow escape restored the zone list's scrolling and immediately clipped the labels — that escape was what let the widened buttons paint outside the rail. A box cannot clip one axis and release the other, so the two cannot both hold at once. Scrolling is the half worth losing, and only while the labels are showing.

Also: the fade is `absolute` rather than `fixed` (it worked only because the rail happens to sit at the viewport edge and nothing enforces that), starts from the panel's own token rather than the page background (they differ, most visibly in dark theme, which put a step exactly at the seam), mirrors in RTL (a logical inset with a physical gradient put the opaque end on the wrong side and reinstated the hard cut it exists to remove), and renders only when the pane is beside the rail rather than collapsed off-canvas.

On the chart, the readout stood 26px above a strip 8px below its header, so it covered the two comparisons a reader hovers in order to reason against. It has headroom now, and it anchors to its bar's edge and grows inwards instead of centring — centring needs the rendered width to stay inside, and the constant guarding the ends was a guess that longer strings walked past.

The tests also mocked a layout value the app never produces, so the middle width tier — where the pane is collapsed and the fade has no work to do — was untested.

hello1101n (approved):

PR #2445, opened 2026-08-11
fix(frontend): a section says which kind of empty it is

dzarlax:
A section with nothing in it can mean two opposite things:

- the sources behind it are connected and this person did none of that
  work in the period — a fact about them, worth asking about;
- nothing feeds the section at all — a fact about the install, and
  nobody's performance.

The distinction exists (`partState` in `lib/insight/coverage.ts`) and the
sections page already draws it for the whole list via `PersonCoverage`.
It was lost in the two places a reader meets one section at a time.

### The section screen

`SingleGroupView` asked `groupHasData` — whether any metric of the group
holds a value for this person — and rendered one sentence for both
answers. Opening the one section a reader cared enough to click threw
away what the page they came from had told them.

It now reads the same tenant-wide definitions listing the section
navigation already asks for (so it is served from cache) and says which
absence it is. Two consequences worth stating:

- the listing is part of the answer, not decoration, so the screen waits
  for it rather than showing one sentence and swapping it for the other;
- when that request fails it claims **neither** absence. With no listing
  every section looks unreachable, and announcing that nothing is
  measured here for anyone on the strength of a request that never
  arrived is the same class of fabrication the coverage helper exists to
  avoid.

### The navigation mark

The sections nav drew one grey dot for both, so the distinction was
invisible until the section was opened. The standings hook has computed
`peersHaveData` for exactly this since the coverage work landed and
nothing read it.

A filled grey dot now means the section reads and holds nothing for this
person this period. A hollow ring means nothing feeds it — not worth
opening at all until that changes. The tooltip says which.

### Copy

| State | Section screen |
|---|---|
| Sources connected, person absent | Nothing recorded here for the selected period. |
| Nothing feeds the section | No data reaches us for this section — nothing is measured here for anyone yet. |
| Listing could not be read | Nothing to show here for the selected period. |

The first two reuse the vocabulary `PersonCoverage` already uses on the
sections page, so the two screens do not describe the same state in
different words.

### Verification

Four new tests: the section screen names each of the three states, and
the nav marks the two empties apart. Full unit suite green (1087),
typecheck and lint clean.

## Summary by CodeRabbit

* **New Features**
  * Added clearer section indicators distinguishing unavailable data from sections where a person has no data.
  * Added context-specific tooltips and empty-state messages for missing, unreachable, loading, and unavailable metrics.
  * Added retry support when metric data cannot be loaded.

* **Bug Fixes**
  * Improved section rendering while metric definitions load or cannot be retrieved.
  * Clarified no-data messaging for individual reporting periods.
  * Prevented status indicators from appearing for pending sections.

### DIFF
--- a/src/frontend/src/components/portal/context-pane.test.tsx
+++ b/src/frontend/src/components/portal/context-pane.test.tsx
@@ -205,5 +205,9 @@ describe("ContextPane", () => {
     pane();
     const button = screen.getByText("Git output").closest("button")!;
     expect(button.querySelector("span[aria-hidden]")).toBeNull();
+    // And says nothing either. Both flags read false while the queries are in
+    // flight, so a tooltip that trusted them would announce the strongest
+    // claim of the three on an answer the hook has not given.
+    expect(button.getAttribute("title")).toBeNull();
   });
 });
--- a/src/frontend/src/components/portal/context-pane.test.tsx
+++ b/src/frontend/src/components/portal/context-pane.test.tsx
@@ -24,6 +24,7 @@ const mocks = vi.hoisted(() => ({
     status: string;
     phrase: string;
     hasData: boolean;
+    peersHaveData: boolean;
     isPending: boolean;
   }>,
 }));
@@ -136,6 +137,7 @@ describe("ContextPane", () => {
         status: "bad",
         phrase: "4 of 6 behind peers",
         hasData: true,
+        peersHaveData: true,
         isPending: false,
       },
     ];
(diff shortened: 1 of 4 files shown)


dzarlax:
Fixed in cce2a43, and the case is worse than described: while the
queries are in flight `peersHaveData` reads false too, so the tooltip
did not merely show "No data this period" — it showed "No data reaches
us for this section", the strongest of the three claims, on an answer
the hook had not given.

Falling back to `standing.phrase` would not have covered it either: the
phrase is derived from empty rank counts at that moment, so it is
whatever `sectionStandingPhrase` returns for nothing. A pending row now
carries no title at all, which is what the hidden mark beside it already
meant.

The pending test asserts the absent title.

hello1101n (approved):

Issue #2360, opened 2026-08-10
Seeder creates a tenant metric override on a builtin-only stand, breaking export


ktursunov:
## Summary

A fresh demo seed unexpectedly creates a tenant-scoped metric-definition override marked as `origin='custom'`. The seed contract is now builtin-only: tenant custom metrics should exist only when they are explicitly created through the custom-metrics API.

The unwanted seeded override also causes `GET /v1/metrics/export` to return 500, preventing otherwise valid custom metrics from being exported.

## Steps to reproduce

1. Start a fresh compose stand and run the demo seed.
2. Call `GET /v1/metrics`.
3. Observe that it returns 200 but lists one unexpected custom item: label `Stand tenant override` on `ai.accepted_edit_actions`.
4. Optionally create a legitimate custom metric with `POST /v1/metrics`; creation returns 201.
5. Call `GET /v1/metrics/export`.

## Expected

- Immediately after seeding, `GET /v1/metrics` returns an empty custom-metric collection.
- Builtin metrics remain available through the builtin metric catalogue.
- After a custom metric is explicitly created, `GET /v1/metrics/export` returns 200 and includes that metric.

## Actual

- Seeding creates a tenant override even though the stand should contain builtin metrics only.
- `GET /v1/metrics/export` returns 500 before and after creating a legitimate custom metric.
- Deleting the legitimate custom metric does not change the result.

Response:

```json
{"type":"…err.internal.v1~","title":"Internal","status":500,"instance":"/v1/metrics/export","context":{}}
```

Analytics log:

```
analytics::api::metrics: custom metric database operation failed
error=Custom Error: corrupt custom metric row: observation_sql = "NULL"
```

## Root cause

The analytics seeder copies a builtin metric definition into tenant scope, changes its label, and marks the copy as `custom`. The copied definition still points to its builtin managed source, where `observation_sql` is correctly SQL `NULL`.

Export treats every definition marked `custom` as a fully authored custom metric and requires inline `observation_sql`. It therefore mistakes the seed-created override for a corrupt custom metric and aborts the entire export.

The primary contract violation is the seeder creating the tenant override. Export hardening for unexpected or legacy override rows can be handled separately.

## Acceptance criteria

- The demo analytics seed creates no tenant metric-definition overrides.
- A freshly seeded stand contains no tenant custom metrics.
- Seed manifest/profile output no longer advertises a `definition_override`.
[...]

PR #2455, opened 2026-08-12
feat(compose): test-stand up --build runs the whole product from this tree

ktursunov:
## What

`./dev-compose.sh test-stand up --build` now runs the stand entirely from the working tree: the backend compiled from source (the old `--build-backend` path — that spelling is removed; `--build` is the only flag) **and** the frontend pnpm-built, served by the existing `front-built` nginx profile. With the frontend now living in this repo, a local change anywhere in the product can be exercised under the stand's real Keycloak login and dbt readiness gate.

The refusal guard is extended symmetrically: without `--build`, `up` refuses when the tree differs from `origin/main` under `src/frontend/` as well as `src/backend/` — the same protection against reporting green for code that never ran, now via one shared `test_stand_tree_matches_charts` helper.

## Drift fixed along the way

Reviving `FRONTEND_MODE=built` surfaced three integration gaps between the compose dev stack and the in-repo frontend, each fixed where it lives (they affect the dev stack's built mode generally, not just the test stand):

1. **Mode switch died on a name conflict** — the three frontend variants share one `container_name` but are separate compose services, and a profile-inactive sibling's container is not an orphan, so an in-place `up` after a mode switch could not replace it. `cmd_up` now removes the old variant's container first, matched by its compose-service label (same pattern as the existing fakeidp cleanup).
2. **CSP template variable** — the stock nginx entrypoint substitutes only *defined* env vars, so the template's `${SENTRY_CONNECT_SRC}` reached nginx verbatim: `[emerg] unknown "sentry_connect_src" variable`. The `front-built` service now defines it (empty by default), matching the published image's unconditional substitution, and drops `CSP_REMOTE`, which the template no longer reads.
3. **Listen port** — the in-repo template declares `listen 8080` (the published image's non-root layout) while `front-built` still assumed stock nginx's 80; the gateway proxied to a refused port and answered 502. Port map and `FRONTEND_INTERNAL_PORT` now say 8080.

## Verification

On a live bring-up of `test-stand up --build`: the stand reaches the readiness gate, the SPA the gateway serves is **byte-identical** to the freshly built `src/frontend/dist`, and the real-login UI journeys (login, seeded-data) plus the analytics custom-metric API tests pass against the all-local stack — 19 passed, 1 xfailed (the known #2360 export entry). The mode-switch cleanup was exercised live (ghcr → built on a running stack).

### DIFF
--- a/.claude/skills/insight-stand/SKILL.md
+++ b/.claude/skills/insight-stand/SKILL.md
@@ -43,8 +43,7 @@ Two consequences you will meet immediately:
 pass `--build` (the backend compiled from source AND the frontend pnpm-built,
 served by the front-built nginx); `up` otherwise refuses when the tree differs
 from `origin/main` under `src/backend/` or `src/frontend/`, since the pinned
-images would not be what ran. `--build-backend` survives as an alias of
-`--build`.
+images would not be what ran.
 
 **A path argument does not narrow the run.** The verb appends it to a hardcoded
 `tests/stand` and pytest unions path arguments, so `test-stand test
--- a/.claude/skills/insight-stand/SKILL.md
+++ b/.claude/skills/insight-stand/SKILL.md
@@ -38,11 +38,13 @@ Two consequences you will meet immediately:
 ./dev-compose.sh test-stand down    # stop AND remove volumes
 ```
 
-`up` pulls each backend service pinned to its own chart's `appVersion` — never
-`:latest`, never compiled here. To test a *backend* change pass
-`--build-backend`; `up` otherwise refuses when the tree differs from
-`origin/main` under `src/backend/`, since the pinned images would not be what
-ran.
+`up` pulls the frontend and each backend service pinned to its own chart's
+`appVersion` — never `:latest`, never compiled here. To test a local change
+pass `--build` (the backend compiled from source AND the frontend pnpm-built,
+served by the front-built nginx); `up` otherwise refuses when the tree differs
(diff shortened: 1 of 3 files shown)


ktursunov:
Dispatch run on this branch: [31570896023](https://github.com/constructorfabric/insight/actions/runs/31570896023) — **api-smoke ✅** (full contract suite + harness meta tests + coverage gate on a stand brought up through the refactored pinned path), ui-journeys 15/17.

The two red journeys are exactly the known date-dependent drill-down flake — `test_git_commit_bucket_…` and `test_table_total_…` — which #2453 fixes. This branch sits on main without that fix, and the ui lane runs the suite baked into the published ui-tests image regardless. Identical pair failed on the pre-change main dispatch ([31560385144](https://github.com/constructorfabric/insight/actions/runs/31560385144)), so this PR introduces no regression: the default (non `--build`) path — guards, env writer, frontend pin, compose changes — is green on a fresh runner. The `--build` path itself was verified on a live local bring-up (see the PR body).

ktursunov:
Re-ran at head 3b31509a (alias removal): [31572054036](https://github.com/constructorfabric/insight/actions/runs/31572054036) — same picture: **api-smoke ✅**, ui-journeys 15/17 with exactly the two known date-flake drill-down journeys (#2453's subject; baked ui-tests image). No regression from the alias removal.

aleksdotbar (approved):

PR #2457, opened 2026-08-12
security(keycloak): scope the public HTTPRoute to named realms

ktursunov:
## What

The Keycloak subchart's HTTPRoute matched a bare `/kc/realms` PathPrefix, which publishes **every** realm on the shared Gateway — including `master`. Keycloak's `master` realm exposes the `admin-cli` client with the direct password grant, so a bare-prefix route puts the IdP admin account's token endpoint on the public internet with no browser flow, PKCE, or challenge in front of it — a brute-force / password-spray target whose compromise is full auth-plane takeover (every realm, every user, the signing keys).

The route template already took care to keep `/kc/admin` and the admin REST API cluster-internal; this closes the sibling gap on the `/kc/realms` side.

## Change

- `route.realms` (new, **required** when `route.enabled`): the list of realms whose `/kc/realms/<realm>` OIDC surface is published. The route now emits one PathPrefix per named realm plus `/kc/resources` (theme assets).
- An empty `route.realms` **fails the render** rather than falling back to the wide prefix — fail-closed, so a future route can't silently re-widen.
- `master` and every unlisted realm stay cluster-internal.

## Verified

`helm template` renders: `realms: [insight]` → only `/kc/realms/insight` + `/kc/resources`; `realms: []` with the route enabled → render error naming the missing value. Multi-realm (broker) lists render one prefix each.

## Consumers

Any environment that sets `keycloak.route.enabled: true` must now also set `keycloak.route.realms` (e.g. `[insight]`). The one such environment in-tree arrives with a separate change; there is no route enabled on `main` today, so nothing here breaks on merge.

### DIFF
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -160,6 +160,10 @@ jobs:
               - 'deploy/compose/ui-tests.Dockerfile'
             umbrella:
               - 'charts/insight/**'
+              # The keycloak subchart is a file:// dependency packed into the
+              # published umbrella, so a leaf-chart-only change must also make
+              # publish-chart eligible.
+              - 'src/backend/services/keycloak/helm/**'
 
       # Compute the canonical build tag once per workflow run so every
       # downstream job (backend builds + matrix connector builds + toolbox +
--- a/.github/workflows/keycloak-helm.yml
+++ b/.github/workflows/keycloak-helm.yml
@@ -1,0 +1,50 @@
+name: Keycloak — Helm contract
+
+# Render-contract gate for the keycloak subchart's HTTPRoute realm allowlist:
+# the original finding was a bare /kc/realms PathPrefix publishing the master
+# realm's admin-cli token endpoint, so the exact set of published path values
+# is contract, not plumbing. Pure `helm template` + assertions — no cluster,
+# no images.
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - "src/backend/services/keycloak/helm/**"
+      - "charts/insight/**"
+      - "deploy/gitops/environments/local/values.yaml.template"
+      - ".github/workflows/keycloak-helm.yml"
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  helm-contract:
+    name: render contract
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      # Pinned to full SHAs (mutable tags are repointable — supply-chain
+      # hardening, same as identity-resolution-helm.yml).
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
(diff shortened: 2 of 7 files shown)


ktursunov:
Please add these five pieces to the PR:

1. **Trigger chart publication for Keycloak-only chart changes** in `.github/workflows/build-images.yml`:

```yaml
umbrella:
  - 'charts/insight/**'
  - 'src/backend/services/keycloak/helm/**'
```

Without this, a merge that changes only the Keycloak leaf chart does not make `publish-chart` eligible, so the updated subchart never reaches the published umbrella chart.

2. **Validate every realm entry** in `src/backend/services/keycloak/helm/templates/httproute.yaml`:

```gotemplate
{{- range .Values.route.realms }}
{{- $realm := required "keycloak.route.realms entries must be non-empty" . }}
- path:
    type: PathPrefix
    value: {{ printf "/kc/realms/%s" $realm | quote }}
{{- end }}
```

Keep the existing empty-list guard. A list containing an empty or null item currently passes that guard and renders `/kc/realms/`; Gateway API ignores the trailing slash for `PathPrefix`, making this equivalent to the unsafe bare `/kc/realms` prefix.

3. **Update the committed local deployment overlay** in `deploy/gitops/environments/local/values.yaml.template`:

```yaml
keycloak:
  route:
    enabled: true
    realms: [insight]
```

That overlay already enables the Keycloak route, so it fails to render after `route.realms` becomes required unless the realm is added atomically.

4. **Mirror and document the public value** in `charts/insight/values.yaml`:

```yaml
keycloak:
  route:
    enabled: false
    realms: []
```

5. **Add automated Helm render-contract tests** covering:

- An empty list fails when the route is enabled.
- `[""]` and `[null]` fail.
- One realm renders only its exact realm prefix plus `/kc/resources`.
- Multiple realms render one exact prefix each.
- No successful render contains `/kc/realms` or `/kc/realms/` as a complete path value.
- The committed local GitOps overlay renders successfully.

Optional hardening: explicitly reject `master` if publishing that realm should never be supported, even through deliberate configuration.

---

**Done — all five pieces plus the optional hardening landed in 01889b30:**

1. **Publish trigger** — `build-images.yml` counts `src/backend/services/keycloak/helm/**` toward the umbrella filter.
2. **Per-entry validation** — each `route.realms` entry passes through `required` (empty and null entries fail the render instead of degrading to the bare `/kc/realms/` prefix), and `master` is rejected outright.
3. **Local overlay** — `deploy/gitops/environments/local/values.yaml.template` carries `realms: [insight]`.
4. **Umbrella values** — `charts/insight/values.yaml` mirrors and documents `keycloak.route.realms: []`.
5. **Render-contract tests** — `src/backend/services/keycloak/helm/tests/test_httproute_contract.py`, run by the new `keycloak-helm.yml` workflow (mirrors the identity-resolution helm-contract lane): every bullet above is a test, including that no successful render contains `/kc/realms` or `/kc/realms/` as a complete path value and that the committed local overlay renders the umbrella. 9 tests green locally; `helm lint` clean on both charts.

*(This comment was accidentally overwritten by a tooling flag and restored; the completion summary that replaced it is appended above rather than kept as a separate comment.)*

PR #2458, opened 2026-08-12
ci(e2e-stand): trigger the stand suite on any backend, frontend or tests change

ktursunov:
## What

Adds an explicit `paths:` filter to the `pull_request` trigger of `.github/workflows/e2e-stand.yml`, mirroring the `changes` job's relevance regex, and broadens both in the same step:

- `tests/**` (was `tests/stand/`, `tests/lib/`, `tests/pyproject.toml`, `tests/uv.lock`)
- `src/frontend/**` (was `src/frontend/helm/` only)
- `src/backend/**` (unchanged)
- stand infra unchanged: `deploy/compose/**`, seed tool, `docker-compose.yml`, `dev-compose.sh`, backend `openapi.json` docs, this workflow, the trace-redaction script

## Why

The compose-stand suite must go with any change to the backend, the frontend, or the tests themselves. The previous scope missed frontend source changes and non-stand test changes.

## Notes

- On `pull_request` every job still skips by design, so the new filter only prunes no-op skipped runs from the Actions tab.
- The `changes` job keeps doing the same filtering for `merge_group` (the event supports no `paths:` key); the two lists are documented as needing to stay in sync.
- While the paths filter exists, `Stand E2E` can only be made required on `merge_group`, never as a PR-level required check — a required check whose workflow never triggers reports as pending forever and would block queue entry. Verified the current main ruleset requires only "Run E2E suite" and "secrets (diff)", so nothing breaks today; the workflow comments record the constraint.

### DIFF
--- a/.github/workflows/e2e-stand.yml
+++ b/.github/workflows/e2e-stand.yml
@@ -21,9 +21,9 @@ name: E2E — Compose stand
 #                published ui-tests image.
 #
 # Neither is required on its own. The `Stand E2E` umbrella below is the check
-# meant to be required in branch protection, and the whole workflow is a
-# merge-queue gate rather than a per-push one: on pull_request every job here
-# skips (branch protection counts a skipped job as a pass), and the lanes run
+# meant to be required for the merge queue (merge_group), and the whole
+# workflow is a merge-queue gate rather than a per-push one: pull_request is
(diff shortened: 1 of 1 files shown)


ktursunov:
Absorbed into #2453 (merged tests-2163 in as 1f3b4392) so the workflow hardening and the stand fixes land and get exercised together.

PR #2459, opened 2026-08-12
feat(gitops): prod environment — InsightCfabric stand under gitops management

ktursunov:
Adds the `prod` gitops environment for the InsightCfabric stand, shaped after the reference insight-gitops flow, plus the small pieces the rest of the tree needs to support it.

## What's in here

- **`environments/prod/`** — inventory (cluster identity, edge, `chartVersion:` pin), umbrella `values.yaml`, flat per-service values overlays, concrete operator-CR manifests under `manifests/`, the broker realm as code, and an env-scoped Makefile (`kube-ctx` guard → `bootstrap` → `secrets` → `system` → `deploy CONFIRM=yes-deploy-prod`). The env Makefile exists because this stand runs operator-managed datastores one-namespace-per-service, which the top-level chart-based `system-*` targets don't fit; it keeps the same target vocabulary.
- **`bootstrap/prod/`** — concrete L0 manifests (storage class, EnvoyProxy + shared Gateway, cert-manager CA chain). No placeholder rendering anywhere; committed YAML is the applied state.
- **SealedSecrets** — every credential the stand needs is committed as cluster-bound ciphertext under `sealed-secrets/<ns>/`, applied by `make secrets` (waits for Synced). No cleartext, templates, or sealing tooling in the tree; rotation is a plain `kubeseal` against the cluster controller.
- **Top-level Makefile** — chart-version resolution now honors a per-env `inventory.chartVersion` pin before falling back to `.insight-version` (same order as the reference repo).
- **Tenant contract guard** — the static check now recognizes the social-provider mapper config key (`userAttribute`) when verifying the `idp_sub` stamp; the OIDC key alone rejected valid GitHub-broker registrations.

## Verification

- `kubectl diff` over every committed manifest against the live stand: zero drift.
- `helm upgrade --dry-run=server` at the pinned chart rendered byte-identical to the deployed release at porting time; the stand runs the pinned chart deployed through `make deploy`, and the hardening in this PR (ClickHouse/Keeper securityContext) has been applied and re-verified drift-free.
- All SealedSecrets applied and Synced; realm contract guard passes for both committed realms.

## Notes for review

- Layer naming follows the deployment glossary (L0/L2/L3, no L1).
- `bootstrap:`/`system:` inventory toggles are deliberately false — they drive the top-level targets; the env Makefile owns the operator path (documented in the inventory header and env README).


ktursunov:
Note for reviewers on the one red check: **Trivy** fails on (a) dependency CVEs in lockfiles this PR does not touch — they surface as "new" only because the merge-ref scan is fresher than main's nightly baseline, and will fold into it on the next scheduled run — and (b) a set of waived-by-design misconfiguration findings on the storage-init chown Job (root is its purpose; rationale in the manifest) and the Airbyte launcher RBAC, pending dismissal. Every finding this PR could genuinely fix is fixed: `sast` and `Semgrep OSS` are green with zero open alerts on the gitops paths.

deploy/gitops/environments/prod/inventory.yaml
```diff
@@ -0,0 +1,107 @@
+##
```
ktursunov, line 1: [P0] Move concrete deployment state out of this repository

The root repository policy forbids GitHub-visible production-derived infrastructure and deployment details. This newly added concrete environment tree is therefore outside the allowed confidentiality boundary. Move the environment-specific overlays, manifests, ciphertext, and RBAC into the authorized private GitOps repository, and keep only clearly synthetic reusable templates in this repository.

deploy/gitops/environments/prod/values.yaml
```diff
@@ -0,0 +1,180 @@
...
+    requests:
+      cpu: 200m
+      memory: 256Mi
+    limits:
+      cpu: 1
+      memory: 512Mi
+
+authenticator:
+  replicaCount: 1
+  overrideEnabled: true
```
ktursunov: [P0] Disable unrestricted view-as in protected environments

The chart values document this switch as dev/demo-only and require it to remain false anywhere real users can authenticate. Enabling it lets a caller request an override identity without an authorization boundary in this configuration. Keep this false here; if administrator impersonation is needed, implement explicit role checks, allowlisting, and audit logging as a separate capability.

deploy/gitops/environments/prod/Makefile
```diff
@@ -0,0 +1,412 @@
...
+# --force-conflicts alone against a client-side release is rejected outright.
+# Helm 3 gets no flags.
+HELM_UPGRADE_HELP := $(shell helm upgrade --help 2>/dev/null)
+HELM_APPLY_FLAGS := $(if $(findstring --server-side,$(HELM_UPGRADE_HELP)),--server-side=true --force-conflicts,$(if $(findstring --force-conflicts,$(HELM_UPGRADE_HELP)),--force-conflicts,))
+
+# ── Sealed-secrets pins ──────────────────────────────────────────────────────
+# Everything under sealed-secrets/ is sealed against THIS cluster's controller
+# key and decrypts only here. Re-seal everything if the key rotates or the
+# cluster is rebuilt.
+SEALED_SECRETS_VERSION   ?= 2.17.4
```
ktursunov: [P1] Upgrade the controller past the rotate scope-widening fix

Chart 2.17.4 deploys controller 0.31.0, which is below the 0.36.0 patched version for GHSA-465p-v42x-3fmj. An actor who can invoke the rotate endpoint can use attacker-controlled template annotations to widen sealing scope. Pin a supported chart whose controller appVersion is at least 0.36.0 and verify upgrade compatibility. Advisory: https://github.com/bitnami-labs/sealed-secrets/security/advisories/GHSA-465p-v42x-3fmj

deploy/gitops/environments/prod/Makefile
```diff
@@ -0,0 +1,412 @@
...
+
+deploy-insight: kube-ctx confirm keycloak-broker-realms argo-rbac
+	@for s in insight-db-creds insight-oidc insight-keycloak-admin insight-keycloak-config github-oauth insight-authenticator-signing-keys; do \
+	  kubectl -n $(NS_APP) get secret $$s >/dev/null 2>&1 || \
+	    { echo "ERROR: Secret $(NS_APP)/$$s missing — run 'make secrets'"; exit 1; }; done
+	helm upgrade --install $(RELEASE) $(INSIGHT_CHART) \
+	  --version $(INSIGHT_VERSION) \
+	  --namespace $(NS_APP) --create-namespace \
+	  --values values.yaml \
+	  --set-string authenticator.oidc.clientSecret="$$(kubectl -n $(NS_APP) get secret insight-oidc -o jsonpath='{.data.client-secret}' | base64 --decode)" \
```
ktursunov: [P1] Do not inject plaintext credentials through Helm values

This command substitution decrypts the credential into the Helm process arguments, and the set value is retained in Helm release configuration. That creates plaintext copies outside the Kubernetes Secret and defeats the intended Sealed Secrets boundary. Extend the chart to consume an existing Secret/key via secretKeyRef, then pass only the Secret name and key.

deploy/gitops/environments/prod/values.yaml
```diff
@@ -0,0 +1,180 @@
...
+    requests:
+      cpu: 500m
+      memory: 1Gi
+    limits:
+      cpu: 2
+      memory: 2Gi
+keycloakConfig:
+  enabled: true
+  url: http://insight-keycloak:8085/kc
+  allowInsecureUrl: true
```
ktursunov, line 45: [P1] Do not send administrator credentials over plain HTTP

The chart explicitly limits allowInsecureUrl to dev/CI because the config Job authenticates to this URL with administrator credentials. Use HTTPS with a trusted CA or another authenticated encrypted in-cluster endpoint, and leave allowInsecureUrl false.

deploy/gitops/environments/prod/Makefile
```diff
@@ -0,0 +1,412 @@
...
+	@# not race the ownership fix.
+	kubectl -n $(AIRBYTE_NAMESPACE) delete job -l app.kubernetes.io/name=airbyte-storage-init --ignore-not-found
+	kubectl apply -f manifests/airbyte/storage-init.yaml
+	kubectl -n $(AIRBYTE_NAMESPACE) wait --for=condition=Complete job -l app.kubernetes.io/name=airbyte-storage-init --timeout=$(AIRBYTE_TIMEOUT)
+	kubectl apply -f manifests/airbyte/rbac.yaml
+	helm repo add airbyte https://airbytehq.github.io/helm-charts --force-update
+	@# Postgres password: reuse the chart-generated one when the release
+	@# already exists; only a fresh install mints a new one. Passing nothing
+	@# would silently fall back to the chart default (the literal 'airbyte').
+	@existing=$$(kubectl -n $(AIRBYTE_NAMESPACE) get secret airbyte-airbyte-secrets \
```
ktursunov: [P1] Do not rotate the database password on lookup failures

Redirecting errors and ending the lookup with true makes Forbidden, timeout, transport, and decode failures indistinguishable from Secret NotFound. Any such failure produces a new random password and can break an existing database during upgrade. Generate only after an explicit NotFound result and fail on every other error. Prefer an existing-Secret reference so the password is never stored as a Helm value.

ktursunov (commented):
Review outcome: blocking changes are required.

Sealed Secrets is a reasonable Git transport mechanism when strict object scope, a supported controller, an external source of truth, and tested key recovery are all in place. The manifests here use strict name/namespace scope and do not contain plaintext Secret fields, but the surrounding deployment flow currently weakens those guarantees. The main blockers are the repository confidentiality boundary, a dev-only impersonation switch enabled in a protected environment, a controller version below a published security fix, plaintext credentials flowing through Helm release values, insecure credential transport, and failure paths that can silently report success or rotate a credential.

I added implementation-focused inline comments with the required fixes. This review is submitted as a comment because GitHub does not allow an author to formally request changes on their own PR.

deploy/gitops/environments/prod/Makefile
```diff
@@ -0,0 +1,412 @@
...
+	@# One-shot instance setup: with auth enabled a fresh instance sits in a
+	@# wizard until email+org are posted. Idempotent — replayed every run,
+	@# from inside the cluster so it exercises the Service DNS name.
+	@kubectl -n $(AIRBYTE_NAMESPACE) delete pod airbyte-setup-oneshot --ignore-not-found >/dev/null 2>&1; \
+	kubectl -n $(AIRBYTE_NAMESPACE) run airbyte-setup-oneshot --restart=Never --image=$(AIRBYTE_CURL_IMAGE) \
+	  --command -- sh -c 'B=http://airbyte-airbyte-server-svc:8001; \
+	    curl -sf -X POST "$$B/api/v1/instance_configuration/setup" \
+	      -H "Content-Type: application/json" \
+	      -d "{\"email\":\"$(AIRBYTE_SETUP_EMAIL)\",\"organizationName\":\"$(AIRBYTE_SETUP_ORG)\",\"initialSetupComplete\":true,\"anonymousDataCollection\":false}" \
+	    || true'
```
ktursunov: [P1] Do not make setup failures look successful

The unconditional true means DNS, authentication, transport, and server errors all make this Pod succeed, so the deployment continues without completing setup. If an already-configured response is the intended idempotent case, inspect the HTTP status/body and accept only that exact response; fail every other result.

deploy/gitops/environments/prod/Makefile
```diff
@@ -0,0 +1,412 @@
...
+        keycloak-broker-realms argo-rbac status system-status rollback secrets
+
+# Both stand kubeconfigs ship their context named `default` — rename it to
+# `insight-prod` once per machine (repo README prerequisite #4):
+#   kubectl config rename-context default insight-prod
+# The context name alone identifies nothing, so the API server URL from the
+# inventory is the real wrong-cluster protection. Every mutating target
+# depends on this — including the per-component ones.
+kube-ctx:
+	@CUR=$$(kubectl config current-context 2>/dev/null || true); \
```
ktursunov: [P1] Pin every mutation to the validated context

This prerequisite validates the current context once, but all later kubectl and Helm commands reread mutable global kubeconfig state. A context change between the check and a mutation bypasses the guard. Pass --context to every kubectl invocation and --kube-context to every Helm invocation, or run the target with an isolated kubeconfig.

deploy/gitops/environments/prod/README.md
```diff
@@ -0,0 +1,82 @@
...
+  The equivalent targets in this env's Makefile are the managers; same
+  vocabulary (`bootstrap-*`, `system-<svc>`), operator implementations.
+- **Edge is a shared Envoy Gateway** (`envoy-gateway-system/insight`) with the
+  floating IP pinned on its Service (see `inventory.edge`); the chart's own
+  HTTPRoutes (`gateway.route`, `keycloak.route`) attach to it. Cloudflare
+  fronts the hostname; zone SSL mode must stay "Full".
+- **Secrets are sealed per service namespace** — the schema's
+  `secrets.services` list is joined by a `secrets.datastores` list (each entry
+  naming its own namespace), matching the one-namespace-per-service topology.
+  Ciphertext is bound to this cluster's controller key: a cluster rebuild or
```
ktursunov, line 75: [P1] Define a recoverable secret lifecycle

Re-sealing after a controller-key loss or cluster rebuild requires either recovery of the controller private key or authoritative plaintext credentials from an external source. The committed ciphertext alone is not a recovery source. Document and automate a password-manager-backed source plus encrypted off-cluster key backup/restore, or explicitly define full credential regeneration; test the restore path before treating this as the whole secret story.

ktursunov:
Status after the fix rounds (branch rebased onto current main, all checks green). This supersedes my earlier note about Trivy — the waivers landed in `.trivyignore.yaml` and the check passes.

**Addressed from the review:**
- `confirm` gate fails closed; every kubectl/helm invocation is pinned to the inventory context and `kube-ctx` validates the *named* context's API server (TOCTOU closed)
- airbyte postgres password: only an explicit NotFound mints a new one; empty values and other lookup failures abort
- airbyte setup one-shot no longer masks failures
- oidc client secret is verified non-empty before the upgrade (full fix — consuming it via secretKeyRef — needs chart support, tracked as follow-up)
- sealed-secrets controller 2.17.4 → 2.19.1 (0.38.4, covers GHSA-465p-v42x-3fmj), upgraded on the stand, all 12 SealedSecrets re-verified Synced
- `authenticator.overrideEnabled: false` committed **and made effective on the stand** — the flag alone was not enough: the old `override_enabled` key in the config Secret predated Helm's field ownership, so server-side apply never removed it; the orphaned key was patched out and the authenticator restarted. All four chart-rendered Secrets now match the render exactly.
- Trivy misconfigurations: real fixes (read-only rootfs, RuntimeDefault seccomp, resources on the storage-init Jobs) plus scoped, statement-carrying waivers for root-by-design findings

**Open, tracked as follow-ups:** chart-level secretKeyRef for the oidc secret; keycloak-config over in-cluster TLS; documented key-recovery/backup path; dependency-CVE bumps in the lockfiles (pre-existing, not this PR).

A second review pass has surfaced ten further findings (top-level Makefile parity with the env hardening, two bugs in the fix round itself, a leftover sample DB account in the MariaDB CR); fixes are queued.

ktursunov:
@coderabbitai review

ktursunov:
Closing: this stand's gitops configuration now lives in the internal insight-gitops repository as environment `cfabric`, alongside the other stands, rather than in the application repo.

Nothing here consumed it — no workflow, script or Makefile target referenced `deploy/gitops/environments/prod` or `deploy/gitops/bootstrap/prod`.

Three changes on this branch were environment-independent and are not carried over; if they are wanted they should come back as their own PR:

- `deploy/compose/keycloak/tests/tenant_contract_guard.py` — accept the social-provider `userAttribute` mapper key alongside `user.attribute`, so a valid GitHub identity-provider mapper is not rejected.
- `deploy/gitops/.gitignore` — un-ignore `*.env.template` and `secrets.local.*.template`, which `secrets.local.*` was swallowing.
- `deploy/gitops/Makefile` — resolve the chart version from a per-env inventory `chartVersion:` pin before falling back to `.insight-version`.

PR #2465, opened 2026-08-12
fix(frontend): make a section's wide table admit what it is hiding

dzarlax:
A pass over the two table blocks at the top of a person's section
screen. Everything here is about the same mismatch: the table holds far
more than the box shows, and nothing on screen admitted it.

### What was wrong

| | Before |
|---|---|
| Header | Empty. The metric selector is chart-only and the plain heading needs a single metric, so a grouped multi-metric table matched neither branch. |
| Grand total, scrolled across | Gone. Its cell spans every group, so its left edge is off-screen once the table scrolls; left-aligned inside it, the totals went with it and left the sticky label beside nothing. |
| Grand total, scrolled down | Gone, for the ordinary reason a long table scrolls. |
| Edges | Nothing said the table continued. The overlay scrollbar stays hidden until a scroll begins, so at first sight there is no affordance at all, and a column cut mid-group reads as a rendering fault. |
| Card height | Fixed for both presentations, though only a chart needs it. A short table sat above a slab of empty card. |

### What it does now

- The block is named by its grouping — the metrics are already named on the columns — and carries the group count. That count is load-bearing: a reader seeing three groups beside a grand total covering eleven is looking at numbers that do not add up, with nothing on screen to explain the difference.
- The totals stick past the label column, and the footer is pinned to the foot of the box with a top edge, so it reads as a summary over more rows rather than as the end of the table.
- Arrows appear on each side that still holds something, and page by just under a frame so the edge row or column stays on screen.
- A table sizes to its rows, keeps the old figure as a ceiling, and offers to drop the ceiling when the rows overrun it. The offer appears only when it would do something, and is not persisted — a card remembered at full height would greet the next visit as a wall.

### Two things deliberately not done

**A metric picker on the table.** The chart below has one, but that is a constraint rather than a feature: a stacked chart cannot draw two metrics at once. Several metrics side by side is what the table is for, so narrowing it to one would have made it a worse chart without touching the actual complaint.

**An edge fade, and a forced-visible scrollbar.** Both were built and removed. A fade suggests more by veiling what lies under it, and over a pale table whose edge cells are frequently empty there is nothing to veil. Forcing the scrollbar visible does not survive a platform that draws scrollbars as overlays — the CSS applies and the bar still takes no space.

### Verification

Behaviour was checked in a browser by measurement rather than by eye: the totals' position at both ends of the horizontal scroll, the footer pinned while the box scrolls, the card shrinking to content and capping, and the arrows appearing per side against the measured overflow on each axis.

Tests cover the header and its count, the card's height per presentation, the sticky totals and the pinned footer. **The arrows and the full-height control are not covered**: they depend on measured overflow, and jsdom reports every dimension as zero, so both stay hidden there. Faking layout to assert them would test the fake. Full unit suite green (1091), typecheck and lint clean.

## Summary by CodeRabbit

* **New Features**
  * Added horizontal scroll controls for timeseries tables with directional paging and smooth navigation.
  * Added expand/collapse controls for tables with vertical overflow.
  * Grouped tables now display grouping labels and column counts.
  * Grand totals remain visible during horizontal scrolling, while footers stay fixed at the bottom.

* **Bug Fixes**
  * Improved table and chart sizing for clearer, more consistent layouts.
  * Table navigation updates appropriately as content size and scrolling change.


src/frontend/src/components/widgets/metric-views/metric-timeseries-table.tsx
```diff
@@ -19,13 +29,70 @@ import type { MetricTimeseriesTableConfig } from "@/lib/metrics/timeseries-table
 export interface MetricTimeseriesTableProps {
   model: MetricTimeseriesModel;
   config?: MetricTimeseriesTableConfig;
+  /**
+   * Whether the rows overrun the box vertically — the card above offers to
+   * let the table out to full height, and only when that would do anything.
+   * The fact travels up rather than the node: the card has no use for the
+   * element, and passing a ref back into a child's ref merge is what the
+   * immutability rule forbids.
+   */
```
aleksdotbar, line 35: redundant, remove

aleksdotbar (dismissed):
Clean up the walls of comments please.
Most of the code is self-explanatory and doesnt need a comment.

dzarlax:
All four applied in e9927f0.

**Arrows towards an empty side are no longer rendered.** The opacity
guard was the wrong tool — the button stayed in the tab order and the
accessibility tree either way. This one is now covered by a test: with
nothing overflowing, no paging control exists in the DOM at all. It
fails without the fix, since the old code rendered four hidden ones.

**The observer watches the box as well as the table.** Agreed on the
reasoning: a box that narrows around an unchanged table is exactly the
case the old code missed.

**Expansion is cleared with the overflow flag** on a presentation
change, in the same render-phase reset.

**The note on `onVerticalOverflow` is trimmed** — thanks, it was
justifying an approach that is not in the code, which is noise for
anyone reading the prop.

Verified in a browser afterwards: with no horizontal overflow and 107px
of vertical, exactly one control is rendered — "Show later rows" — and
no hidden ones sit beside it.

Two of the four still have no unit test, for the reason already in the
description: they turn on measured overflow, and every dimension is
zero in jsdom, so neither the arrows nor the full-height control can be
exercised there.

Full suite green (1092), typecheck and lint clean.

dzarlax:
@aleksdotbar your note generalised, in 062021c3.

Measured before touching anything: **105 of 352 added non-blank lines
were comments — 30%**. It is now 36 of 267, so about 13%.

Three kinds went out:

- **History.** "This was a render loop once", "a fade was tried and does
  not work". It is in the preceding commits and in git blame, and it
  ages worst of all — a note about a loop that no longer exists.
- **Restatements.** Five lines above `SIDES.filter(side => more[side])`
  explaining that only available sides render. The filter says it. This
  was your line.
- **Rationale for roads not taken**, including the one you flagged.

What stayed names a constraint the code cannot hold and that someone
would otherwise undo: `scrollLeft` is negative in RTL, `model` must
stay out of the effect deps, the reset happens during render because
the lint rule forbids an effect, the arrows are rendered rather than
hidden so they stay out of the tab order.

Two numbers became names rather than prose — `EDGE_SLACK_PX` for the
sub-pixel tolerance, `NEARLY_A_FRAME` for the paging step — which is
the better half of the fix: the comment was standing in for something
that wanted a name.

No behaviour change. Full suite green (1092), typecheck and lint clean.

hello1101n (approved):

hello1101n (approved):

PR #2475, opened 2026-08-13
ci(e2e-stand): always report the required umbrella check

ktursunov:
## What

- Run the Compose-stand workflow on every pull request so its stable `Stand E2E` context always exists.
- Keep the path decision inside the cheap `changes` job, so `api-smoke` and `ui-journeys` still run only for relevant changes.
- Compare pull-request changes from the merge base to avoid running the expensive stand for unrelated changes that landed on `main`.
- Make the umbrella job fail closed when relevance is unknown or either required lane does not succeed.

## Why

The repository ruleset currently requires the conditional `api-smoke` and `ui-journeys` jobs. When a pull request does not match the workflow-level path filter, the workflow never starts and GitHub waits indefinitely for status contexts that cannot be reported.

The stable umbrella produces the intended behavior:

| Change | Lane result | `Stand E2E` |
| --- | --- | --- |
| Stand-relevant | Both lanes succeed | Pass |
| Stand-relevant | Either lane fails or skips | Fail |
| Not stand-relevant | Both lanes skip | Pass |

## Ruleset follow-up

After this change lands, update **Settings → Rules → Rulesets → Soft Protection → Require status checks to pass**:

- remove `api-smoke`
- remove `ui-journeys`
- add `Stand E2E`
- retain `Run E2E suite` and `secrets (diff)`

## Validation

- `actionlint .github/workflows/e2e-stand.yml`
- `pre-commit run --files .github/workflows/e2e-stand.yml`
- Exercised the umbrella shell step for irrelevant, both-success, lane-failure, and invalid-relevance cases.

### DIFF
--- a/.github/workflows/e2e-stand.yml
+++ b/.github/workflows/e2e-stand.yml
@@ -23,9 +23,10 @@ name: E2E — Compose stand
 #                what executes.
 #
 # Neither is required on its own; `Stand E2E` below is the one stable context
-# for branch protection. The stand goes with any change it covers: the lanes
-# run on every pull_request whose diff matches the paths filter, once per
-# queue batch on merge_group, nightly, and on dispatch. Superseded PR runs
+# for branch protection. The workflow starts on every pull request so that
+# required context always reports; the cheap `changes` job decides whether the
+# two stand lanes need to run. Relevant changes run on the pull request, once
+# per queue batch on merge_group, nightly, and on dispatch. Superseded PR runs
 # are cancelled by the concurrency group. Wall-clock and flake history are in
 # .cf-studio/.plans/stage1-compose-e2e-suite/out/ci-wiring.md.
 #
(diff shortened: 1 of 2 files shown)


ktursunov:
## End-to-end gate evidence

Both relevant-change outcomes were exercised on this pull request:

- **Failed lane blocks:** [run 31659223459](https://github.com/constructorfabric/insight/actions/runs/31659223459) used a temporary API-test probe. `api-smoke` failed, `ui-journeys` passed, and `Stand E2E` failed with `api-smoke=failure`. GitHub reported the pull request as blocked at that head.
- **Successful lanes pass:** [run 31659587440](https://github.com/constructorfabric/insight/actions/runs/31659587440) ran from the final clean head. `api-smoke`, `ui-journeys`, and `Stand E2E` all passed.

The intentional failure probe was rewritten out of the branch. The retained `tests/stand/README.md` change keeps the final head in the stand-relevant path set, so the green run exercised both real lanes rather than a skipped/irrelevant path.

hello1101n (approved):

Issue #2163, opened 2026-08-04
Login with GitHub


ktursunov:
## Goal

Support **GitHub as a login provider for Insight** — a user signs in with their GitHub
account instead of Microsoft Entra, Okta, or a username/password. GitHub is the identity
every engineering org using this product already has, so it is the lowest-friction way to
get real people into an Insight instance.

## User flow

1. User opens `https://<instance>/` and lands on the sign-in screen.
2. The screen offers **"Continue with GitHub"** alongside the instance's other configured
   sign-in options.
3. User clicks it → GitHub asks them to authorize Insight (first time only; returning
   users skip straight through).
4. GitHub returns them to Insight, which identifies them by their GitHub **primary
   verified email**.
5. User lands on their Insight dashboard, signed in as the person that email belongs to.

**Cost to the user: one click** — two the very first time, counting GitHub's authorization
screen.

Unhappy paths:

- **The GitHub email matches no person in the instance** → sign-in is refused with the
  existing "unknown person" behaviour and audit event. No account is auto-created.
- **The GitHub account has no verified email** → cannot sign in; the user is told why.
- **User cancels at GitHub** → returns to the sign-in screen with the other options intact.

## Scope

- GitHub becomes a selectable sign-in provider for an Insight instance, configured per
  instance and off by default.
- An instance can offer GitHub *alongside* its existing providers — enabling it does not
  disable Entra, Okta, or seeded logins.
- Access control is unchanged: the GitHub email must resolve to an existing person.
  Signing in with GitHub grants no access that email would not already have.
- Credentials for the GitHub integration are held as instance secrets — never committed
  to the repo or baked into an image.

**Constraint worth knowing up front:** GitHub is not an OpenID Connect provider for user
sign-in — its OAuth Apps are plain OAuth 2.0
([docs](https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps)),
returning an opaque access token with no id_token, no discovery document and no JWKS.
(`token.actions.githubusercontent.com` serves Actions workloads, not people.) Insight's
sign-in path expects an OIDC issuer, so GitHub has to be adapted rather than plugged in
directly. How that adaptation is done is an implementation decision — see #1782.

## Acceptance criteria

- [ ] AC-1. On an instance with GitHub enabled, "Continue with GitHub" appears on the sign-in screen and a user whose GitHub primary verified email matches a person completes sign-in with a working session.
[...]

PR #2497, opened 2026-08-13
test(authenticator): GitHub-brokered login e2e against the documented realm shape (#2163)

ktursunov:
## What

Implements **#2163 Testing scenario 8** (`auth-rig`): the authenticator e2e rig gains the GitHub-brokered login leg, run by `run-e2e.sh` in the existing `Authenticator — e2e, endpoint coverage` lane.

- **`tests/github-stub.py`** — a rig-local GitHub: authorize (auto-approve), token exchange (client pair checked), `/user` with a null email, `/user/emails` with a non-primary decoy plus the primary verified email. Keycloak's built-in `github` provider reaches it through its GitHub-Enterprise `baseUrl`/`apiUrl` seam, so the **real social importer** runs with no real GitHub involved.
- **`kc-realm-overlay.py`** — registers the identity provider in the documented shape (`trustEmail`, `defaultScope: user:email`, the `hardcoded-attribute-idp-mapper` tenant pin, the `github-user-attribute-mapper` `jsonField: id -> userAttribute: idp_sub`) plus the canonical realm's `idp_sub` token passthrough; the host-keyed second realm stays broker-free.
- **`tests/e2e_broker.rs`** — two loops:
  - *known email*: full chain through the authenticator (`kc_idp_hint=github`, first-broker-login handled if served) → session cookie; gateway JWT tenant = the pin, a single string, deliberately **not** the roster tenant; federated-identity link carries the GitHub numeric id; the id_token claim contract (`tenant_id` pin + `idp_sub`) read straight off the IdP; session survives token rotation (15s realm lifespan).
  - *unknown email*: the brokered login completes at Keycloak but resolves to no person → `auth_error=access_denied` redirect, **no session cookie** (AC-2's mechanism, minus the audit-stream count that stays with scenario 2 on a deployed stand).

## Why

Scenario 8 is the only #2163 scenario that can run continuously in CI: it freezes the operator instruction's realm shape (#2470) as an executable contract long before the dedicated deployment exists. The GitHub-specific leg stays with scenarios 4 and 7.

## Verified

- Both loops' Keycloak-side behaviour replayed against a standalone Keycloak 26.4 import of the overlay realm: 6-hop chain to the RP callback, tenant pin + `idp_sub` present in the id_token, federated-identity link correct, `emailVerified` true via `trustEmail`.
- One deliberate finding baked into the test design: KC 26 hides unmanaged user attributes from admin-API reads, so the token claims — not user representations — are the observable proof the mappers stamped.
- `cargo test --no-run`, `cargo clippy --tests` (0 findings), `cargo fmt --check`, ruff clean.

Refs #2163.

### DIFF
--- a/src/backend/services/authenticator/tests/common/kc.rs
+++ b/src/backend/services/authenticator/tests/common/kc.rs
@@ -134,8 +134,7 @@ pub async fn broker_login_github(
 }
 
 /// The brokered hop chain up to (not including) the authenticator callback:
-/// returns the callback URL so the caller can either follow it or exchange
-/// the code straight at the IdP (the claim-contract assertions).
+/// returns the callback URL for the caller to follow.
 pub async fn broker_callback_url(http: &super::Client, auth_base: &str, stub_user: &str) -> String {
     let login = http
         .get(format!("{auth_base}/auth/login"))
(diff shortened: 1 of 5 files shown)


ktursunov:
Closing: the brokered leg here runs against a stubbed upstream, so it is not the full end-to-end proof the scenario is meant to give. The coverage approach for #2163 scenario 8 will be revisited.

PR #2501, opened 2026-08-13
seed(stand): two years of activity on both stands

ktursunov:
## What

The compose stand seeded 60 days and the deployed stand 365. Both seed 730 now, so a developer and CI look at the same shape of data, and the range finally reaches back far enough for a year-over-year view.

## The trap this walks into on purpose

730 days is wider than any single request the analytics API will answer — `MAX_PERIOD_DAYS` is 400 and the check is `>=`, on both `/v1/metric-results` and `/v1/metric-drilldown`. `INFRA.md` already names this exact failure ("**730-day seed window**"): a window-derived request starts answering 400 and it reads like a broad data problem rather than a period that is simply too long.

The mitigation was already built — `tests/stand/api/analytics/query_window` returns the queryable tail of the seeded range, and the package docstring says every request wanting "a real period" should go through it. Four call sites still read `data_window` raw, and each would have gone red the moment the window grew:

- `test_drilldown.py` `_request_for` and `_period_value`
- `test_drilldown_rebuild.py` `_request` and `_period_value`

They go through the helper now, so the suite asks about the most recent 399 days whatever the seed window is. `INFRA.md`'s row is rewritten to match: the guard is no longer "keep the seed inside the API limit", it is "no test may read `data_window` raw".

A reconciliation has to pass the **same** clamped period to both sides — the evidence walk and the scalar it is checked against — or it compares two different periods and the disagreement looks like a data bug. Both pairs above are converted together for that reason.

## What the first CI run caught

`test_export_cells_match_the_page_and_hostile_values_stay_inert` failed with "no evidence title begins with `=`". The seed plants deliberately hostile commit titles for the export-escaping test, and it dealt them to the dev lead's **earliest** commits. Clamping the query to the tail of a 730-day window put them out of reach — the stand did contain every title, and the suite could not see one.

They are dealt to the most recent commits now, which the tail always covers. The dealing moved to after the row loop rather than inside it, and finds the message column by name, so the row tuple's shape stays the one thing the link-parity tests read positionally. Verified offline against the real column list: all six titles land on the final day of the window at both 60 and 730 days.

## Cost

**The paged walk grows on the compose lane.** `test_git_commit_drilldown_pages_and_reconciles` pages with `limit=1` and no budget, so its request count tracks the queried period — which on compose goes from 60 days of data to a clamped 399. Measured offline: 434 → 2,877 rows for that selection. At CI's latency that is a minute or two, not a lane-breaker, and it is bounded from here on: the clamp holds the walk at 399 days however wide the seed becomes. The deployed stand barely moves (365 → 399).

**The compose seed itself gets bigger** — 60 → 730 days is roughly 12× the rows for every developer bring-up and for the `Stand E2E` gate. Measured in this PR's own run: api-smoke 8m31s against a 45-minute ceiling, ui-journeys 5m39s and green. `SEED_DAYS` still overrides per run.


ktursunov:
Folded into #2498 — the four commits belong together: the full-suite lane is what makes the seed changes worth running, and the hostile-title fix only exists because the wider window exposed it. Closing this in favour of the combined PR.

PR #2502, opened 2026-08-13
feat(seed): generate wiki silver rows so the wiki metrics answer with data

ktursunov:
## What

The stand seeder has a generator per activity domain except wiki. So `silver.class_wiki_*` was always empty, `wiki_metric_evidence` built over nothing, and all four registered `wiki.*` metrics answered 200 with no rows — on every stand, compose and deployed alike, because both run the same seeder package.

Worth being precise about what was missing: the metrics themselves were never absent. They are compiled into the analytics registry and reconciled into MariaDB at service start, and their observation tables come from the committed DDL snapshot, so the schema probe reports them healthy and the catalogue advertises them. "Defined but has no data", not "does not exist". This adds the data.

## The generator

`generators/wiki.py`, modelled on `collab.py`, writing `class_wiki_pages`, `class_wiki_activity` and `class_wiki_engagement` for everyone with a team, scaled by a new per-team `outline` weight, the persona multiplier and the weekday multiplier. Volumes are sized against `docs/testing/REFERENCE-ORGS.md` §4.

One structural departure from `collab.py`: pages are planned before any row is emitted. The evidence model joins engagement onto `(tenant_id, source_id, page_id)` INNER and credits the **page's** author with the comments, so drawing comments independently would produce rows that silently vanish in the join. Comments and the activity row's `pages_created` both derive from that one planned list.

Registers the three relations in `RESET_TARGETS` (without which `truncate()` raises), adds the generator to the silver run, and adds `wiki_metric_observations` to the stand readiness gate — `dev-compose.sh` had it excluded with a note saying it was absent on purpose because no generator existed.

## The suite had built on the absence

Four places asserted the empty state, and each is now a positive assertion in the file's existing idiom:

- The person view asserted the Wiki card shows "No data"; it now asserts a populated card alongside the other four domains.
- The team view asserted "No metrics with peer data for this period"; same treatment.
- Every roster member was asserted to have an **unrecorded** "Page edits" cell; that label now joins the recorded-metric tuple.
- `test_supported_metric_with_no_evidence_returns_an_empty_page` hardcoded `wiki.pages_created` precisely *because* it had no evidence. It moves to `collab.files_engaged`, whose measure reads `class_collab_document_activity` — still unseeded. I verified against the deployed stand that it behaves identically to what it replaces: present, enabled, and 200 with zero rows. The same metric is added to `EXPORT_SHAPES`, whose docstring promises a capable-but-empty export case that `wiki.pages_created` no longer provides.

Four wiki entries in the drilldown reconciliation matrix have been short-circuiting on the empty branch; they start doing real sum/count reconciliation for the first time. Several gold dbt tests (unique grain, entity-id shape, non-negativity) have been vacuous zero-row passes and become live — the unique-grain one polices exactly the engagement join fan-out that is the riskiest part of this change.

## Coverage this gives up

`unrecorded_metric_cell(…, "Page edits")` was the only assertion that an unmeasured cell renders as "not recorded" rather than `0`, and no team-grid column is guaranteed unrecorded any more. Dropped rather than papered over, and recorded as a gap in the scenarios invariants.

## What is unverified

The seeder cannot run without a live ClickHouse, so this was checked offline against a fake client built from the real DDL column lists: 277 page / 840 activity / 377 engagement rows over the 60-day default for 24 people, no engagement row whose `(tenant_id, source_id, page_id)` is missing from pages, no date outside the window, distinct `unique_key` on all three tables, and every author carrying activity inside the SPA's default month. Column names were separately diffed against the DDL snapshot — every column written exists, so nothing is silently dropped.

Still unproven until this runs on a stand: that dbt builds the wiki gold relations non-empty from these rows, that identity resolves the seeded author emails for the wiki family, and that the domain card and the Page-edits column actually render populated. The `Stand E2E` gate on this PR brings up compose, seeds it and runs both lanes, so it proves all three before merge.

## Summary by CodeRabbit

* **New Features**
  * Added seeded Wiki activity, page, and engagement data for supported team views.
  * Wiki data is now included in readiness checks and dashboard coverage.
  * Team views now display Wiki page-edit metrics and peer data.

* **Bug Fixes**
  * Improved analytics coverage for metrics with no available evidence, including exports and empty results.

* **Tests**
  * Updated dashboard and drilldown validation to reflect populated Wiki data and empty-evidence scenarios.

### DIFF
--- a/.claude/skills/stand-scenarios/invariants.md
+++ b/.claude/skills/stand-scenarios/invariants.md
@@ -28,14 +28,17 @@ Enforced at: the API returns `null` rather than `0` for an unmeasured value
 renders `—` / "not recorded" / an explicit empty card.
 
 Partially covered: `test_the_personal_dashboard_renders_every_metric_domain`
-asserts the unseeded Wiki domain shows "No data" and
-"No metrics with data for this period.", and that populated KPI tiles are
-`not_to_have_text("—")`. `test_the_team_view_lists_every_report_the_roster_declares`
-asserts an unrecorded cell renders as unrecorded rather than as a number.
+asserts that populated KPI tiles are `not_to_have_text("—")`, and
+`test_the_team_view_lists_every_report_the_roster_declares` asserts that every
+member's cell renders for every column — recorded or an honest "not recorded" —
+without demanding a value.
 
-Gap worth a claim: the **API** half — that an unmeasured metric answers `null`
-and never `0`. The UI cannot prove it; a backend that returned `0` and a
-frontend that rendered `—` for falsy would look identical on screen.
+Gaps worth a claim: every metric domain is seeded now, so no view on this stand
+asserts the empty state itself — that a domain with no data renders its explicit
+empty card, and that an unmeasured cell reads "not recorded" rather than `0`.
+And the **API** half: that an unmeasured metric answers `null` and never `0`.
+The UI cannot prove that one at all; a backend that returned `0` and a frontend
+that rendered `—` for falsy would look identical on screen.
 
 ## R2 · Confidence and limitations travel with every conclusion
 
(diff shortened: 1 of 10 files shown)


ktursunov:
Folded into #2498 — the four commits belong together: the full-suite lane is what makes the seed changes worth running, and the hostile-title fix only exists because the wider window exposed it. Closing this in favour of the combined PR.

PR #2508, opened 2026-08-13
feat(frontend): Reports — build and export a custom metric report

### LINKED ISSUE #1570: PRESENT 12: Reporting: METRICS ability to export selected or all metrics for a period to CSV
Overview:
The Reporting feature enables users to export data for selected metrics over a specified time period. The data can be aggregated based on the granularity selected by the user: yearly, quarterly, monthly, or weekly.
The primary export format is CSV. The API should also support JSON as an alternative response format.

Planing:
Scope:
Assignee:
Estimation:

**Planning:**
Scope: -
Assignee: Sergey Mozhaev, Aleksandr Barkhatov, Roman Mitasov
Estimation (days): 5
Estimation QA (days): 2

dzarlax:
Closes #1570.

Reports → Report builder: pick any set of metrics and a granularity, preview the table, download it as CSV or Excel. Design discussed on #1570; this is the frontend-only build it concluded with, and **nothing here is blocked on the backend**.

## What a reader does

Picks metrics, picks daily / weekly / monthly / quarterly / yearly, presses **Build report**, sees the table it is about to export, downloads it. People come from the topbar's scope and the period from the topbar's period, so the report is about whatever the rest of the portal is already about.

One row per person per bucket, org attributes repeated on every row so the file can be pivoted without editing first:

| Person | Email | Division | Department | Job title | Manager | Manager email | Status | Period | From | To | Commits |
|---|---|---|---|---|---|---|---|---|---|---|---|
| Jane Doe | jane.doe@example.com | Platform | Core | Engineer | Sam Smith | sam.smith@example.com | Active | 2026-05 | 2026-05-13 | 2026-05-31 | 128 |

`From`/`To` are the bucket's real span **clipped to the requested period** — a report starting mid-May shows its first month starting on the 13th, so nobody compares a part-month with a whole one.

## Decisions worth reviewing

**Granularity is asked of the server where it has one.** Day, week and month exist there, so the server computes each bucket and every metric can be reported — a ratio is that bucket's own ratio. Quarter and year do not exist, so they are months added together, and only there does "may this be added" arise. The trap that makes the question necessary is `distinct_count`: formatted as an integer, reads as a counter, and someone active in two months is distinct in each.

**Which metrics may be added is asked, not assumed.** `computation` rides on results rather than on the definition listing, so a one-day probe over the catalogue collects it. That keeps the answer true for the installation being looked at; a list of metric keys in the client would drift the first time a metric is added, silently. Publishing `computation` on the listing would remove the probe — worth doing, not required.

**Nothing is hidden from the picker.** A metric that cannot go in the report is listed with the reason: not computing here, never observed, or not something a quarter can add up. On an install that does not ingest a source this removes whole families, and a family that simply vanishes leaves the reader unable to tell "not measured here" from "I misremembered the name".

**Requests split along two axes.** Metrics because the endpoint refuses more than fifty; people because of the projected-row limit. That limit is checked per view and never sees the metric count, so satisfying it alone would let one response carry an unbounded amount — the split is sized on the product of metrics, people and buckets instead, which also makes each request comparable in weight and the progress honest.

**Failure yields nothing.** A run resolves only when every batch has arrived; one that fails or is cancelled produces no file, because a file missing a few batches reads as a complete one once it is open. Same reason a built report is stamped with what produced it and dropped when any of it changes: a table built monthly must not still be downloadable after the reader switches to yearly.

## Not in scope

- **Rows about teams or org units.** The results API knows one entity type, `person`. Aggregating people into a unit is a backend capability, not an export option.
- **Ratios and medians at quarter or year.** They need `Bucket::Quarter` and `Bucket::Year` computed where the data is. Nothing here blocks adding them later.

## Verification

Driven end to end in a browser against a populated instance: the picker's families and disabled reasons, building, the progress count, the preview dialog, both downloads, the clipping of `From`/`To` at both ends of the period, and that switching granularity withdraws a stale report.

Unit tests cover the parts where a wrong answer would be quiet: rollup arithmetic and the empty-versus-zero distinction, the additivity rule per computation, the request split against both caps, the clipped spans, batch merging across chunks, and the CSV formula guard. Full suite green (1143), typecheck and lint clean.

**One gap stated plainly:** the tooltip carrying a disabled metric's reason did not open on hover in testing — the label wraps a disabled control, which the tooltip primitive appears not to treat as a trigger. A native `title` carries the same text as a fallback, so the reason is always reachable, but it renders as a system tooltip rather than in the app's style. Worth a follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added a report builder for selecting metrics, people, date ranges, and daily, weekly, monthly, quarterly, or yearly views.
  * Added report previews with progress tracking and cancellation.
  * Added CSV and Excel downloads with formatted headers and spreadsheet-safe values.
  * Added explanations when metrics are unavailable or incompatible with selected settings.
  * Reports are now available in the navigation.
* **Bug Fixes**
  * Improved handling of missing data and report results across time periods.

### DIFF
--- a/src/frontend/src/api/metric-results-client.ts
+++ b/src/frontend/src/api/metric-results-client.ts
@@ -181,7 +181,8 @@ export interface MetricResultsResponse {
 }
 
 export async function queryMetricResults(
-  body: MetricResultsRequest
+  body: MetricResultsRequest,
+  signal?: AbortSignal
 ): Promise<MetricResultsResponse> {
   // Refuse a request the backend is guaranteed to reject (400 invalid_argument,
   // "entity.ids must not be empty"). Callers are expected to keep the query
@@ -196,6 +197,7 @@ export async function queryMetricResults(
     method: "POST",
     headers: { "Content-Type": "application/json" },
     body: JSON.stringify(body),
+    signal,
   });
   if (!res.ok) {
     const errorBody = await res.json().catch(() => null);
--- a/src/frontend/src/components/portal/report-builder-view.test.tsx
+++ b/src/frontend/src/components/portal/report-builder-view.test.tsx
@@ -165,6 +165,28 @@ describe("ReportBuilderView", () => {
     );
   });
 
+  it("withdraws the file when the scope becomes a different roster of the same size", async () => {
+    // Counting people would call these the same report. They are not, and the
+    // file carries the names of whoever was in scope when it was built.
+    const user = userEvent.setup();
+    const { rerender } = render(<ReportBuilderView />);
+    await user.click(box("Commits"));
+    await user.click(screen.getByRole("button", { name: "Build report" }));
+    await screen.findByRole("dialog");
+    await closeDialog(user);
+    await screen.findByRole("button", { name: /rows ·/ });
+
+    mocks.scope = {
+      ...mocks.scope,
+      pivot: { ...person, person_id: "P2", display_name: "Sam Smith" },
+      roster: [{ person_id: "P2" }],
+    };
+    rerender(<ReportBuilderView />);
+    await waitFor(() =>
+      expect(screen.queryByRole("button", { name: /rows ·/ })).toBeNull(),
+    );
+  });
+
   it("downloads nothing when the run fails, and says so", async () => {
     mocks.run.mockRejectedValueOnce(new Error("network"));
     const user = userEvent.setup();
--- a/src/frontend/src/components/portal/report-builder-view.test.tsx
+++ b/src/frontend/src/components/portal/report-builder-view.test.tsx
@@ -1,0 +1,180 @@
+// @vitest-environment jsdom
+/**
+ * The report builder's promises to the reader: never offer a metric the file
+ * cannot honestly contain, never offer a file built from something other than
+ * what is on screen, and never hand back a partial one.
+ */
+import { render, screen, waitFor } from "@testing-library/react";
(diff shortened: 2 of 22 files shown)


hello1101n (approved):

dzarlax:
Four applied in 240546d, two were already in.

**Unknown computation counted as additive.** The worst of the six, and worse than described: I had also written a test asserting that exact behaviour, so the rule was locked in rather than merely missed. An unknown computation is now refused wherever buckets are added, and the test asserts the refusal plus the fact that finer granularities are unaffected — nothing is summed there.

**The recipe counted people instead of naming them.** Two different rosters of equal size read as the same report, so a table built for one scope stayed downloadable under another — and the file carries the names of whoever was in scope when it was built. Now keyed on the person ids, with a test that swaps the scope for a same-size roster and expects the offer to be withdrawn.

**Non-finite numbers in XLSX.** Agreed, and the inconsistency was the giveaway: the CSV writer already refused them and the XLSX one did not. Both now drop them, with a test.

**The probe's cache key counted metrics instead of naming them.** Same shape of bug as the recipe one. Keyed on the sorted key set.

The two coverage comments — the matrix writers and the report helpers — arrived with the preceding commit, before this review round: `matrix-download.test.ts`, `report-catalogue.test.ts` and the additions to `reports.test.ts`.

Full suite green (1172), typecheck and lint clean.

aleksdotbar (approved):

PR #2514, opened 2026-08-13
Reach evidence from distribution cards, and make the evidence table navigable

dzarlax:
## What

Two changes to the supporting-data (evidence) surface.

**Distributions can be drilled.** A distribution card sat in the same grid as
the summary card and the breakdown, took the same props, and was the only one
of the three with no way into the records behind it. It now carries the same
actions menu. The menu gates itself on the metric declaring a drilldown, so
cards for metrics without one look exactly as they did.

**The evidence table can be navigated.** It was a plain paginated list:
sorting and finding a record were both impossible once a result ran past a
screenful. Now:

- column headers sort, cycling ascending -> descending -> unsorted, with
  `aria-sort` on each header
- a search box filters across every column, matching the text as displayed
- the header states how many records are in view (`12 of 340 records`)

Searching or sorting over a partly-loaded result would answer for whichever
pages had been scrolled past, which is worse than not offering it — so
narrowing pulls the remaining pages in. The table's own scroll-triggered
paging cannot do this: a search that hides most rows also stops the scroll
that would load more. While that runs the counts read "so far", and a failed
page stops the pulling rather than retrying in a loop.

## Not included

No per-column totals. Whether a column may be summed depends on the metric's
computation, which rides on metric *results* and is not available where the
dialog is opened from; a sum over a column of medians or ratios would be a
number nobody should read. Worth doing once that is threaded through.

## Verification

- `pnpm typecheck`, `eslint`, `vitest run --project unit` (1197 passing)
- Exercised in a browser against a running instance: the menu appears on
  distribution cards for drillable metrics and is absent for the rest;
  sorting, search, the record count and the no-match state all behave as
  described.

## Summary by CodeRabbit

* **New Features**
  * Added search and match counts to metric evidence tables.
  * Added sortable evidence columns with ascending, descending, and reset states.
  * Evidence searches can automatically load additional records as needed.
  * Added clear-search messaging when no records match.
  * Added supporting-data actions to histogram metrics, including empty states.
* **Accessibility**
  * Added sort-state announcements and improved numeric-column alignment.

### DIFF
--- a/src/frontend/src/components/metric-evidence-dialog.test.tsx
+++ b/src/frontend/src/components/metric-evidence-dialog.test.tsx
@@ -405,6 +405,60 @@ describe("MetricEvidenceDialog", () => {
       act(() => onSortChange(key));
     }
 
+    it("does not call a search empty while pages are still coming", async () => {
+      const user = userEvent.setup();
+      renderDialog({ fetchNextPage: vi.fn(), hasNextPage: true });
+
+      await user.type(
+        screen.getByRole("searchbox", { name: "Search records" }),
+        "nothing"
+      );
+      expect(
+        screen.getByText("Nothing matched yet — still loading the rest")
+      ).toBeInTheDocument();
+      expect(screen.getByText("0 of 3 records so far")).toBeInTheDocument();
+      expect(
+        screen.queryByText("No records match this search")
+      ).not.toBeInTheDocument();
+    });
+
+    it("says the rest could not be loaded rather than claiming no match", async () => {
+      const user = userEvent.setup();
+      const fetchNextPage = vi.fn();
+      renderDialog({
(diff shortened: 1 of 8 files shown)


dzarlax:
Good catch on the no-match state — fixed in 7abdd81e3.

The branch replaced the table, and with it the retry the table owns, so a
failed page could leave "No records match this search" on screen while
matching rows sat in a page nobody had loaded. It now distinguishes three
cases: still loading, could not load the rest (with a retry), and genuinely
nothing matched. The record count keeps its "so far" after a failure too —
it had been dropping the qualifier once loading stopped, which made a short
total read as final.

Three regression tests added, one per case.

ktursunov (approved):

Issue #2479, opened 2026-08-13
AI cost: Claude Team


hello1101n:
**Expected result:** 
1. It should be possible on V1 instance see usage and actual costs of Claude teams subscription.
2. It should be possible on insight dev see usage and actual costs of Claude teams subscription.

**Action_items**:
1. Provide list what claude returns. (to Nikita)
2. What VZ should do to integrate with claude connector? (Do they have it already? Ask Roman Mitasov)
3. Communicate with Vitaly about updates in claude.
4. Invoice - should be configurable.
5. Rename (potential token costs) (usage)
6. Based on claude list - understand what metrics do we want.
7. Understand what and how can we show?
8. Implement frontend
9. Deploy to VZ connector to review it.
10. Review corner cases. (claude app) (Potentially deploy to VZ to review claude) (Web issues?)

------------------
Umbrella for release 08-25 item "AI cost: Claude team". Scope decided in #2425 §2: **subscription + overusage**.

## What we want to see

- **Extra usage cost** — the money the vendor billed a seat on top of its fee, per person / team / month, together with how close that seat is to the ceiling that blocks further spend.
- **Per-seat cost** — the price of the seat itself, taken from the vendor invoice, and the underused seats it exposes.

Both readable side by side, never summed into one figure.

## What "real cost" means here

- **Overusage** — what the vendor bills a seat once the usage included in its fee is exhausted. `used_credits` is that money; `monthly_credit_limit` is an enforced ceiling on spend, not the included allowance, so `max(0, used − limit)` is not overage. Keys: `ai.extra_usage_cost`, `ai.extra_usage_utilisation` (proximity to being blocked).
- **Subscription** — the seat price, obtainable only from vendor invoices. Keys: `ai.seat_cost`, `ai.seat_underuse`. Seat waste in money cannot be computed without it.
- The two are never summed — per #1607: token-metered and per-seat stay separable, not-tracked instead of `0` where a source is missing, no `$/PR` as fact.
- **`ai.cost` is neither** — it is potential cost, all consumption at vendor rates. `ai.extra_usage_cost` is the real overspend. The two must not be mixed.

## Children

- #2428 — seat economics through the metrics API (extra usage) · PR #2431
- #2429 — Claude Team vendor invoices (per-seat cost) · PR #2432, stacked on #2431
[...]

Issue #2361, opened 2026-08-10
Metric drilldown intermittently returns 500 while the stand suite runs


ktursunov:
## Summary
During a full run of the stand analytics suite, `POST /v1/metric-drilldown` intermittently answers 500, so drilldown requests that succeed in isolation fail when the service is busy.

## Steps to Reproduce
1. Run the stand analytics API suite (~80 tests, a rapid mix of `/v1/metric-results` and `/v1/metric-drilldown` calls for several persons) against a seeded compose stand.
2. Two to three drilldown tests fail with 500; which ones varies between runs.
3. Re-run the failed tests alone, or after a ~30-second pause — they pass.

**Expected:** 200 on every well-formed drilldown request, regardless of concurrent load.
**Actual:** intermittent 500 `application/problem+json` on `/v1/metric-drilldown`.

## Additional information
- Server log at the moment of each 500, verbatim:
  ```
  analytics::domain::person_visibility: visibility check failed
  error=identity service returned 429 Too Many Requests
  ```
  followed by `canonical error response (server) status=500 … description="metric access could not be authorized"`.
- Response body: `{"title":"Internal","status":500,"detail":"An internal error occurred. Please retry later.","instance":"/v1/metric-drilldown"}`.
- Boundary observations from the same stand: 300 sequential single-person paged drilldown requests (`limit=1`, ~50 req/s) all return 200; 48 requests spread across 6 persons all return 200. Only the full-suite mix has triggered it (10 such log lines across three suite runs in one morning).

## Notes
related to #2109

ktursunov:
Closing as no longer reproducing.

The three drilldown tests this issue was tracking were marked non-strict xfail on
2026-08-11 so the suite would record the flake instead of gating on it. Every
compose-stand `api-smoke` run recorded since then has **xpassed all three** —
the 500 has not occurred once:

| run | date | api-smoke summary |
| --- | --- | --- |
| [31457748699](https://github.com/constructorfabric/insight/actions/runs/31457748699) | 2026-08-11 | 9 failed, 205 passed, 2 xfailed, **3 xpassed**, 61 errors |
| [31564085386](https://github.com/constructorfabric/insight/actions/runs/31564085386) | 2026-08-12 | 279 passed, 2 xfailed, **3 xpassed** |
| [31668000915](https://github.com/constructorfabric/insight/actions/runs/31668000915) | 2026-08-13 | 300 passed, 2 xfailed, **3 xpassed** |
[...]

PR #2524, opened 2026-08-14
fix(analytics): name the billing month and the ceiling of a seat-month row

### LINKED ISSUE #2523: Drilldown hides the billing month and the ceiling of a seat-month row
Part of #2479. Found while testing #2431 against a populated instance.

`insight.ai_cost_metric_evidence` carries a `details` map holding `billing_month`, `ceiling_usd` and `ceiling_set` — put there deliberately by #2428. None of it reaches `POST /v1/metric-drilldown`. The evidence dialog for `ai.extra_usage_cost` is two columns:

```
COLUMNS: [{"key":"date"},{"key":"value"}]
ROW0:    {"values":{"date":"2026-07-30","value":9.76}}
```

## Where it comes from

`evidence_presentation()` in `src/backend/services/analytics/src/domain/metric_drilldown/presentation.rs` decides which keys of `details` become columns. Its match names `git`, `task` and `wiki`; `ai_cost` falls through to the catch-all, whose `detail_keys` is empty. Nothing ties "a source was added" to "its presentation was described", and no test covers the pairing — the drilldown sweep asserts only that a sum metric projects `value` and a ratio projects both sides.

## Why it matters for this metric in particular

A seat-month row is dated at the day its snapshot was last READ, not at the month it bills for. That is deliberate: a row dated the first of the month falls outside short rolling windows and the current month would vanish from the dashboard. The consequence is that the date alone cannot place the amount — on a populated instance the July row is dated 2026-07-30 and the August one 2026-08-13, and a month read after it closed reads as the next month's money.

The ceiling is invisible for the same reason, so the amount has nothing to be judged against, and a seat that was never given a ceiling cannot be told from one whose ceiling failed to load. The ceiling is also what decides whether `ai.extra_usage_utilisation` answers at all: no ceiling, no ratio.

## Expected

The evidence dialog for `ai.extra_usage_cost` names the billing month and the ceiling. `ai.extra_usage_utilisation` is a ratio and keeps its numerator/denominator presentation, where per-row details do not apply.

Gregory91G:
Closes #2523. Part of #2479. Found while testing #2431 against a populated instance.

The gold model writes `billing_month`, `ceiling_usd` and `ceiling_set` into the evidence row's `details`; `evidence_presentation` named `git`, `task` and `wiki` only, so `ai_cost` fell to the catch-all and none of it reached the response. The evidence dialog for `ai.extra_usage_cost` was a date and an amount — and the date is the day the snapshot was READ, not the month it bills for, so nothing in the row placed the money. On a populated instance the July row is dated 2026-07-30 and the August one 2026-08-13.

## What lands

- **`presentation.rs`** — an `ai_cost` arm projecting `billing_month` and `ceiling_usd`, with the two column labels. `ceiling_set` stays out: the gold model writes an empty `ceiling_usd` exactly when no ceiling exists, so a blank cell already says it, and a second boolean column would be noise.
- **`tests/stand/api/analytics/test_ai_cost.py`** — five properties of the seat-month metrics that nothing asserted today: honest-NULL for a seat with no ceiling (a ratio, never a `0`), a window holding the read day returning the month in full while the same month cut off the day before returns nothing, the usage-priced and the billed key answering side by side, a seat-tier breakdown that adds up, and the row naming its billing month and ceiling.

Every expectation is derived from what the stand serves — the read day and the billing month come out of the evidence itself — so the tests survive a seed anchored to another date.

## Test plan

Run against a stand brought up from this branch (`test-stand up --build`), with the analytics binary rebuilt from it:

- [x] `cargo test -p analytics metric_drilldown` — 33 passed, 0 failed.
- [x] `cargo clippy -p analytics --all-targets -- -D warnings` — clean.
- [x] `cargo fmt --all --check` — clean.
- [x] The five new tests — 5 passed. The fifth fails without the fix, which is how the gap was found.
- [x] `test-stand test --ignore=tests/stand/ui` — 307 passed, 1 skipped, 2 xfailed, 3 xpassed, 0 failed. Same xfail/xpass set as before the change (#2360, #2361).
- [x] `pre-commit run --files` on both paths — `ruff check`, `ruff format` and the file-hygiene hooks pass.

The UI lane was not run locally; nothing here changes a frontend surface, and the two new columns are additive to a payload the dialog renders generically.


Gregory91G:
@coderabbitai review

Gregory91G:
@coderabbitai review

hello1101n (approved):

PR #2526, opened 2026-08-14
fix(claude-team): read seat spend before the billing month closes

### LINKED ISSUE #2525: Claude Team seat spend loses the tail of every billing month
Part of #2479. Found by comparing July against the reference implementation.

`/api/organizations/{org}/overage_spend_limits` reports the billing month **in progress** and carries no history: once the month rolls over, the endpoint answers for the new one and the old month can never be re-read. A month's figure is therefore frozen at whatever the last sync inside that month saw, and everything spent after it is lost for good.

`claude-team-default-sync` runs `0 4 * * *`. Even when every run succeeds, the last read of a month lands at 04:0x on its last day, so the final ~20 hours are never captured. A failed or missing run on the last day widens that to two days or more.

## Measured

`silver.class_ai_overage` on insight-dev, last read per billing month, and the part of the month that follows it:

| Billing month | Last read (UTC) | Hours of the month never read |
|---|---|---|
| 2026-06 | 2026-06-24 04:01:49 | 164 |
| 2026-07 | 2026-07-30 04:02:00 | 44 |

Against the reference implementation's export for July (145 seats):

- our total $12,780.60 against theirs $13,415.12 — **$634.52 short, 4.7%**;
- 44 of the 57 seats with a non-zero amount agree to the cent ($5,493.09);
- 13 disagree, **every one of them in the same direction** — ours is never larger;
- three of the 13 read $0.00 on our side, which is what a seat that only started spending in the unread tail looks like.

The transformation itself is exact: joining bronze to silver over the same month gives 0 mismatches on both the amount and the ceiling, so nothing is being miscomputed — the rows simply stop earlier than the month does.

## What would fix it

A read as late as the month allows, and an alarm when it does not happen:

- schedule the last read of each month close to 23:59 UTC on its final day, or read often enough near the boundary that the tail is minutes rather than hours;
- a run at 00:0x on the 1st does NOT help — by then the endpoint answers for the new month;
- a missing or failed sync on the last day of a month is unrecoverable, so it deserves an alert rather than the silent shortfall it produces today.

Note for whoever picks this up: the fix is about the month boundary, not about raising frequency everywhere. Hourly polling all month long would not improve a figure that only matters at its close.

Gregory91G:
Closes #2525. Part of #2479.

`/api/organizations/{org}/overage_spend_limits` reports the billing month **in progress** and carries no history. A month's spend is therefore frozen at the last read that happened inside it, and everything after that read is lost for good — the next sync answers for the new month and cannot go back.

Reading at 04:00 UTC dropped the last twenty hours of every month by construction. A failed or missing run on the last day dropped a day more.

## Measured against the reference implementation

July, 145 seats, compared per address:

| | |
|---|---|
| Our total | $12,780.60 |
| Reference | $13,415.12 |
| Short by | **$634.52 — 4.7%** |
| Non-zero seats agreeing to the cent | 44 of 57 |
| Disagreeing | 13, **every one of them short**, never over |

Our final read of July landed 2026-07-30 04:02 UTC — **44 hours unread**. June's landed 2026-06-24, leaving 164 hours.

The transformation is exact: joining bronze to silver over the same month gives zero mismatches on both the amount and the ceiling. Nothing is miscomputed; the rows simply stop before the month does.

## What lands

- **`descriptor.yaml`** — `50 23 * * *` instead of `0 4 * * *`, leaving ten minutes of a month unread instead of twenty hours. Still off-peak: the other connectors occupy 02:00-06:00 and 11:00, and the customer's proxy runs a single browser instance. Both streams tolerate the move — the code-metrics stream walks a trailing window and re-reads it, so it is unaffected by the hour.
- **`assert_ai_overage_read_before_the_month_closed`** — a warning when the month that just closed was last read more than six hours before it ended. The schedule puts the final read ten minutes before the boundary; one failed run there makes it a day. Only the just-closed month is judged: an open month is still being read, and an older one can be neither re-read nor acted on.

No descriptor version bump: the reconciler applies the CronWorkflow unconditionally (`kubectl apply`, no-op when unchanged — `lib/reconcile.sh`), and ADR-0015's semver governs the connector's spec and data shape, neither of which moves here.

## Test plan

- [x] `dbt parse` — clean; the new test is in the manifest as `test.ingestion.assert_ai_overage_read_before_the_month_closed`.
- [x] The check's SQL run against a populated instance: returns exactly one row — July, last read 2026-07-30 04:02, `hours_unread = 44`. It catches the defect that prompted it.
- [x] The same SQL after the current month closes returns nothing once a read lands near the boundary, which is what the six-hour line is for.
- [x] `pre-commit run --files` on both paths — yamlfmt and the file-hygiene hooks pass.

An earlier draft of the check aliased the closed month to `period_month` in its SELECT; the alias shadowed the column, the WHERE compared the alias with itself, and every month collapsed into one group whose latest read belonged to the open month — inverting the result to a silent pass. The comment in the file records it so the shape is not reintroduced.


Gregory91G:
@coderabbitai review

Gregory91G:
@coderabbitai review

hello1101n:
failed checks

hello1101n (approved):

PR #2529, opened 2026-08-14
fix(frontend): drop the shadcn CLI from dependencies and pin brace-expansion

### LINKED ISSUE #2534: frontend: the shadcn CLI sits in production dependencies and carries 15 advisories with it
Nine high advisories sit in `src/frontend/pnpm-lock.yaml`, plus eight medium and four low. Almost all of them arrive through one manifest entry.

## Where they come from

`shadcn` is declared in **`dependencies`**. It is a scaffolding CLI that also ships one stylesheet, and in the production tree it pulls an MCP SDK, `hono`, `express-rate-limit` and `ajv` behind it:

```
js-yaml     4.3.0  <- cosmiconfig                       <- shadcn
fast-uri    3.1.4  <- ajv <- @modelcontextprotocol/sdk  <- shadcn
ip-address 10.2.0  <- express-rate-limit <- …/sdk       <- shadcn
hono, @hono/node-server, qs, body-parser, nanoid        <- shadcn
```

That accounts for 15 of the 20 findings. The rest are `brace-expansion` (three copies) and `uuid`, both behind `exceljs`, and `esbuild` behind vite.

Trivy scans production dependencies only for pnpm — the scan log says *"Suppressing dependencies for development and testing"* — so the classification of that one entry is what puts those packages in the report.

## Why it cannot simply be deleted

The package is a build input. `src/frontend/src/index.css` opens with:

```css
@import "shadcn/tailwind.css";
```

Deleting the entry fails the production build with `Can't resolve 'shadcn/tailwind.css' in '/app/src'`. A grep for `import … from "shadcn"` over the TypeScript sources finds nothing, so the package reads as unused until the build runs — lint and typecheck both pass without it.

## What is correct

Move it to **`devDependencies`**. The import is resolved at build time and nothing ships the package, so that is where it belongs on the merits; it also takes its subtree out of the Trivy report, which is the outcome originally wanted.

Then pin `brace-expansion` per major to the patched releases. Note that the pins must go in `package.json`'s existing `pnpm.overrides` — when that block exists, pnpm resolves from it and an `overrides:` block in `pnpm-workspace.yaml` is ignored without warning.

## Out of scope

`uuid` 8.3.2 behind `exceljs`: its fixed releases start at 11.1.1, three majors past what exceljs declares, so forcing it risks the Excel export path. Needs its own change with the export exercised.

`esbuild` 0.27.7 (low) comes from vite and moves when vite does.

## Acceptance

- `pnpm build` passes.
- `trivy fs` over the frontend reports no high findings; only the documented `uuid` medium remains.
- The lock's own `overrides:` block shows the `brace-expansion` pins actually applied.

Gregory91G:
Closes #2534.

Nine high advisories sit in `src/frontend/pnpm-lock.yaml`, plus eight medium and four low. `pnpm why` puts most of them behind a single entry.

## Where they come from

```
js-yaml     4.3.0  <- cosmiconfig            <- shadcn 4.7.0
fast-uri    3.1.4  <- ajv <- @modelcontextprotocol/sdk <- shadcn 4.7.0
ip-address 10.2.0  <- express-rate-limit <- @modelcontextprotocol/sdk <- shadcn 4.7.0
hono, @hono/node-server, qs, body-parser, nanoid                      <- shadcn 4.7.0

brace-expansion 1.1.16 / 2.1.2 / 5.0.8   <- exceljs 4.4.0
uuid            8.3.2                    <- exceljs 4.4.0
esbuild         0.27.7                   <- vite
```

`shadcn` is a scaffolding CLI, but it is also a build input: `src/index.css` opens with `@import "shadcn/tailwind.css"`, so the package has to be installed for the stylesheet to resolve. What it does not need to be is a *production* dependency — the import is resolved at build time and nothing ships it. Sitting in `dependencies`, it pulled an MCP SDK, an HTTP server and a rate limiter into the production tree.

Trivy scans production dependencies only for pnpm (*"Suppressing dependencies for development and testing"*), so removing the entry takes those packages out of the report as well as out of the tree.

## What this changes

- `shadcn` moved from `dependencies` to `devDependencies`.
- `brace-expansion` pinned per major in `package.json`'s existing `pnpm.overrides`, so each of the three copies in the tree moves to its fixed patch: `1.1.18`, `2.1.4`, `5.0.9`. `pnpm` reads overrides from the manifest when it has them, so a block in `pnpm-workspace.yaml` would be ignored.

The lock loses 1645 lines and resolves **701 packages instead of 894**.

## Deliberately left alone

`uuid` 8.3.2 behind `exceljs`. Its only fixed releases are 11.1.1 and above — three majors past what exceljs declares — so forcing it risks the Excel export path at runtime. That belongs in its own change, with the export exercised.

`esbuild` 0.27.7 (low) comes from vite and moves when vite does.

## Test plan

- [x] `pnpm why` traced every flagged package to its root before touching anything.
- [x] `pnpm build` passes. Removing `shadcn` outright does not: `Can't resolve 'shadcn/tailwind.css' in '/app/src'`.
- [x] The lock records the overrides that apply — `brace-expansion` resolves to `1.1.18`, `2.1.4` and `5.0.9`.
- [x] Lock regenerated with pnpm 10.33.0 on Node 24 and `pnpm install --frozen-lockfile` passes on the result.
- [x] `shadcn`, `fast-uri`, `ip-address` and `hono` are gone from the lock entirely; `js-yaml` survives only through a dev dependency.
- [ ] `Frontend — frontend`: lint, typecheck and both vitest projects — my local run exceeded its time budget, so this PR's own job is the check that matters. Nothing in the source references the removed package, so the expectation is no change.
- [x] `trivy fs` over the built tree: **9 high, 8 medium and 3 low on `main` become one medium** — the `uuid` behind exceljs left alone above.
- [ ] `trivy-fs` on this PR: the authoritative count, on CI rather than a laptop.

## Summary by CodeRabbit

* **Chores**
  * Updated dependency configuration to improve security coverage.
  * Moved a build-time tool to development-only dependencies.
  * Expanded version safeguards for a transitive package.
  * Documented dependency pins related to security advisories.


aleksdotbar (changes requested):
You can't just drop shadcn from dependencies

"It is imported nowhere in the source" is a false claim. It is imported in index.css

aleksdotbar (approved):

PR #2539, opened 2026-08-14
Name the evidence dialog for the metric it shows

dzarlax:
Fixes a regression I introduced in #2531, caught by the deployed-stand
journeys on the next PR.

## What broke

#2531 gave every entry into the evidence dialog the metric picker, which means
every entry now passes more than one target. The multi-target branch names the
dialog from a `title` the caller supplies — and only the timeseries block
supplies one. Opening from a card, a chart point or a roster cell fell through
to the placeholder `"Metric evidence"`, where the name had been the metric's
own label.

Two consequences:

- a screen reader announces a generic placeholder instead of the metric
- three committed browser journeys locate the dialog by that name and fail:
  `test_collaboration_card_menu_opens_and_exports_supporting_data`,
  `test_git_commit_bucket_opens_and_exports_supporting_data`,
  `test_team_heatmap_cell_opens_that_members_supporting_data`

#2531's own checks were green because the stand was still serving the previous
frontend when they ran; the failure surfaced only after it caught up.

## The fix

The dialog falls back to the active target's label rather than a placeholder.
A caller that names the whole set still wins — the timeseries block's joined
title is asserted by `test_timeseries_block_evidence.py`, so naming every
dialog after its active metric would have broken that instead.

## Verification

- `pnpm typecheck`, `eslint`, `vitest run --project unit` (1334 passing)
- Two unit tests added, one per branch of the fallback. Mutation-checked:
  restoring the placeholder fails the second one.
- The three stand journeys locate `get_by_role("dialog", name=<metric>)`, which
  this restores; the stand lanes on this PR are the real proof.


dzarlax:
Superseded by #2540, which carries this commit (authorship preserved) together
with the CI change that lets a frontend-only diff prove itself. Merged there.

PR #2542, opened 2026-08-14
ci: probe the --build-backend lane (do not merge)

ktursunov:
Throwaway verification branch for #2540, not for merge.

It carries #2540's two CI commits plus a comment-only touch under `src/backend/`, so `changes` answers `backend=true frontend=false` and both lanes take `--build-backend` with the frontend still pinned — the one flag combination the symmetric wiring introduces that no run has taken yet.

What to read in the run:

- `changes` logs `relevant=true backend=true frontend=false`
- both lanes log `the backend is compiled from this tree, not pulled` and pin `insight-frontend:<appVersion>`, with no `Pinning the backend to published images` line
- `api-smoke` passes against the compiled backend

`ui-journeys` is expected to FAIL here on the three evidence-dialog journeys: this branch has no frontend fix, so it runs main's published SPA, which is the very regression #2540 exists to make testable. That failure is the control, not a defect in the lane.

Closed once read.

### DIFF
--- a/.github/workflows/e2e-stand.yml
+++ b/.github/workflows/e2e-stand.yml
@@ -6,21 +6,13 @@ name: E2E — Compose stand
 # exercises the in-process ingestion rig and now carries one blocking gate
 # (metric coverage) — that workflow is untouched by this one.
 #
-# By default nothing here builds the product. `test-stand up` pulls the frontend
-# and the four backend services, each pinned to its own chart's appVersion;
-# compiling the backend cost ~26 min per job.
-#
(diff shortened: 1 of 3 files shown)


ktursunov:
Read: `changes` answered `relevant=true backend=true frontend=false`, both lanes logged `the backend is compiled from this tree, not pulled` and compiled all three Rust binaries while pinning the frontend, and api-smoke passed 329 against that build. ui-journeys failed on exactly the three evidence-dialog journeys — the control, since this branch pins the pre-fix SPA. The --build-backend lane is verified; closing.

PR #2554, opened 2026-08-14
Plain words across the portal

dzarlax:
A copy pass over the portal, so the screens read to someone who has not been
in the room where the words were chosen.

Three habits ran through the interface, and each one is fixed the same way
everywhere:

1. **The product spoke about itself in the first person.** "What we can see",
   "how much of the work reaches us", "nothing reaches us here". A reader
   cannot tell who "we" are, and hears a claim about people where the subject
   is which sources have data.
2. **Internal vocabulary reached the screen.** "semantic layer", "bullet-only
   direction", "not wired yet", "exceeds one request", "a summable headline
   metric and ≥2 units of ≥4 people", "IQR", "on-par", "bus factor".
3. **Two words for one thing.** "part of work" on one screen, "section" on
   another, for the same five things; "members" and "people" side by side.

## By screen

**Data coverage** (was "What we can see")

| before | after |
| --- | --- |
| Overview · What we can see | Overview · Data coverage |
| how much of the work reaches us, by part and by person | which sections have data, and for how many people |
| people are seen in fewer than half of their work — everything else this product says about them rests on that fraction | people have data for fewer than half the sections. Everything shown about them comes from that part only |
| By part of work / By person · parts we can see | By section / By person · sections with data |
| nothing reaches us here | no data |

**Attention needed**

| before | after |
| --- | --- |
| N of M people need a look — most often … | N of M people stand out this period — most often … |
| unusually high · team median $3,562 | well above the team median of $3,562 |
| down 40% vs last period | down 40% from last period |
| fell this period / ongoing / no earlier period | dropped this period / still low / nothing to compare |
| no commits (team median 10) | no commits — the team median is 10 |

**Comparisons on a person's page**

| before | after |
| --- | --- |
| Top issue / Top win | Furthest behind / Furthest ahead |
| No peer data | No comparison |
| N on-par metrics | N metrics near the median |
| No critical issues this period | Nothing stands out as low this period |

**Directions** — titles now say what the chart shows: "Flow health · median" →
"Typical time, from start to done"; "PR cycle-time distribution (events)" →
"How long pull requests stayed open"; "Bus factor · top 10% of contributors" →
"How much of the work sits with the busiest tenth".

**States a reader meets when something is missing** — "Not in the semantic
layer yet — bullet-only direction" → "Not available yet for this direction";
"Admin surface — not wired yet" → "Not built yet"; "Trend needs a narrower
period or a smaller scope — this many people over this window exceeds one
request" → "Too many people over too long a period to chart at once. Pick a
shorter period or a smaller group."; "No comparable units for this lens at
this slice (needs a summable headline metric and ≥2 units of ≥4 people)" →
"Nothing to compare here: this needs at least two groups of four or more
people, and a metric that can be added up."

## Two things that were not only wording

- A missing section for a person was labelled **"no connector"**, while the
  comment beside the state that produces it warns that a disabled metric and a
  broken schema land there too, and that sending someone to plumb a live
  connector is the wrong way to be wrong. It now reads "no data from".
- The attention panel **printed its count twice** — as a chip and as the
  opening words of the sentence next to it — and carried a "· rule-based"
  tail. The duplicate chip and the tail are gone.

## Verification

- `pnpm typecheck`, `eslint`, `vitest run --project unit` (1334 passing)
- Roughly twenty tests pinned the old wording and were updated with it.
- Read on a running instance: Data coverage, Attention needed and a Direction,
  each checked as rendered text rather than in source.

Metric names and descriptions come from the backend catalogue and are
deliberately untouched here; they are a separate pass.

## Summary by CodeRabbit

- **User Experience**
  - Refined portal terminology to use clearer, more consistent language such as “people,” “sections,” and “Data coverage.”
  - Improved empty, unavailable, and comparison states with clearer guidance, including “No comparison” and “No data source is connected.”
  - Updated attention, peer-status, metric, and drilldown labels for easier interpretation.
  - Renamed navigation and report labels, and clarified roadmap and “Coming soon” messaging.
- **Accessibility**
  - Improved the accessible description for UTC date handling.


dzarlax:
Went through the five review comments. Four are addressed in a5716f7; one is declined.

**One phrase per peer standing — taken.** The shared label table already reads `in_pack: "Near the median"`, so the three remaining "on par" strings were the odd ones out, not the new phrase. All three now follow the table: the row chip in `team-metric-group-card.tsx`, the supporting-metric line in `peer-story.tsx`, and the two doc comments that named the vocabulary. The value line that read `on par · <cohort> median: N` folds into `near the <cohort> median: N` rather than carrying both words for one idea.

**A separate no-comparison state in the triage row — taken, and it was a real claim.** `rollupStatus` returned `in_pack` whenever nothing was behind or ahead, including when nothing about the person could be ranked at all, so the row said "near the median" with no median behind it. `TriageRow` now carries `rankable`, and the row reads "no comparison" at zero. `MembersOverview` already computed the counts it needs. Covered by a new test.

**Finish member → people — taken.** Both strings: the grid caption and the empty-roster line, which now reads "Nobody in this team yet."

**Cover the one-person summary branch — taken.** `attentionSummary([], 0, 1)` is asserted directly.

**Mention the metric count in the no-trend message — declined.** `pickTrendBucket` does weigh metrics, but the metric set is fixed by the section's configuration: there is no control on the screen that changes it. Naming it as a remedy would point the reader at a dial they cannot turn. The two the message already names, period and group size, are the ones they can.

hello1101n (approved):

PR #2591, opened 2026-08-17
fix(frontend): fit the settings controls inside the rail popover

### LINKED ISSUE #2590: The settings menu clips the planned-sections row and squashes the Focus segments
## Summary

The portal rail opens the settings menu in a 224px popover that hosts the sidebar footer. Two of its controls are wider than that and neither degrades: the planned-sections row wraps onto a second line the fixed-height row then clips, and the four Focus segments are pinned to equal widths narrower than their own labels, so the text runs over the segment dividers.

## Steps to Reproduce

1. Open the portal at 1440x900.
2. Click the settings button at the bottom of the icon rail.
3. Read the row under Portal, then the Focus segments below it.

**Expected:** every row shows its whole label on one line, and each Focus label sits inside its own segment.
**Actual:** "Show planned sections" breaks after "planned" and the second line is cut off at the row's bottom edge; "Critical" and "Rewards" touch the dividers on both sides, so the segments read as squashed.

Reproduces on a local dev build with mock data.

Settings menu: the planned-sections label wraps and is clipped, and the Focus labels overrun their segments

## Additional information

- The row is 216px wide with 8px padding, leaving 200px; its label, switch and the gaps between them measure 209px.
- `SidebarMenuButton` truncates `[&>span:last-child]`. These rows put the `Switch` after the label, so the label span is no longer the last child and nothing keeps it on one line.
- Each Focus segment gets a quarter of the row — a 35px content box — while "Critical" renders at 40.5px and "Rewards" at 49.5px. The segments have no overflow handling, so the text spills rather than shrinking the padding.
- Only the rail popover is affected. The same controls in the 256px sidebar have room for both.

hello1101n:
Fixes #2590.

The rail's settings popover was 224px while it mounts the sidebar footer unchanged. The planned-sections label wrapped and the fixed-height row clipped the second line; the Focus segments were pinned to equal widths narrower than their labels, so the text ran over the dividers.

- The popover takes the sidebar's own 16rem, the width its content was built for. The row now has 23px to spare, so a longer label or a translation does not put it back.
- Each Focus segment sizes to its own label and shares out what is left over, instead of taking a fixed quarter.
- A Storybook browser test renders the controls at the popover's width and fails when a label no longer fits.

| Before | After |
|---|---|
| Settings menu before | Settings menu after |


ktursunov (approved):

hello1101n:
After, with the popover at the sidebar's width — the label is back to "Show planned sections" and stays on one line:

Settings menu after the fix

The same menu in place, portal at 1440x900 with mock data:

Portal with the settings menu open beside the rail

ktursunov (approved):

PR #2597, opened 2026-08-17
test(stand): make the download journeys prove what the file holds

ktursunov:
`tests/stand/ui/` holds the browser tests that click through a deployed Insight
like a user would. Three of them open a dialog and export a file —
`test_git_output_timeseries`, `test_collaboration_card_evidence`,
`test_metric_evidence_drilldown` — and each checked only that the filename ended
in `.csv` or `.xlsx` and that the file was bigger than zero bytes. A file holding
the wrong period, a header and no rows, or zeros where the screen showed blanks
passes that too.

## What they assert now

| Before | Now |
| --- | --- |
| filename ends in `.csv` / `.xlsx` | same, and the file is opened and read back |
| file is larger than 0 bytes | CSV and XLSX must hold the same table as each other |
| — | the exported record count equals the row count the grid itself declares (`aria-rowcount`) |
| — | every row visible on screen equals the row exported for it |

No expected value is written into a test, so re-seeding the stand cannot break
them. Each comparison is between two things the stand produced independently:
what the SPA rendered, and what the export path serialized.

## How

One new helper, `tests/stand/ui/downloads.py`. It saves the download, reads
`.csv` with the standard library and `.xlsx` with `openpyxl` (new dependency,
locked), and puts every cell — CSV text, XLSX number, rendered DOM text — through
one normalizer. `29`, `29.0` and a rendered `+2,705` compare equal, while an empty
cell stays distinct from a zero. That last distinction is the point: a serializer
that writes `0` for a missing value passes anything that merely counts rows.

## Verification

All 17 tests in `tests/stand/ui` pass against a seeded compose stand. To confirm
the new assertions can fail, the reader was broken three ways on purpose — a
missing cell read as `0`, the last record dropped, every number shifted — and the
three tests failed each time.

## One call for the reviewer

The XLSX export of the Git output timeseries carries the `Total` and `Grand total`
rows the table shows; the CSV stops at the last bucket. The test compares the
bucket rows only, so it takes no position on which is right — but one of them
probably is.

### DIFF
--- a/tests/pyproject.toml
+++ b/tests/pyproject.toml
@@ -50,6 +50,11 @@ dependencies = [
     # MINT anything: the assertion is a request for a token, exchanged at the
     # deployed endpoint, exactly as a real service does it.
     "pyjwt[crypto]>=2.9",
+    # Reads the .xlsx half of every export a journey downloads. A workbook is a
+    # zip of XML, so a hand-rolled reader is possible and is a liability: the
+    # assertion would then depend on test code nobody reviews as carefully as
+    # the product. Read-only here — the suite writes no spreadsheets.
+    "openpyxl>=3.1",
 ]
 
 [dependency-groups]
--- a/tests/stand/ui/downloads.py
+++ b/tests/stand/ui/downloads.py
@@ -1,0 +1,133 @@
+"""Read back what the browser downloaded, and what the grid behind it showed.
+
+A download event and a byte count prove the transport worked, which was never
+the part in doubt. What a journey has to prove is that the file holds the table
+the user was looking at — and CSV and XLSX are written by different code
+(client-side for timeseries exports, `rust_xlsxwriter` beside the `csv` crate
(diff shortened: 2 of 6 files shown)


ktursunov:
Correcting my own note here: the CSV/XLSX footer difference is by design, not a defect.

`24fe8baf fix(metrics): omit CSV footer rows` removed the footers from the CSV deliberately, and `metric-timeseries-export.test.ts` locks both halves — "exports grouped CSV data without footer rows" asserts the CSV contains neither `Total` nor `Grand total`, while the workbook test asserts `A6`/`A7` hold them. The reasoning holds up: the workbook's grand total is a formatted composite string in a merged cell, which would corrupt the CSV's machine-readable shape.

So there is nothing to decide and nothing to file. What was worth fixing is this journey, which compared bucket rows only and therefore asserted neither half of that contract — #2608 asserts the split in both directions.

PR #2605, opened 2026-08-17
identity: let the console browse the roster and every account, a page at a time

mozhaev-dev:
Both search tabs of the identities console opened blank and refused to answer without
terms, so an operator could reach a person or an account only by guessing a value to
type. Both now list the full set with keyset pagination, and terms narrow that same
list. The review queue's binding read became one query rather than a per-batch
fan-out, and both halves of its join are bounded.

Touches identity-resolution (the two listings, the shared cursor, the evidence fold,
the binding repo) and the console frontend. No migration, no config.

```bash
cd src/backend && INTEGRATION_TESTS_MARIADB_URL=mysql://… cargo test -p identity-resolution
cd src/frontend && pnpm vitest run --project=unit
```

By hand on mocks (`VITE_ENABLE_MOCKS=true pnpm dev`) → Manage → Identities: both tabs
list before anything is typed, and the assign field inside a case dialog lists nobody
until you type.

Worth a second opinion: past the binding ceiling the queue answers an error rather
than a partial list, since a partial one would show bound accounts as unbound.


src/backend/services/identity-resolution/src/api/persons.rs
```diff
@@ -72,40 +88,118 @@ pub async fn search_persons(
...
-
+    let limit = listing::clamp_limit(params.limit, DEFAULT_LIMIT, MAX_LIMIT);
     let (named, values) = partition_terms(&terms);
 
-    // Over-fetch by one: the extra row is the truncation probe, never served.
-    let mut ids = if named.is_empty() {
-        persons_repo::search_persons_by_current_values(&state.db, tenant, &values, &[], limit + 1)
-            .await
-            .map_err(|e| read_err(&e))?
+    let query = terms.join(" ");
```
aleksdotbar, line 94: Why is that needed?

mozhaev-dev: The cursor carries the query it was issued for and is compared by equality, so it needs one spelling per search. `search_terms` has already split `q` on whitespace, so `?q=%20iva%20%20example%20` and `?q=iva+example` are the same search — joining the terms back with a single space makes them the same string too.

Without it the first page works and the second one 400s on a stray space:

```
GET /v1/persons?q=iva%20%20example        → next_cursor issued for "iva  example"
GET /v1/persons?q=iva%20example&cursor=…  → 400, "issued for a different query"
```

PR #2618, opened 2026-08-18
refactor(identity-helm): flatten the org-chart values to top level

aleksdotbar:
## Problem

`orgChartSourceType`, `expandSubordinates` and `maxDepth` are declared under an `identityResolution:` block inside the identity-resolution chart's own values, while every sibling block in that file — `image`, `service`, `gateway`, `seed`, `sync`, the probes — is top-level.

Because the umbrella aliases the whole subchart as `identityResolution`, the only path that reached those three keys was `identityResolution.identityResolution.orgChartSourceType`. A value set at the conventional `identityResolution.orgChartSourceType` was read by nothing, and Helm ignores unknown values rather than failing, so the render silently carried the defaults.

## Fix

- The three keys move to the top level of the chart's values; the ConfigMap references drop the `identityResolution.` prefix.
- A `fail` guard refuses to render when anything is still set under the retired path, so a stale values file is a loud error rather than a silent fallback to `bamboohr` / `true` / `16`.

Nothing in this repository sets the moved keys, so the default render is unchanged.

## Verification

- New render-contract test asserts the defaults, that a top-level override now reaches the gear config, and that the retired path fails the render.
- `pytest src/backend/services/identity-resolution/helm/tests/` — 22 passed (5 new, 17 existing CronJob contract tests).
- Subchart and umbrella renders both emit `bamboohr` / `true` / `16` by default; the umbrella honours `identityResolution.orgChartSourceType` and fails on the retired path.
- `helm lint` clean on both charts.

### DIFF
--- a/src/backend/services/identity-resolution/helm/templates/configmap.yaml
+++ b/src/backend/services/identity-resolution/helm/templates/configmap.yaml
@@ -119,9 +119,9 @@ data:
       identity-resolution:
         config:
           database_url: ""
-          org_chart_source_type: {{ .Values.identityResolution.orgChartSourceType | quote }}
-          expand_subordinates: {{ .Values.identityResolution.expandSubordinates }}
-          max_depth: {{ .Values.identityResolution.maxDepth }}
+          org_chart_source_type: {{ .Values.orgChartSourceType | quote }}
+          expand_subordinates: {{ .Values.expandSubordinates }}
+          max_depth: {{ .Values.maxDepth }}
           clickhouse_url: ""
           clickhouse_database: "identity"
           clickhouse_user: ""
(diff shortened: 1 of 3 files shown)


hello1101n (approved):

ktursunov (approved):

src/backend/services/identity-resolution/helm/templates/configmap.yaml
```diff
@@ -1,3 +1,6 @@
+{{- if .Values.identityResolution }}
```
ktursunov: Might be not required check. No deployment with that option on yet

aleksdotbar: Thanks

PR #2622, opened 2026-08-18
test(stand-ui): prove a real browser's screens reach the usage summary

hello1101n:
**Why.** The only writer this feature has is the browser — the SDK's `session_start` and `main.tsx`'s history subscription. The stand API tests post a hand-built body, so a build whose collector never starts passes every one of them.

**What changed.** One browser journey sweeps every screen the rail and the pane offer, then asserts each one is listed on Manage / Platform usage. The expectation is derived from the URLs the sweep visited, not from a list in the test.

**The pane is the denominator.** The journey pins `insight.portal.showPlanned` off, so only built screens render and a new one enters coverage without the test being touched.

**It brings up its own contexts instead of changing the suite's shell default.** Making the portal the default for every journey is a separate change.

On a local compose stand the sweep opens 27 screens and the journey runs in 22s.

- The SDK flushes on a five-second timer and nothing re-sends on unload, so the sweeper's context stays open until its events land.
- Rows are addressed by `data-index`, not position: the tables are virtualized, and the recycler invalidates a positional handle mid-read.
- The journey needs a source-served SPA — a front image built before the usage screen shipped marks the pane entry `readiness: "unbuilt"`.

**Out of scope.** Page views only — the `recordUsageEvent` actions, view-as sessions and the return-to-screen freshness case are not covered.

**Verified.** `pytest tests/stand/ui/test_platform_usage.py` against a local compose stand serving the SPA from source: 1 passed, twice. `ruff check` and `ruff format --check` clean.


hello1101n:
Superseded by #2636 — the stand-ui journey now rides with the frontend fix it covers, as a single commit on `AP-0-usage-action-event-path`.

PR #2644, opened 2026-08-19
fix(seed): stop refusing to reseed a stand the suite has run against

ktursunov:
`Deploy test stand after publish` fails at `Stage 3/5 — seed the stand` on every run, and stages 4 and 5 skip behind it — so smoke and the full stand suite have not run since the suite started running per merge. This restores the pipeline and makes a guard refusal legible in the CI log.

## The defect

Two contracts, each correct alone, first met when the suite began running after every merge:

- The suite's resolution round trip binds a scratch account and "cleans up" by excluding it — but both verbs append to identity's append-only journal, so two `persons` rows carrying `operator-bind` / `operator-exclude` survive every run **by design**. Their value carries the suite's `stand-scratch` mark, which exists precisely so its rows are identifiable on sight.
- The seed preflight counts any `persons` row whose `reason` is outside the seeder's namespace as "this stand is somebody's directory" and refuses — the guard that keeps the silver step's TRUNCATE off a stand that genuinely holds someone's data.

So one full suite run permanently blocked every later seed. The refusal never reached the log: the workflow republishes only `==>`/`ERROR:` lines, and the Job is reaped an hour later, so a guard refusal read exactly like a crash.

## What changed

| Piece | Now |
| --- | --- |
| `_count_foreign_persons` | exempts the suite's rows by its `stand-scratch` value namespace **under its own connector instance**, as ONE negated conjunction |
| `scratch.py` | owns that connector identity (`SCRATCH_SOURCE_TYPE` / `SCRATCH_SOURCE_ID`) beside `SCRATCH_PREFIX`, with the rule that produced it; `insight_seed.config` mirrors all three and a drift test fails at the edit rather than weeks later as a refused seed |
| `seed-stand.sh` | on a failed Job, re-emits everything from `PreflightError:` onward under an `ERROR:` prefix — capped at 40 lines, since this is the one place pod output crosses into the allowlisted channel |

The exemption is deliberately not keyed on the value alone: that would exempt anything writing a `stand-scratch` value, whoever wrote it. Nor as three separately negated terms — that would subtract every row under the suite's connector and every row carrying the prefix independently, which is a weaker guard rather than a narrower one. An operator's real corrections on that same connector still refuse, which is the guard's whole purpose.

## Verification

- The predicate is tested against an engine rather than asserted about as SQL text: an in-memory sqlite `persons` table (with `value_effective` as the generated column it really is) over nine rows — the suite's bind and exclude rows, the prefix carried in `value_full_text`, the same prefix from another connector type and from another instance of the same type, an operator correction on the suite's own instance, a foreign row with no value at all, the seeder's own row, another tenant's. Each row alone, then all at once, so no exemption steals another writer's rows.
- End to end on a stand, before and after: with the round trip's rows present, the pre-change guard refuses with `PreflightError` naming exactly 2 rows and the seed aborts; with the change, the same stand and the same rows give `preflight ok` and the seed completes. One file different.
- The full suite then ran against the deployed CI test stand: **api 354 passed / 19 skipped / 2 xfailed**, **ui 18 passed**. Skips are stand-shape (no service principals, no second tenant, opt-in rebuild lane); both xfails are known and referenced.
- Seed package: 42 `unittest discover` / 47 pytest, ruff, `ruff format --check`, mypy clean. The exact CI invocation inside the built image passes with the checkout-only drift test skipping, as designed. `bash -n` and shellcheck clean.

## Notes for review

- **The identity projection has the same class of gap and is deliberately left alone.** `AssignmentKind::MintedFromRoster` stamps `roster-mint` and a binding the projection did not have to decide is stamped with nothing at all — both outside the reasons this guard accounts for, so a projection run that minted or reused could resurrect the refusal from the other side. It has not fired because the seeded roster all carry addresses, so the projection stamps `auto-seed-link`, which was already exempt. Widening the guard for it belongs in its own change, not in one asked to narrow.
- Nothing in the code stops a future stand test from writing a correction under a different connector, and the seed refuses again when one does. `merge` and `detach` happy-paths would do exactly that. The constant's comment says so; enforcing it needs a check that does not exist yet.

### DIFF
--- a/src/ingestion/tools/seed/insight_seed/config.py
+++ b/src/ingestion/tools/seed/insight_seed/config.py
@@ -29,6 +29,16 @@ SEED_REASON_PREFIX = "seed.py "
 #: The persons-seed's email-link reason (AUTO_SEED_LINK_REASON): a rebuildable projection, not foreign data.
 PERSONS_SEED_LINK_REASON = "auto-seed-link"
 
+#: The stand test suite's scratch namespace, and the connector instance it writes
+#: its journal rows under. The suite cannot delete what its resolution round trip
+#: appends (the journal is append-only), so preflight recognises those rows by
+#: this triple — a value prefix alone would exempt anything that writes one.
+#: INVARIANT: must match SCRATCH_PREFIX, SCRATCH_SOURCE_TYPE and
+#: SCRATCH_SOURCE_ID in tests/stand/api/scratch.py.
+STAND_SCRATCH_PREFIX = "stand-scratch"
+STAND_SCRATCH_SOURCE_TYPE = "github"
+STAND_SCRATCH_SOURCE_ID = "01900000-0000-7000-8000-00000000feed"
+
 TENANT_ENV = "TENANT_DEFAULT_ID"
 ANALYTICS_DB_ENV = "MARIADB_ANALYTICS_DB"
 IDENTITY_DB_ENV = "MARIADB_DB"
--- a/src/ingestion/tools/seed/insight_seed/preflight.py
+++ b/src/ingestion/tools/seed/insight_seed/preflight.py
@@ -14,7 +14,14 @@ Two signals, because neither sees everything:
 * `persons` rows in the target tenant whose `reason` is outside this seeder's
   namespace. Identity rows are additive, so this one is about not mixing demo
   people into somebody's directory — but it is also the ONLY signal that works
-  on a single-tenant stand, so the silver step consults it too.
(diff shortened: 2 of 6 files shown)


aleksdotbar (changes requested):
The preflight exception is too broad for a safety guard: any foreign row with a `stand-scratch` value bypasses the refusal. Restrict it to the suite’s fixed source type and source ID as well, and cover a same-prefix row from a different source.

aleksdotbar (approved):

PR #2647, opened 2026-08-19
ci: stop reinstalling Chromium in Compose stand tests

aleksdotbar:
**Why.** Browser setup downloads Chromium and system dependencies during every Compose stand run. This step can delay or prevent the test suite from starting.

**What changed.** UI journeys now run in a pinned Playwright browser image that already contains Chromium and its required system libraries. The runner still creates the locked Python test environment and executes the current pull request's test source against the existing stand through the gateway network.

**Expected effect.** Removes repeated browser installation from the critical path. It does not remove the stand startup or backend-image build time; a cold image pull may still take time.

**Verified.** Workflow YAML parsing, actionlint, shell syntax, and diff checks pass. The pinned image was exercised locally with the locked dependency environment; `insight_stand` imports and Chromium starts successfully.


hello1101n (approved):

ktursunov (approved):

aleksdotbar:
Closed as not relevant for now

PR #2656, opened 2026-08-19
ci(stand): hold the stand for drafts and bound the browser install

ktursunov:
Two ways the compose stand lane spent runner time nobody could use, both found while pushing #2644.

| Before | Now |
| --- | --- |
| The stand ran on every push to every pull request, drafts included — 45+ minutes per push, for an answer nobody acts on while the change is still moving | A draft reports the required check without running the lanes; the stand runs from the moment it is marked ready for review, and on every push after |
| `Install the suite's browser` had no ceiling and runs under `continue-on-error`, so a stall there was silent: no browser, no journeys, the job's whole budget gone, and neither lane reported | Bounded at 10 minutes, so the UI lane fails honestly and fast while the API lane still reports |

## Holding the stand for drafts

The `changes` job reports `draft` and holds the lanes back — no `resolve-target`, no stand, no browser. Two constraints ruled out the simpler implementations:

- **The required check must still report.** GitHub leaves a required check whose workflow never ran pending forever — the same reason this workflow deliberately carries no `paths:` filter. So the reporter jobs keep running and pass; only the stand is skipped. Gating the lane while leaving `relevant=true` would instead have made the reporters fail with "API smoke did not pass", since they read an empty lane result as a failure.
- **`ready_for_review` is load-bearing, not tidiness.** The default type list stops at `synchronize`. Without that type, marking a pull request ready fires no event, the draft run's green `Stand E2E` stands as the required check, and a change could merge having never met the stand.

The reporters read `draft` before relevance and say so in their own words rather than borrowing the irrelevant-diff message, because on a draft the diff was never examined. `merge_group`, `schedule` and `workflow_dispatch` carry no pull request, so the draft context renders empty and they are never held back — an empty value reads as "not a draft" rather than tripping `set -u`.

## Bounding the browser install

This is the one step in the job that can block on a system package manager — `--with-deps` installs the browser's apt dependencies — and it runs under `continue-on-error`, which hides the difference between slow and stuck. Ten minutes matches the same command's ceiling in `run-stand-suite.yml`, against a couple of minutes when it works. A step that times out leaves `outcome` failure, which `Record lane results` already reads — so the UI lane fails, the API lane still reports independently (the INVARIANT at the top of the file), and teardown still runs.

The sibling steps stay unbounded deliberately: they fail rather than hang, and the job's own ceiling covers them. This one needed its own because `continue-on-error` means nothing else would ever say so.

## Verification

Extracted the real `changes` filter step out of the workflow and ran it against each event shape, checking the outputs it writes:

| Case | `draft` | `relevant` |
| --- | --- | --- |
| draft pull request | `true` | `false` (lanes held) |
| ready pull request | `false` | `true` (judged on its diff) |
| merge_group (empty context) | `false` | `true` |
| schedule (no base sha) | `false` | `true` |

The workflow parses as YAML with the wiring intact: four trigger types, `draft` among the `changes` outputs, all three reporter jobs reading it, and the install step carrying both the ceiling and `continue-on-error`.

## Notes for review

What cannot be proven locally is the `ready_for_review` trigger itself, which only exists once the file is on `main` — worth watching on the first draft marked ready after this lands.

**On the cause, stated no further than the evidence goes.** What prompted the second change was observed twice on #2644: the lane ended at exactly 55m00s, both times with the install step the one in progress, and with `Bring the stand up`, `Run the API suite` and `API coverage gate` already green inside it. That 55-minute figure matches nothing in this workflow — the job's own ceiling is 45 or 75 — so the kill comes from outside it, and GitHub has since re-queued the job, which is what it does after a runner-level failure rather than a hung command. So this change does **not** claim the install was hanging, and does not claim to fix those kills. It removes the ambiguity: with a ceiling, a genuinely stalled install fails at ten minutes and says so, and anything still dying at 55 is demonstrably not this step.


aleksdotbar (changes requested):
Draft PRs currently fail `changes`: `emit false false false` passes unused arguments, while `emit` reads unset shell variables under `set -u`. Initialize the outputs before calling `emit` so drafts report `draft=true` and skip the stand.

PR #2695, opened 2026-08-20
feat(connectors): key GitHub identity on the immutable member id

mozhaev-dev:
**Why.** GitHub logins are renameable by their owner and re-registrable once freed, and their letter-case is preserved inconsistently across the pipeline — keying identity on the login breaks sign-in on a rename and can hand one person's history to another on reuse.

**What changed.** `github-directory` now keys the account — bronze `unique_key`, the fields-history entity, and the `value_type='id'` binding — on the member's immutable numeric GitHub id, stringified; the `github` connector claims commit e-mails by `author_id`/`committer_id`. The login remains a `username` observation for display and search.

**No migration between the two keyings — deliberate.** An instance that ran the login keying wipes its identity data and re-syncs; documented in the connector README.

**The shared `identity_inputs_from_history` macro changed for every connector.** Its observation key now includes `source_id` (same numeric entity id under two connections must not collide), and duplicate same-instant rows are collapsed.

**Out of scope.** The IdP/broker config flip — the external-id claim must carry the numeric id as a string — is a per-instance deploy step applied together with the data reload; recipe in the `github-directory` README.

**Verified.** Both git connectors' unit suites, strict manifest validation, `dbt parse`, and a scratch-ClickHouse build of the changed models covering the multi-org and multi-connection collision cases. The rewritten e2e binding tests run in the identity lane.


aleksdotbar:
The pre-2017 noreply branch can keep its account link without weakening the id keying: bronze already holds login↔id pairs wherever GitHub matched an account (`pull_request_commits` author/committer, `commit_authors`, the roster's `login`+`member_id`). One CTE resolves the legacy form's login through that map:

```sql
login_to_id AS (
    SELECT login, any(id) AS account_id
    FROM (...login/id pairs from those sources...)
    GROUP BY login
    HAVING uniqExact(id) = 1   -- ambiguous login resolves nobody
)
```

The `HAVING` keeps the rename/recycle protection: a login seen with two ids maps to nobody, and anything unmapped still falls through to the unowned-email queue as designed. On a seeded stand this recovers both legacy addresses that currently lose their link (one of them a real member), ~2 CTE-lines of cost, no API calls.

Caveat worth a comment in the model: the map reflects current login ownership, so a recycled login whose old owner never appears in our data would map to the new one — same misattribution the login keying has today, but now bounded to pre-2017 addresses with an unambiguous map.

mozhaev-dev:
Landed in 8c745121b, thanks — with two adjustments to the sketch: the map lowercases the login on both sides (logins are case-preserving, the address's local part is not reliably cased), and it is scoped per (tenant, source) like the rest of the model, so the same login on two hosts never cross-resolves. `HAVING uniqExact(id) = 1` kept as proposed — an ambiguous login resolves nobody and the address falls through to the unowned queue. The current-ownership caveat is a comment on the CTE.

Verified on a scratch ClickHouse with synthetic rows: legacy address with a roster pair resolves to the numeric account, modern form unaffected, unmapped and ambiguous logins stay unowned. The combined e2e test now seeds a legacy-noreply commit and expects its e-mail claim under the numeric account id.

aleksdotbar (approved):

Issue #2709, opened 2026-08-20
feat(ai-cost): distribute the billed extra-usage cost across days


Gregory91G:
Child of #2479.

## Summary

`ai.extra_usage_cost` plots one point per month, on the day a snapshot was last read. The spend happened across the month, so the chart attributes a month's money to a single day.

The vendor endpoint reports only a month-to-date total for the month in progress and keeps no history — a daily figure can only be a difference between consecutive snapshots. Those snapshots do not survive today: bronze keys a seat's reading by the extraction month and the table is a ReplacingMergeTree, so each day's read replaces the one before it and only a month's closing figure is left. The connector's bronze key has to carry the read day first — the same move #2528 already made from seat-only to seat-and-month. Daily detail therefore begins at deployment; readings already overwritten cannot be recovered, so months before it keep their single point permanently.

## What lands

- **`ai.extra_usage_cost`** stays as it is: month grain, the vendor's exact figure.
- **`ai.daily_approximate_extra_usage_cost`** — new: the same money distributed across days, computed as the day-over-day difference of the cumulative total.

Per complete month `sum(daily) == monthly`, by construction rather than by convention: a cumulative series is corrected to its suffix minimum before differencing, which also keeps every daily value non-negative and the series monotonic.

Approximate by nature, and the wording says so — the first point of a month covers the days before the first snapshot, and a missed run makes the next point cover several days. The sum stays exact. Labels split accordingly: the monthly metric reads as exact, the daily one carries `approximate distribution`.

## Salvaging what has not merged yet

A ReplacingMergeTree collapses on merge, not on insert, so readings that were
logically replaced may still be physically present in unmerged parts. The first
build of the new daily staging model is the one chance to keep them: it reads
bronze without `FINAL` and dedups per day rather than per bronze key, so
whatever has survived is captured and then held by a relation that is never
rebuilt. This recovers a tail of recent days, never a history, and how much
depends only on what ClickHouse has not merged by then.

Two things follow. The model's first build should happen as early as possible
after the change lands. And the deploy is worth doing with merges paused on that
one table, so nothing is collapsed between the connector change and the capture:

```
SYSTEM STOP MERGES bronze_claude_team.claude_team_overage_spend;
-- run dbt so claude_team__ai_overage_daily is built
SYSTEM START MERGES bronze_claude_team.claude_team_overage_spend;
```

Both invariants are recorded in the model, because a later reader would
reasonably take the missing `FINAL` for an oversight and "fix" it.

[...]

PR #2715, opened 2026-08-20
feat(ai-cost): serve the billed extra-usage cost distributed across days

Gregory91G:
Refs #2709. Reads the per-day readings #2714 landed.

**Why.** `ai.extra_usage_cost` reports a month's billed spend against the day the seat was last read, so a chart of it puts a month's money on a single day. The vendor states only a running month-to-date total and keeps no history, so a day's figure exists nowhere and can only be recovered as the step between two readings.

**What changed.** Those steps are served as `ai.daily_approximate_extra_usage_cost`, over the readings `class_ai_overage_daily` carries. The monthly metric keeps its grain and its exact figure.

**Two metrics, never added.** The monthly one is the vendor's own number and stays exact; the daily one is the same money placed approximately, and both descriptions say which is which. The split is what lets a short window keep working: the monthly metric returns a whole month by design, the daily one returns only the days in range.

**A correction rewrites the days it belongs to.** A running total may not fall; where it does the vendor has revised the month down, and the revision belongs to the days already reported rather than the day it arrived. Each reading is corrected to the minimum of itself and every later one before differencing. No step is negative, the series only rises, and the steps telescope to the month's final reading — so `sum(daily) == monthly` over a whole month holds by construction, not by arrangement.

Fixture, cents as the vendor reports them and dollars as the metric serves them. erin is read 500 then 400:

```
person   Dec 05   Dec 06   Dec 07   corrected steps   window sum
alice       100      300      300    100 / 200 / 0     $3
bob         200      500        —    200 / 300         $5
carol       300      700        —    300 / 400         $7
dave        400      900        —    400 / 500         $9
erin        500      400        —    400 / 0           $4
```

A naive difference would emit −$1.00 for erin on Dec 06; the case asserts 0.00, and asserts no point anywhere is below zero.

**Closes a gap #2714 left.** That PR added `snapshot_date` to the connector schema and the bronze DDL but not to `src/ingestion/tests/e2e/metrics/schemas/bronze_claude_team.claude_team_overage_spend.yaml`, a hand-maintained mirror carrying `additionalProperties: false`. The gap was latent until some fixture first set the field, and this branch's template is that fixture — so the property is added here, along with the `drilldown_matrix.py` entry every new metric needs.

**And one this branch would have left.** The evidence model emits four measure keys while the `accepted_values` test on `ai_cost_metric_observations.measure_key` named two, so a `dbt build` selecting the gold tag fails on the contract once the seat or daily rows exist. `daily_extra_usage_usd` arrives here; `seat_cost_usd` was already emitted and already absent, and it is added alongside rather than left half-applied. The gap stayed invisible because the metric fixtures reach gold through `dbt run`, which skips tests.

**Scope widened after review: the AI cost explanations are rewritten.** `explanation` reaches a hover tooltip and, joined to `description`, a screen reader — so a metric key in it names something the reader cannot look up, and a connector version names something they cannot see. All four cost metrics now refer to each other by label, `ai.cost` and `ai.seat_cost` included even though they predate this branch: a convention applied to half the family is worse than none. The correction rule and the span each point covers stay documented where they execute, in the gold evidence model.

**Out of scope.** Nothing in the UI reads the new metric yet — the AI & Cost chart still plots the monthly figure. `ai.extra_usage_utilisation` deliberately stays cumulative: proximity to a ceiling is a running total, and a daily step over a ceiling means nothing.

**Verified.** The `ai` shard of Bronze-to-API run locally from an emptied warehouse, sharded exactly as CI shards it — 25 passed in 7m00s, the new fixture among them. That lane is deferred to the merge queue on pull requests, so this is the only evidence available before queueing. `cargo test -p analytics` 444 passed, 0 failed, including the passport-drift and registry-count gates. `dbt parse` clean on a full non-partial parse. All 46 metric fixtures load, and dropping the new schema property again fails exactly the four that share the `claude_team_overage` template. The drilldown matrix set-equals the registry at 66 metrics, and the metric-coverage gate asserts all four required views. Not run: the deployed-stand suite, which needs a stand.

## Summary by CodeRabbit

* **New Features**
  * Added daily approximate AI extra-usage cost reporting.
  * Daily costs account for usage corrections and preserve monthly billed totals.
  * Added support for daily usage evidence and tool- and seat-tier breakdowns.

* **Documentation**
  * Clarified billing behavior and relationships among AI cost metrics.
  * Updated Claude Team usage schema documentation.

* **Tests**
  * Added end-to-end coverage for daily cost views, corrections, and periods without readings.
  * Updated metric registry expectations.


hello1101n (approved):

aleksdotbar (changes requested):
The metric explanation is user-facing. It should be clear, concise, understandable and should not include implementation details like metric keys

aleksdotbar (approved):

PR #2754, opened 2026-08-22
feat(seed): publish the stand manifest to a durable ConfigMap

ktursunov:
Stacked on #2735 — review that first; this branch's own diff is the second commit.

**Why.** The manifest describes a stand that lives until someone tears it down, but its only copy outliving the seed pod was the Job log, which Kubernetes deletes an hour later. Recovering it after that meant re-seeding the stand — replacing its data to read a description of it.

**What changed.** A completed `--step all` publishes the compact manifest to `configmap/seed-stand-manifest`. `seed-stand.sh` already recovers the document into a shell variable and already runs under a credential granted `configmaps create/update`, so the writer is the script.

**Nothing is added to the seed pod.** No ServiceAccount, no RBAC object, no Kubernetes client, no image or dependency change. An in-pod writer would have needed all four, plus a preflight access review, to reach a capability the script already has.

**Publishing is warn-only.** It happens after every database write in a Job with `backoffLimit: 0`, and the log sentinel still carries the document — a failed publish must not report a seeded stand as broken.

**The compact rendering is stored, not the file rendering.** Measured at the roster ceiling: compact 624 KB (59% of the 1 MiB object cap), file 871 KB (83%). The apply is server-side because a client-side one stamps `last-applied-configuration`, whose 256 KiB cap the document crosses inside the supported range.

**Out of scope.** Readers still parse the log; moving them, and then deleting the chunked transport, are separate changes.

**Verified.** 150 seed tests, ruff, mypy, shellcheck clean; a new test fails if a ceiling bump pushes the manifest past the ConfigMap cap. Stand behaviour is unverified until this deploys.

### DIFF
--- a/.github/workflows/deploy-test-stand.yml
+++ b/.github/workflows/deploy-test-stand.yml
@@ -553,6 +553,20 @@ jobs:
 
       # BETWEEN capture and upload, and always — a failed run is exactly when a
       # trace exists. A Playwright trace replays a real signed-in session: it
+      # Affirms the durable copy matches what this run emitted, so phase 2 can
+      # let readers depend on it. Last and always(): a broken publish turns the
+      # run red without blocking the smoke or suite stages of a seeded stand.
+      - name: Verify the published seed manifest
+        if: ${{ always() && steps.seed.outcome == 'success' }}
+        run: |
+          set -euo pipefail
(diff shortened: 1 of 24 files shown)


ktursunov:
Folded into #2735 — the ConfigMap publish ships with the org-size knob rather than as a separate PR.

PR #2781, opened 2026-08-22
feat(frontend): let the day strip reach the current day

### LINKED ISSUE #2734: Period presets stop at yesterday, so the current day is collected and drawn but never requested
**Why.** Every period preset ends yesterday — `periodToDateRange` subtracts a day before choosing the preset, so today is never requested. Two merged changes exist only to serve the current day: #2694 made it arrive (bronze now reaches the run day, previously two days short) and #2710 taught the strip to draw a day as not-collected or still-open-to-revision. Nothing asks for it, so both are inert for the day they were built for.

**What we want.** The surfaces that read a metric one day at a time reach today, and today is drawn as what it is: partial. #2710 already carries the vocabulary — a provisional day keeps its real height and fades, an uncollected one is a faint wash — so a partial day no longer has to look like a dip.

**Decision taken.** Only the day-reading surfaces include today, not every metric on every screen. A partial day entering a period total is the reason the bound exists, and that reason still holds for a total; it does not hold for a strip that draws each day on its own.

**Scope.** `periodToDateRange` is therefore not the lever. It is shared, and one `period` field serves every view of a batched `/v1/metric-results` request, so widening it there reaches every total as well. A day-reading surface needs its own resolved range — a separate request for those items, or a range resolved at the section rather than at the portal. Custom ranges already pass through untouched and can end today, which is the cheapest way to see the intended behaviour before changing anything.

**Not part of this.** The month-grain cost facts no longer depend on what this decides — they are anchored to the first day of the month they bill for, in #2759.

**Done when.** A day-reading surface shows today; today renders as provisional rather than as a settled value or as silence; period totals are unchanged and still end yesterday.

Gregory91G:
Closes #2734.

**Why.** The day the sources now deliver is never on screen. Every period preset ends yesterday, and the strip inherited that bound, so #2694 (which made the current day arrive) and #2710 (which taught the strip to draw a day as not-collected or still open to revision) are both inert for the day they were built for.

**What changed.** The strip asks for the current day. It runs on its own `/v1/metric-drilldown` request, so the widened window reaches the strip and nothing else — period totals keep the bound they were measured on.

**Only the day-reading surface, not every metric everywhere — the decision re-taken on #2734.** A partial day entering a period total is the reason the bound exists, and that reason still holds for a total. It does not hold for a strip that draws each day on its own and marks an incomplete one.

**Only a window that stops exactly at yesterday is widened.** A custom range left in the past stays there, one already reaching today is untouched, and one sitting at the 400-day request cap is left alone rather than made unanswerable.

## Left in place on purpose

The timeseries view. It asks for its buckets and its period total in one request over one window, so it cannot take the current day for the buckets without taking it for the total — and over a long enough range it buckets by week or month, where a partial day is not the question. Widening it needs a second request, which is a larger change than this decision calls for.

**Verified.** Not run here: `tsc`, `eslint` and `vitest` — Node 24 and pnpm 10.33 are not installed on this machine, so CI is the first run of the two new test blocks (`throughToday` in the range unit, and a strip case that fakes the clock so the fixture's window ends yesterday). No screenshot: the strip needs a stand this box cannot reach, and the rendering it exercises is #2710's, unchanged here.

## Summary by CodeRabbit

- **New Features**
  - Metric activity periods ending yesterday now automatically extend through today.
  - Today is displayed as not yet collected when current-day data is unavailable.
  - Charts and accessibility labels clearly reflect the current day’s collection status.
  - Existing ranges remain unchanged when they already include today, intentionally end earlier, or exceed supported limits.

- **Bug Fixes**
  - Evidence details and daily activity visualizations now use consistent date ranges.


Gregory91G:
Closing unbuilt. The figure this was meant to rescue is served without it: #2760 anchors the month-grain cost facts to the first day of the month they bill for, so their date no longer moves with the sync schedule, and the surfaces that need a window-comparable number read the per-day distribution instead.

What is left over is a day of label, not a day of data. The sources deliver a day after it closes — the daily streams re-read a two- to three-day tail and the catalogue declares a three-day revision window — so the current day on a strip is almost always a not-collected wash rather than a partial bar. Against that, the widening costs a permanent one-day mismatch between the strip and the period total printed above it.

If a source ever delivers the current day as a matter of course, this is one helper and two call sites away.

PR #2858, opened 2026-08-25
feat(analytics): tell an operator which connectors delivered and when

mozhaev-dev:
**Why.** Nothing said whether a connector was still delivering, so a source that quietly stopped looked like one nobody had ever configured.

**What changed.** An append-only ledger, `ingestion_runs.pipeline_events`, written by the pipeline as a run happens and by the reconcile sweep on its tick; the pane reads one relation from it. Nothing on the request path calls the mover or touches bronze.

**Absence is recorded as absence, never as zero.** Delivery, counters, elapsed time and start time are all nullable for this reason: a measured `0` beside recorded records is the finding this page exists for, and a NULL means nobody asked.

**No compose support, by decision.** The stand declares `ingestion: false`, so tests needing recorded runs skip with a reason rather than passing vacuously.

The pane, reading recorded runs

Seeded by hand on a local stand: a delivery mismatch above a failed run, the failed step named, and `0 / not measured` where a zero would have been a lie. A fresh install shows an empty state that says so, and a non-admin is refused by name:

Empty state
Refusal

An Argo DAG's phase is assessed over its *targets* — the tasks nothing depends on. A recorder appended after the work it records becomes the only target, so one succeeding after a failure **would erase it**: a failed sync would read as a green run and dbt would rebuild stale bronze. `dag.target` names the real work, pinned by a test that fails without it:

```
AssertionError: the poll's outcome must reach the phase assessment; with only
the recorder as a target a failed sync reports a green run. targets={'record'}
```

The sweep's frontier moves along `job_created_at`, the field the mover's listing is ordered by — a start time would let a job that waited to begin push the cursor past jobs nothing has read.

**Out of scope.** `ingestion.reconcile.claimHorizonSeconds` is chart-declared and required, but not bound to the workflow layer's real GC retention, which this chart does not configure. Draft #2713 is superseded.

**Verified.** Stand suite (444) with both operations in its coverage gate, against a ledger holding rows — an empty one hides a schema mismatch. Migration applied and re-applied on a live warehouse. Every review fix carries a test checked to fail without it.

## Summary by CodeRabbit

* **New Features**
  * Added an admin-only Connector Health dashboard showing connector status, sync outcomes, delivery metrics, storage usage, stream details, and recent run history.
  * Added API access for overall connector health and connector-specific run history.
  * Added ingestion run tracking, reconciliation, and status recording across sync and pipeline workflows.
  * Added configurable reconciliation coverage horizon and support for identifying workflow and out-of-band syncs.

* **Tests**
  * Added comprehensive coverage for dashboard states, API responses, permissions, workflow tracking, and reconciliation behavior.

* **Documentation**
  * Added Connector Health product requirements, design documentation, and API specifications.


aleksdotbar (approved):

mozhaev-dev:
Closing in favour of a smaller first iteration.

**Why.** This branch answers "did the data arrive" by recording from inside the ingestion
pipeline and cross-checking the mover's own count against storage. That capability is real
and it works, but it is two thirds of a 54-file change and it puts recording code on the
critical ingestion path — where a recorder appended after the work it records becomes the
only task its DAG's phase is read from, and one succeeding after a failure would mask that
failure. Guarding that hazard is most of the remaining complexity.

**What replaces it.** One writer instead of two: the reconcile loop reads the mover's job
history on its tick and records it. The page then answers which connectors are configured,
when each last synced, how that sync ended, and the connector's recent sync history — all
of it outliving the mover's own job retention, which a deleted connection takes with it.

What it deliberately will not claim is delivery. The mover's account is uncorroborated, so
no state on that page says "delivering" and the record count is labelled as reported.

**Not lost.** The branch stays for reference — the delivery measurement, the exit-handler
recording and the phase-contract test are the second iteration, if the first proves the page
is worth deepening.

PR #2868, opened 2026-08-26
feat(analytics): bin a histogram per dimension tuple, pooled across entities

mitasovr:
**Why.** The histogram view bins one distribution per entity, so "how long do pull requests in this repository stay open" has no shape today — binning per person and merging the bins is not the repository's distribution.

**What changed.** `{view: "histogram"}` takes optional `dimensions`. With them, every selected entity's events pool into one distribution per dimension tuple and rows carry `dimensions` instead of an `entity_id` — the same no-entity-grain shape rollup already returns. The bin math is unchanged: fixed width over the partition's own `[min, max]`, the partition being the tuple rather than the person.

**The dimensionless request is byte-identical to before.** `{"view": "histogram"}` still parses, still bins per entity, still lists every requested entity — pinned by a deserialization test and the untouched per-entity tests.

**A pooled tuple with no events has no row,** where the per-entity shape lists every requested entity with empty bins. There is no roster to list tuples against, so absence is absence — rollup's rule.

**No group-limit support.** Group cardinality is bounded by the query row limit and the post-hoc view row limit, as it is for an uncapped rollup; a cap can be added when a caller needs one.

**Out of scope.** The frontend only gains the response type — no UI reads pooled bins yet; the Repositories lens small-multiples that motivated this ship separately.

**Verified.** `cargo test -p analytics` (561), `cargo clippy --all-targets -D warnings`, the OpenAPI drift gate regenerated, and the generated stand schemas refreshed. The compiled SQL was run against ClickHouse 25.7.5 on synthetic rows: two people's events in one repository pool to a shared `[10, 40]` range, and a repository whose values are all identical collapses to a single bin whose count is both people's — the two facts the new e2e fixture asserts.

### DIFF
--- a/docs/components/backend/analytics/openapi.json
+++ b/docs/components/backend/analytics/openapi.json
@@ -624,20 +624,29 @@
         "type": "object"
       },
       "HistogramValueDto": {
+        "description": "One histogram row. Per-entity shape: `entity_id` set, `dimensions` absent,\nevery requested entity listed. Pooled shape (dimensioned request):\n`dimensions` set, `entity_id` absent, one row per observed dimension tuple\nover all selected entities' events — no entity grain, like rollup.",
         "properties": {
           "bins": {
-            "description": "Empty when the entity has no events in the period — the entity is\nstill listed, mirroring the period view's every-requested-entity rule.",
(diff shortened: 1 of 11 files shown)


mitasovr:

PR #2870, opened 2026-08-26
feat(frontend): build a report whose rows are repositories

mitasovr:
**Why.** A report could only be a person per period. Asking "what happened in each repository" had nowhere to go: the table skipped every dimensioned series and keyed each row on a person.

**What changed.** A **Rows** control picks between People (unchanged) and Repositories. The repository shape asks for the same timeseries grouped by `repository` and sums every person's contribution into the repository's own row, so a row is one repository per period with no person columns. CSV and XLSX export the table they are given, so both shapes download the same way.

Report builder in repository mode: the Rows control, the changed subtitle, and non-additive metrics greyed out

**Only additive metrics are offered per repository.** Every row there is people added up, so the question the quarterly rollup already asks arrives at every granularity — a repository's median is not the median of its people's medians. The picker states that per metric instead of hiding it.

**No group limit on the request.** Repositories are few enough to name them all, and a capped request folds the tail into a remainder series a per-repository table cannot label.

**Grouped requests carry a budget, not a bound.** How many repositories a person touched is data, so batching assumes a small number per person to keep grouped requests about as heavy as ungrouped ones; the server's row limit is what actually refuses an oversized answer.

**Out of scope.** Other dimensions (project, team) — the row mode is written as a dimension so adding one is a list entry, but only repositories are wired.

**Verified.** The full frontend suite (2136), `tsc -b`, and eslint. New cases cover the summing across people, the label-over-id naming, the ordering, the ignored ungrouped series, the availability gate in both modes, and the grouped request shape. The screenshot is a local mock-data stand.

### DIFF
--- a/src/frontend/src/components/portal/report-builder-view.test.tsx
+++ b/src/frontend/src/components/portal/report-builder-view.test.tsx
@@ -164,6 +164,25 @@ describe("ReportBuilderView", () => {
     expect(isDisabled("Commits")).toBe(false);
   });
 
+  it("refuses a non-additive metric in repository rows at any grain", async () => {
+    // Per repository every row is people added up, so the additivity question
+    // arrives immediately rather than only at a quarter.
+    const user = userEvent.setup();
+    render(<ReportBuilderView />);
+    expect(isDisabled("Merge rate")).toBe(false);
+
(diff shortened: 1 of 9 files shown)


mitasovr:

PR #2872, opened 2026-08-26
test(stand): cover the rollup view and the Development lenses

mitasovr:
**Why.** The `rollup` view had no deployed-path test, and the Repositories and Quality lenses none at all — both were placeholders until recently, so a nav entry still routing to the roadmap note would pass every test in the suite.

**What changed.** Four contract cases join `analytics/test_results.py` (the module that owns `POST /v1/metric-results`, per the one-module-per-path-group rule): the rollup's shape, its reconciliation against the period view, its visibility gate, and the `comment_target` vocabulary. One browser journey reads both lenses in a single sign-in.

**The rollup's oracle is the API's own period view, not a written-down number.** For a summed metric, every repository's share added together is the person's total over the same window; the two travel different SQL, so a drift is a real defect in one of them and neither figure is hand-authored — which keeps it honest on any seed.

**The rollup gate gets its own security case even though the 200 already has one.** A rollup row carries no entity id, so a skipped gate would leak an outsider's work as an unattributed total that no other assertion could trace.

**Sections are asserted as "at least one", the heading strictly.** Several sections drop themselves by design — a composition or a table with a single row is an empty shell — so naming one would tie the journey to how many repositories a stand happens to hold. The heading is the claim that catches the regression that matters.

**A stand with no git data skips rather than fails.** Nothing here can seed a git connector, so absence is distinguished from breakage: the lens must say "not connected yet", and the table's absence is a skip with the reason.

**Out of scope.** No new operation rows — every case uses `POST /v1/metric-results`, already catalogued and already swept for 401.

**Verified.** Collection (14 API cases in the module, 2 journeys), ruff check and format. Not run against a stand: none is up here, so the assertions are written from the handlers and the lens configs rather than observed — CI's stand lane is what executes them.

### DIFF
--- a/tests/stand/api/analytics/test_results.py
+++ b/tests/stand/api/analytics/test_results.py
@@ -24,10 +24,12 @@ from insight_stand import ApiClient, ApiResponse, Manifest, analytics_path
 from insight_stand.api import JsonValue
 
 from ..schemas import (
+    BreakdownView,
     MetricDefinitionListResponse,
     MetricResultsResponse,
     PeriodView,
     ProblemDocument,
+    RollupView,
 )
 from . import query_window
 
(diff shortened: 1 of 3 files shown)


mitasovr:

PR #2881, opened 2026-08-26
feat(ai-cost): bind an invoice line's tier to the tier a seat carries

### LINKED ISSUE #2620: Only the first connected AI billing account's invoices are counted in AI cost metrics
## Summary

Once two AI billing accounts are connected, the AI cost metrics only count invoices from whichever one synced first — the other account's invoiced spend never appears.

## Steps to Reproduce

1. Connect one AI vendor's invoices and let it sync, so invoice data has landed.
2. Connect a second invoice source — another AI vendor, or a second workspace of the same vendor — and run the next sync.
3. Look for the second source's invoices in the AI cost metrics.

**Expected:** every connected account's invoices are picked up on their own, whatever the other accounts have already synced.
**Actual:** the second account's invoices are absent at step 3, and later syncs never bring them in. They are discarded before the shared invoice table is reached.

Both models decide what is new by comparing against the newest row in the whole table, rather than the newest row for that account:

```sql
-- silver/ai/class_ai_invoice.sql
WHERE _version > (SELECT max(_version) FROM {{ this }})

-- connectors/ai/claude-team-invoices/dbt/claude_team__ai_invoice.sql
_airbyte_extracted_at > (
    SELECT coalesce(max(collected_at), toDateTime64('1970-01-01 00:00:00', 3))
    FROM {{ this }}
)
```

A test that adds a second account's invoices, read before everything already in the table, shows it — and each of the two models matters on its own:

| Comparison scope | Result |
|---|---|
| both models compare per account | 13 passed |
| shared table whole-table, per-connector model per account | `assert [] == ['in_SECOND']` |
| shared table per account, per-connector model whole-table | `assert [] == ['in_SECOND']` |

## Additional information

- Nothing is missing today. Only one account can be connected so far, so the whole-table maximum happens to be that account's own. The loss starts with the second one.
- Serving more than one vendor is the point of this table: it describes itself as "Unified vendor invoices across AI vendors", and the Claude Team model documents itself as the first contributor, defining the shape later vendors follow.
- A second workspace of the same vendor hits this too, not only a second vendor. Two instances of one connector write into the same staging table, which is why `insight_source_id` is part of every `unique_key` and `check_connection` refuses an empty one with "would collide unique_keys across connector instances".
- Nothing reports the loss. The sync succeeds, the build succeeds, and the invoices are simply absent.
- The three other AI tables of this kind (`class_ai_dev_usage`, `class_ai_assistant_usage`, `class_ai_overage`) already compare per account, via `silver_incremental_watermark`. These two are the only AI models left comparing over the whole table.
- Observed on the `claude-team-invoices` path as merged by #2432 (2026-08-18).

## Notes

related to #2607, #2432, #2700, #2855

Gregory91G:
Closes #2620.

**Why.** A month that prices more than one tier emits no `seat_cost_usd`: the invoice names its tier with the vendor's catalogue identifier, the seat with one from another API, and nothing says they are the same tier.

**What changed.** The connector carries Stripe's `price`/`product` ids into bronze, the class contract gains `tier_ref` for the vendor's stable tier identifier, and gold resolves a seat's price by joining it through `config.ai_seat_tier_map`. The display name stays in the vendor-extras blob.

**The binding is operator-authored configuration, not code.** Its keys are one account's catalogue identifiers, which a public repo must not carry. Ownership follows `config.task_value_map` — dbt guarantees the relation, an operator writes the rows.

**An unbound month still prices when it leaves no ambiguity.** One priced tier prices every seat billed in it, unchanged — so the table ships empty and no existing install moves until a binding is written.

**A binding also names the seat instance its prices reach.** The invoice and the seats arrive through separate connectors whose `source_id` never matches, so a tenant running two organisations of one vendor cannot be resolved from the data alone.

**Evidence.** The seat-cost fixture spelled a seat's tier with the invoice's own word for its plan and set a `tier_label` the vendor never sends — the invented agreement is why its assertions passed. Both gone; the connector suite goes 105 → 111 tests.

**Verified.** `dbt parse` clean, 111 connector tests pass, three seat-cost fixtures resolve and schema-validate, and the new price selection was run against ClickHouse across its three branches. Ruff clean at the pinned v0.15.21; the `connectors-ddl` snapshot carries the three new columns.

## Summary by CodeRabbit

* **New Features**
  * Added stable product and price identifiers to AI invoice data.
  * Added configurable mappings between invoice tiers and seat populations.
  * Added support for separate tier pricing across connector instances.

* **Bug Fixes**
  * Improved AI seat-cost attribution when invoices contain multiple tiers.
  * Unmapped or ambiguous tiers are no longer assigned an incorrect price.
  * Preserved correct handling for missing pricing metadata and partial invoices.

* **Tests**
  * Expanded end-to-end coverage for mapped, unmapped, partial, and multi-instance pricing scenarios.

### DIFF
--- a/src/ingestion/silver/task-tracking/schema.yml
+++ b/src/ingestion/silver/task-tracking/schema.yml
@@ -7,11 +7,14 @@ sources:
   # model materializing over them would wipe the decisions they hold. Declared
   # in the task-tracking domain rather than under one connector, because every
   # task source and gold itself read them.
+  # One `config` source serves every domain — a dbt source name is declared once
+  # project-wide, so the AI binding is listed here beside the task ones.
   - name: config
     schema: config
     tables:
       - name: task_field_roles
       - name: task_value_map
+      - name: ai_seat_tier_map
 
 models:
   - name: class_task_field_history
--- a/src/ingestion/silver/task-tracking/schema.yml
+++ b/src/ingestion/silver/task-tracking/schema.yml
@@ -7,14 +7,11 @@ sources:
   # model materializing over them would wipe the decisions they hold. Declared
   # in the task-tracking domain rather than under one connector, because every
   # task source and gold itself read them.
-  # One `config` source serves every domain — a dbt source name is declared once
-  # project-wide, so the AI binding is listed here beside the task ones.
   - name: config
     schema: config
     tables:
       - name: task_field_roles
       - name: task_value_map
-      - name: ai_seat_tier_map
 
 models:
   - name: class_task_field_history
--- a/src/ingestion/connectors/ai/claude-team-invoices/dbt/claude_team__ai_invoice.sql
+++ b/src/ingestion/connectors/ai/claude-team-invoices/dbt/claude_team__ai_invoice.sql
@@ -73,6 +73,10 @@ SELECT
     chain_status,
     category,
     tier_label,
+    -- The vendor's own stable identifier for the tier this line prices. A
+    -- contract column, not an extra: gold resolves a seat's price through it,
+    -- and unlike the display name it survives a rename or a localisation.
+    product_id                                          AS tier_ref,
     toUInt8(coalesce(is_proration, false))              AS is_proration,
     coalesce(currency, invoice_currency, 'usd')         AS currency,
     toStartOfMonth(toDateTime(toInt64(coalesce(period_start_ts, invoice_created_ts, 0)))) AS period_month,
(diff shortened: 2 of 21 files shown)


src/ingestion/silver/task-tracking/schema.yml
```diff
@@ -7,11 +7,14 @@ sources:
...
   # in the task-tracking domain rather than under one connector, because every
   # task source and gold itself read them.
+  # One `config` source serves every domain — a dbt source name is declared once
+  # project-wide, so the AI binding is listed here beside the task ones.
   - name: config
     schema: config
     tables:
       - name: task_field_roles
       - name: task_value_map
+      - name: ai_seat_tier_map
```
aleksdotbar: Not clear why this is here. Please verify this is not a mistake.

aleksdotbar (changes requested):

src/ingestion/silver/task-tracking/schema.yml
Gregory91G: Yes, that should be in it's own place, thanks

aleksdotbar (approved):

Issue #1886, opened 2026-07-23
Jira enrich crashes on every real sync (FINAL on a MergeTree bronze table) — no task metrics ever appear


ktursunov:
## Summary
On a real Airbyte-synced install, the jira ingestion pipeline fails at the **enrich** step every run. The `jira-enrich` binary queries `bronze_jira.jira_issue` with `FINAL`. Airbyte writes that table as a plain `MergeTree`. ClickHouse 25.7 rejects `FINAL` on `MergeTree`, so enrich exits 1. The pipeline aborts before dbt builds silver/gold. No jira/task metrics are ever produced, so the dashboard shows nothing for the person.

## Impact
No jira or task metrics on any real install. The Airbyte sync lands bronze data fine, but the pipeline never reaches silver/gold because enrich dies first.

## Steps to Reproduce
1. Run the jira ingestion pipeline (Airbyte sync → enrich → dbt).
2. The sync succeeds — `bronze_jira` is populated.
3. The enrich step fails with:
   ```
   Error: ClickHouse error: Code: 181. DB::Exception: Storage MergeTree doesn't support FINAL. (ILLEGAL_FINAL) (version 25.7.5.34)
   ```
4. The pipeline aborts. dbt silver/gold never runs. The dashboard has no jira metrics.

**Expected:** enrich reads bronze, materializes silver, and the pipeline continues to dbt.
**Actual:** enrich crashes because `bronze_jira.jira_issue` is a `MergeTree`, and `FINAL` is only valid on `ReplacingMergeTree`.

Every `bronze_jira` table is a plain `MergeTree`:
```
bronze_jira.jira_issue           MergeTree
bronze_jira.jira_issue_history   MergeTree
bronze_jira.jira_comments        MergeTree   … (all 14 tables)
```

## Root Cause
[...]

PR #2882, opened 2026-08-26
feat(ingestion): bronze shape comes from Airbyte — append_dedup, retire RMT promotion

### LINKED ISSUE #2877: Bronze shape should come from Airbyte: switch connections to append_dedup, retire RMT promotion
## Problem

Bronze tables get their `ReplacingMergeTree` shape from us instead of from Airbyte, through a three-part machinery that exists only because of a constraint whose technical justification no longer holds:

- Connections are forced to `destinationSyncMode=append` (`reconcile-connectors/python/normalize_catalog_to_append.py`, `lib/airbyte.sh` — tagged `cpt-dataflow-constraint-airbyte-append`) with the rationale "append_dedup buffers all records in memory until stream COMPLETE, OOMs on large streams, and loses all data on mid-stream pod death".
- Because append leaves plain `MergeTree`, every connector carries a `<connector>__bronze_promoted` bootstrap model (~134 of them) running `promote_bronze_to_rmt` (CTAS + `EXCHANGE`) to convert bronze to `ReplacingMergeTree(_airbyte_extracted_at) ORDER BY unique_key`.
- The connectors-ddl snapshot enshrines that post-promotion shape (including `allow_nullable_key = 1` for the Nullable `unique_key`).

The OOM rationale described the legacy destination. `destination-clickhouse` 2.x (Bulk CDK, direct-load) performs **no destination-side dedup work at all**: `ClickhouseAirbyteClient.upsertTable` throws `NotImplementedError("We rely on Clickhouse's table engine for deduping")`. In 2.x, `append_dedup` uses the identical append-only insert path; the only difference is the created table is `ReplacingMergeTree(version) ORDER BY <primary key>` — exactly the shape our promotion manufactures. There is no buffering, no COMPLETE-gated flush, and no additional loss mode. The defining spec artifact for the constraint no longer exists in the tree (the ID is dangling); the claim survives only in a script comment.

## Proposal

Make Airbyte's table shape canonical for bronze; keep pre-creation as an existence mechanism only.

1. Reconcile builds catalogs with `destinationSyncMode=append_dedup` and `primaryKey=[["unique_key"]]` per stream (rename `normalize_catalog_to_append.py` accordingly). Airbyte then creates bronze as `ReplacingMergeTree(version) ORDER BY unique_key` itself.
2. Delete all `<connector>__bronze_promoted` models and the `promote_bronze_to_rmt` macro; drop the corresponding conventions checks (architecture rule "Bronze tables MUST be promoted", check-dbt-conventions checks 4–5) and replace the append-only rule with the new contract.
3. Regenerate the connectors-ddl snapshot with the new catalog. Since bootstrap-db generation already runs the real destination image to create tables, the dumped DDL becomes byte-identical to what Airbyte would create at first sync — pre-creation degenerates to replaying Airbyte's own DDL earlier in time, and `CREATE TABLE IF NOT EXISTS` no-ops cleanly in both directions.
4. Retire the dangling `cpt-dataflow-constraint-airbyte-append` reference (replace with a constraint documenting the new contract).

### What explicitly stays (out of scope to remove)

- **Bronze pre-creation via the snapshot.** Two load-bearing consumers are independent of who decides the engine:
  - Deploy-time builds: staging models read bronze and silver unions are positional across all connectors, and the migrate hook's dbt gold build runs before any sync — relations must exist at deploy.
  - **The test harness seeds bronze directly with artificial data** (metric/e2e tests insert fixture rows into bronze tables; Airbyte is not part of that stack), so the tables must exist before any seed runs. Fixtures intentionally include duplicate rows to exercise read-time dedup — they depend on the RMT semantics as well as on the tables existing.
- `reconcile_bronze_schema.py` (deploy-window column healing, #1991) and the connectors-ddl CI convergence gate (#1892-class type drift stays a reviewed diff).

Cluster/replicated-mode concerns are deliberately **not** part of this issue — they are tracked in #2010. (Side effect worth knowing: with this change, #2010's bronze handling reduces to destination configuration.)

## Prerequisites

1. **`unique_key` must be NOT NULL.** The 2.x destination strips nullability from primary-key columns (ReplacingMergeTree cannot safely dedup on a nullable key — the reason bronze currently carries `allow_nullable_key=1`). The `{tenant}-{source}-{natural_key}` formula should never yield null; audit all connectors. Known gap: `claude-admin` bronze has no `unique_key` column at all (the tracked promotion exception) and must gain one first.
2. **Version-column semantics per stream.** We pin `_airbyte_extracted_at` as the RMT version; 2.x uses the stream cursor when its type qualifies (`Int64`/`Date`/`DateTime64`) and falls back to `_airbyte_extracted_at` otherwise. Most cursors are ISO-string timestamps (non-qualifying, so the fallback applies), but this needs a per-stream check — cursor-as-version changes which duplicate wins.
3. **Warm-table convergence.** Against an existing Nullable-key table, the 2.x destination detects the primary-key nullability change and performs its own rebuild (temp table + copy + `EXCHANGE`). Verify on a disposable instance that the first `append_dedup` sync converges today's tables without data loss, and measure the one-time rebuild cost per table.
4. **Test fixtures.** Bronze seed data must satisfy NOT NULL `unique_key`; audit metric/e2e fixtures (including the intentional-duplicate cases) against the new shape.

## Acceptance

- Fresh install: bronze tables created by Airbyte (or pre-created identically by the snapshot) as `ReplacingMergeTree ORDER BY unique_key`, no promotion step anywhere; full e2e green including duplicate-dedup metric tests.
- Warm install: first sync after the switch converges existing bronze tables; no domain freeze, no data loss.
- Zero `__bronze_promoted` models in the tree; conventions/skill/docs updated; snapshot regeneration produces no diff against what the destination creates.

cyberantonz:
Implements #2877 — bronze table shape moves from our promotion machinery to the Airbyte destination itself.

## What changed

- **Reconcile** builds catalogs with `destinationSyncMode=append_dedup` + `primaryKey=[["unique_key"]]` for every stream whose schema carries `unique_key` (plain `append` fallback + stderr warning for keyless streams). `normalize_catalog_to_append.py` → `normalize_catalog.py`; the `ab_create_connection` contract comment and the dev-tool `generate-catalog.sh` follow the same rule (`overwrite` is gone everywhere).
- **bootstrap-db generation** mirrors the same catalog rule and drops its promotion step, so the committed snapshot is byte-identical to what a real sync creates.
- **Promotion machinery deleted**: `promote_bronze_to_rmt` macro, all 21 `<connector>__bronze_promoted` bootstrap models, ~120 `-- depends_on` lines, schema.yml entries, README/comment references. The obsolete #1886 promotion-selector regression test goes with the model it guarded.
- **Snapshot regenerated** (full bootstrap pipeline + `dump-ddl.sh`): `unique_key` becomes NOT NULL `String` on all 162 bronze tables (the destination strips PK nullability), `allow_nullable_key` is dropped, engines/ORDER BY otherwise byte-identical — and the two previously-unpromoted tables (`bronze_claude_team.claude_team_invites`, `bronze_zendesk.support_ticket_ids`) converge to the uniform `ReplacingMergeTree(_airbyte_extracted_at) ORDER BY unique_key` shape.
- **`audit_rmt_read_dedup.py`** now treats every `bronze_*` source read as RMT (behavior verified identical against main: same advisory findings) + pathlib/ruff cleanup.
- **Conventions updated**: architecture rules 4 & 8, check-dbt-conventions checks 4 & 5, connector-authoring validate workflow.

## Why it's safe

All four prerequisite audits/verifications are recorded on #2877 (no connector emits a null `unique_key`; every stream's version column falls back to `_airbyte_extracted_at` exactly as promotion pinned; all 1220 test fixture rows already comply). Warm environments converge on the first sync after the switch: the destination detects the primary-key nullability change and rebuilds the table itself (temp + copy + `EXCHANGE`) — verified lossless on a disposable ClickHouse 25.7.5 with the stock destination image.

- `dbt parse` green; field parity: 0 failures; pre-commit green.
- The connectors-ddl workflow re-runs the exact generation pipeline and gates on snapshot convergence.

Closes #2877


ktursunov (approved):

hello1101n:
Are we sure we want to do it in the current release?

hello1101n (changes requested):
let's review

PR #2902, opened 2026-08-27
feat(ingestion): add the Compass connector for service ownership

### LINKED ISSUE #2708: Connector: Atlassian Compass + Teams — service catalog, ownership and deployment events
## Summary

Add an ingestion connector for **Atlassian Compass** (an internal developer portal: a catalog of software components with owners, dependency links and scorecards) together with the **Atlassian Teams** directory that Compass uses for component ownership. One connector covers both.

This gives Insight a **service-ownership dimension it currently lacks**: component → owning team → repository, joining the catalog to the git sources we already ingest.

Design spec (lands with the implementation PR): `src/ingestion/connectors/dev-portal/compass/SPEC.md`

## Why

- Component ownership is enforced in Compass by scorecards, so the ownership signal is dense there. Jira's Team field references the *same* directory entity by the same team UUID, but how densely it is filled is a per-site choice — ownership attribution should be built on Compass, not on that field.
- `REPOSITORY` links on components join to `class_git_repositories.full_name`, the only repo key common to the github/gitlab/bitbucket connectors.
- Scorecards give a per-component health/completeness score with per-criterion detail.

## Scope

Streams (grain in brackets):

- `components` [component] — id, name, type, `ownerId`, labels, links, event sources
- `component_links` [link] — flattened from the component sweep; `REPOSITORY` is the git join
- `component_relationships` [dependency edge]
- `scorecards` [definition + criteria]
- `component_scorecard_scores` [component × scorecard] — total, status, per-criterion detail
- `teams` [team], `team_members` [team × person]
- `deployment_events` [event] — the only incremental stream

Out of scope for v1: metric time series, `PUSH`/`BUILD`/`INCIDENT` events (duplicate the git connectors), Atlassian groups (an ACL primitive with no typed link to teams), team hierarchy (extra experimental opt-in).

## Placement

| Item | Value |
| --- | --- |
| Path | `src/ingestion/connectors/dev-portal/compass/` |
| Category | `dev-portal` — **new category**; also needs adding to the connector skill's category list |
| Type | `nocode` (declarative GraphQL-over-POST, same shape as `git/github-directory`) |
| Namespace | `bronze_compass` |
| Secret | `atlassian_instance_url`, `atlassian_email`, `atlassian_api_token` |

Deliberately **not** merged into the Jira connector: the connector boundary is one descriptor (single version / schedule / namespace / `dbt_select` / secret), so merging would couple a tiny full-refresh source to a heavy tracker's blast radius, mis-tag rows by `data_source`, and force descriptor bumps (full-refresh risk per ADR-0015) for unrelated changes. Same reasoning `git/github-directory` records for splitting from `git/github-v2`.

## Transport notes that shape the manifest

- Single gateway `POST https://api.atlassian.com/graphql`, Basic auth (account email + API token) — no OAuth needed. `cloudId` from `<instance>/_edge/tenant_info`.
- **GraphQL errors arrive as HTTP 200** with `errors[]` → needs `HttpResponseFilter` on the body, as in `git/github-directory`. Most fields also return a union with `QueryError`, which is a *successful* body carrying a business error.
- Operation names are mandatory (anonymous queries are being phased out).
- Cost-based per-user rate limiting with `429` above the budget.
- **Three different traversal mechanisms coexist** — a plain `first`/`after` cursor; pagination nested inside a `query:` input object; and a windowed lookback (`filter: {startFrom, periodicity}`) for score history where the `after` cursor is inert and `hasNextPage` stays true forever. A paginator that trusts `hasNextPage` there loops indefinitely. Also: the bulk team-resolve form returns truncated member lists with a lying `hasNextPage`; per-team pagination is exact.
- `component_scorecard_scores` (and the optional history backfill) require a **field-level** `@optIn(to: "compass-beta")` directive — the header form does not work. Every other stream is on stable fields, so this one can be dropped without affecting the rest if beta churn becomes a problem.

## History strategy

The API emits **no change events and no field-level audit** — scorecard definitions expose only a single "someone changed something" timestamp. Bronze is append-only, so full-refresh streams accumulate dated snapshots for free; silver materialises the diffs that matter:

- SCD intervals for `components.ownerId`, `REPOSITORY` links and `team_members` — these are attribution axes, and without time bounds a re-assignment silently rewrites historical attribution.
- Dated snapshots of **scorecard definitions**: criterion weights renormalise to 100, so a score drop is ambiguous between "the service degraded" and "a criterion was added". Only a same-date definition snapshot separates them.
- `component_scorecard_scores` snapshots are never collapsed — the series is the deliverable.

Asymmetry to plan around: score history is backfillable from the API (one bootstrap run), definition history is not.

## Risks / caveats to carry into implementation

1. **`DEPLOYMENT` does not imply a release.** Any tool can publish events of that type through an external event source, including build/artifact events mapped onto it with a non-production `environment.category`. Release-frequency and lead-time metrics must be keyed on `environment.category = PRODUCTION`; anything else is build or pre-production activity and must not be surfaced as releases. Verify the category distribution before building a DORA-shaped surface.
2. **The event feed is a rolling window (~2 weeks) with no backfill and no usable server-side date filter** (`timeParameters` rejects every window tried). Incremental must be client-side on `lastUpdated`; a missed sync loses events permanently, so this stream sets the daily-cadence floor and dominates request volume (`O(N_components)`).
3. **Scorecard scores measure integration coverage as much as health** — metric-backed criteria score zero where the feeding integration is absent, which is indistinguishable from genuine failure at the score level. Consumers should restrict to Compass-intrinsic criteria or filter on `dataSourceLastUpdated`.
4. **`User` has no `email` field** in the Teams API — the people join key is the Atlassian account id. If our people relation does not carry account ids, the component → team → person chain stops at the team.
5. **Team membership is many-to-many** — a person can be in several teams, so per-person aggregation through membership needs an explicit apportionment rule or it double-counts.

## Open questions

- [ ] Category name `dev-portal` — confirm, and extend the connector skill's category list
- [ ] Teams in silver: extend the existing people/org families or a separate `class_team_*`? Needs the identity owners' call — a third representation of org structure is a maintenance risk
- [ ] Does our people relation carry Atlassian account ids? (blocks the person-level chain)
- [ ] `environment.category` distribution — decides whether `deployment_events` can support release metrics at all
- [ ] Does editing a scorecard criterion preserve its id? If not, "threshold retuned" is indistinguishable from "criterion replaced" in a snapshot diff (verify on a throwaway scorecard)

## Delivery checklist

- [ ] `descriptor.yaml` — strict semver `1.0.0`, `type: nocode`, namespace, schedule, secret fields, `dbt_select`
- [ ] `connector.yaml` — named operations, `errors[]` response filters, per-stream traversal
- [ ] `connectors-config.yaml` entry (without it the bronze database is never created)
- [ ] `class_people.sql` `depends_on` entry, if any stream feeds people
- [ ] connectors-ddl snapshot regenerated per the bootstrap-db README
- [ ] per-stream mock tests — 100% stream coverage, including a `QueryError`-in-200 case and a `hasNextPage`-lies case
- [ ] `scripts/ci/connector_wiring.py` green
- [ ] dbt conventions (RMT engine, `order_by=['unique_key']`, read-time dedup, `delete+insert` on silver)

mitasovr:
Closes #2708.

**Why.** Nothing in Insight maps a repository to the team that owns it. Compass holds that mapping, and its owners are the same directory entities Jira's Team field points at.

**What changed.** A new `dev-portal/compass` nocode connector reads six streams — component catalog, scorecard definitions and results, the Atlassian Teams directory with membership, deployment events — from one GraphQL endpoint. Bronze-only; `dbt/` carries just the bronze→RMT promotion.

**Teams ship inside this connector, not their own.** Same host, credential and cadence; silver reads bronze tables rather than connectors, so later consumers still join them. Reversible by moving two streams.

**Not folded into the jira connector.** A descriptor is one version, schedule, namespace and secret, so folding would couple a tiny full-refresh source to a heavy tracker's blast radius — the rule `git/github-directory` records.

**The requester filters on the response body, and every paginator path is null-safe.** This API answers a rejected query with HTTP 200 plus `errors[]`, and returns unions where a `QueryError` member has no `pageInfo`. Both bit during live reads; SPEC.md carries the rest.

**Also fixes a wiring gap.** `collaboration/zoom` and `git/github-directory` ship mock suites but were missing from `connector-mock-tests` paths, so changing either did not re-run its own tests.

**Out of scope.** Silver — it needs a service-ownership concept silver lacks, plus the identity owners' call on a third org-structure representation. SPEC.md §7; tracked in #2708.

**Verified.** Both manifest validators pass, `connector_wiring.py` clean, 52 mock tests green, all six streams read against a live site. No deploy attempted.

## Summary by CodeRabbit

* **New Features**
  * Added an Atlassian Compass connector for components, scorecards, teams, team members, and deployment events.
  * Added daily synchronization, pagination, authentication, incremental deployment-event updates, and bronze data promotion.
  * Added connection configuration and example secret manifests.

* **Documentation**
  * Added setup, authentication, stream, synchronization, and data-model documentation.

* **Tests**
  * Added comprehensive mock coverage for pagination, retries, error handling, transformations, and incremental synchronization.

### DIFF
--- a/src/ingestion/connectors/dev-portal/compass/descriptor.yaml
+++ b/src/ingestion/connectors/dev-portal/compass/descriptor.yaml
@@ -1,0 +1,23 @@
+# Atlassian Compass — service catalog, ownership, scorecards, deployment events.
+#
+# Bronze-only for now: no silver models beyond the bronze→RMT promotion, so
+# `dbt_select` currently resolves to that promotion view alone. Silver class
+# families are sketched in SPEC.md §7 and deliberately deferred.
+name: compass
+type: nocode
+# Strict semver per ADR-0015 — three components, no leading zeros.
+version: "1.0.0"
+# Daily is a floor, not a preference: the Compass event feed is a rolling
+# window with no backfill, so a skipped run loses deployment events for good.
+schedule: "0 2 * * *"
+workflow: sync
+dbt_select: "tag:compass+"
+
+connection:
+  namespace: "bronze_compass"
+
+secret:
+  required_fields:
+    - atlassian_email
+    - atlassian_api_token
+    - atlassian_cloud_id
--- a/scripts/ci/components.py
+++ b/scripts/ci/components.py
@@ -239,10 +239,17 @@ COMPONENTS = [
         "pytest_args": "--suites-only",
         "cover": False,
         "triggered_by": ["connector-tests-harness"],
+        # Every nocode connector that ships a tests/ suite belongs here, or a
+        # change to that connector does not re-run its own mock tests. CDK
+        # connectors with suites (hubspot, gitlab, bamboohr, …) are covered by
+        # their own components instead.
         "paths": [
             "src/ingestion/connectors/task-tracking/jira",
             "src/ingestion/connectors/git/github",
             "src/ingestion/connectors/git/bitbucket-cloud",
+            "src/ingestion/connectors/git/github-directory",
+            "src/ingestion/connectors/collaboration/zoom",
+            "src/ingestion/connectors/dev-portal/compass",
         ],
     },
     # `src/frontend/helm` falls under this path but has no measured lines, so it
--- a/src/ingestion/connectors/dev-portal/compass/README.md
+++ b/src/ingestion/connectors/dev-portal/compass/README.md
@@ -140,6 +140,13 @@ this data has to respect them.
   component has exactly one owner.
 - **`teams.member_count` is not populated** by the search field this stream
   uses; count `team_members` rows instead.
+- **A failed catalog query fails the sync; one unreadable component does not.**
+  Most fields return a union, so a refusal arrives as a *successful* body with a
+  `message` and no rows. Where an empty answer would be indistinguishable from
+  "there is nothing" — the catalog, the scorecards, the team directory and its
+  membership — the connector fails loudly. A single component that vanished
+  between the catalog sweep and its event read is skipped instead, and the other
+  components continue.
 - **`component_scorecard_scores` rides an experimental field** and needs the
   `@optIn(to: "compass-beta")` directive. Every other stream is on stable
   fields, so if beta churn breaks it, it can be dropped without affecting the
--- a/src/ingestion/connectors/dev-portal/compass/README.md
+++ b/src/ingestion/connectors/dev-portal/compass/README.md
@@ -1,0 +1,157 @@
+# Compass Connector
+
+Extracts the Atlassian Compass service catalog — components, their owning
+teams, repository links, dependency edges, scorecard results and deployment
+events — plus the Atlassian Teams directory those owners come from. Everything
+is read from one GraphQL endpoint with Basic auth (Atlassian account email +
+API token).
+
+Design rationale, traversal contracts and the history strategy live in
+[SPEC.md](SPEC.md). Read it before changing a query — several fields in this API
+behave in ways that a reasonable reading of the schema does not predict.
+
+## Prerequisites
+
+1. An Atlassian account with **Compass product access** on the site.
+2. An API token for that account: id.atlassian.com → Security → API tokens.
+   No OAuth application and no additional scopes are needed; the Teams streams
+   work on the same credential.
+3. The site's **cloud id**, which every Compass field requires:
+
+   ```bash
+   curl -s https://<your-site>.atlassian.net/_edge/tenant_info
+   ```
+
+   It is not discoverable through the GraphQL gateway, so it is configuration
+   rather than something the connector can look up.
+
+## K8s Secret
+
+```yaml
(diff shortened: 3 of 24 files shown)


src/ingestion/connectors/dev-portal/compass/descriptor.yaml
```diff
@@ -0,0 +1,23 @@
...
+schedule: "0 2 * * *"
+workflow: sync
+dbt_select: "tag:compass+"
+
+connection:
+  namespace: "bronze_compass"
+
+secret:
+  required_fields:
+    - atlassian_email
```
aleksdotbar, line 21: Did you double check that there is no way to make it work without email?

aleksdotbar (approved):

PR #2962, opened 2026-08-30
feat(analytics): generate reports in the backend

aleksdotbar:
**Why.** Browser-generated CSV and XLSX exports require fetching all selected metric data and do not cleanly support both person-scoped and tenant-scoped metrics.

**What changed.** Analytics now validates report recipes, hydrates roster attributes, batches metric queries, and produces CSV or XLSX downloads through `/v1/reports`. The frontend uses preview and export APIs instead of building files locally.

**Synchronous generation is deliberate.** Concurrency, query, output, and temporary-storage limits control work without introducing an async job system.

**Report rows use positional values with column metadata.** Internal person IDs remain outside downloaded reports.

Synthetic ClickHouse benchmark: 100 metrics, 500 people, five monthly periods, and two concurrent XLSX reports produced 520,000 cells in 3.8 seconds.

**Out of scope.** Queued asynchronous report generation remains future work.

**Verified.** Backend tests and Clippy; frontend tests, typecheck, and lint; Helm lint; OpenAPI/schema drift; synthetic ClickHouse benchmarks. Full stand route suite was unavailable.


aleksdotbar:
Superseded by the reviewable stack in #2970, #2971, #2968, and #2969.

PR #2971, opened 2026-08-31
feat(analytics): preview reports

aleksdotbar:
**Why.** Report preview needs one backend contract that supports both person-scoped metrics and tenant-scoped metrics without reproducing analytics logic in the browser.

**What changed.** Adds `POST /v1/reports/preview` with recipe validation, visible-profile hydration, dynamic columns, period planning, bounded query batches, and positional response rows.

**Subject determines metric compatibility.** People and tenant metrics cannot be mixed; frontend tabs are not part of the API recipe.

**Profile attributes remain optional.** Analytics includes populated provider-backed attributes and never adds internal person IDs as report columns.

**Stack.** 2 of 4; depends on #2970 and is followed by #2968.

23 focused report tests passed.

**Out of scope.** File export and frontend adoption remain in downstream PRs in this stack.

**Verified.** Focused report tests and Clippy; analytics OpenAPI and generated-schema drift checks.

### DIFF
--- a/docs/components/backend/analytics/openapi.json
+++ b/docs/components/backend/analytics/openapi.json
@@ -2221,6 +2221,169 @@
         ],
         "type": "object"
       },
+      "ReportColumnDataType": {
+        "enum": [
+          "text",
+          "date",
+          "number"
+        ],
+        "type": "string"
+      },
+      "ReportColumnMetadata": {
+        "properties": {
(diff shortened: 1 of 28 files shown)


aleksdotbar:
Combined into #2968 because preview and export are one backend reporting feature.

Issue #3029, opened 2026-09-01
Claude Code and Codex daily usage can be recorded against the previous day


Gregory91G:
## Summary
A day's per-person AI usage can be stored under the previous date, so daily AI cost and coding activity read high on one day and low on the next, and never correct themselves.

**Version:** `insight-0.6.40` (appVersion `2026.09.01.14.15-2c9a5fe`)
**Reproduction rate:** 3/3 for the request windows (offline, no vendor call needed)
**State:** a connector that has already synced at least once, so its cursor resumes from stored state.

## Steps to Reproduce
1. Take the `day_cursor` parameters from `src/ingestion/connectors/ai/chatgpt-team/connector.yaml` — `start_datetime` = today−7, `end_datetime` = today−1, `step: P1D`, `cursor_granularity: P1D`, `lookback_window: P2D`.
2. Drive the CDK's own slicing engine with them (`airbyte-cdk==7.0.4`, `airbyte_cdk.sources.streams.concurrent.cursor.ConcurrentCursor`; `end_datetime` maps to `end_provider`, `step` to `slice_range`) and print every `(start_time, end_time)` it yields.
3. Every window covers one calendar day except the last, which covers two. With stored state at today−2 it is the only window the run produces.

**Expected:** every request window covers exactly one day. The comment above the cursor states the API treats `start_date`/`end_date` as an inclusive single-day filter.
**Actual:** the last window of every run covers two days.

The vendor does not return a date inside the per-user object. Both manifests add one with an `AddedFieldDefinition` that reads `{{ stream_interval.start_time }}` — the first day of the window — into `date` (chatgpt-team) and `metric_date` (claude-team), and into `unique_key`.

## Additional information
- Windows printed for three cursor shapes, same start and end:

  | cursor | windows |
  |---|---|
  | `end_datetime` set, no `lookback_window` | six one-day windows, then one covering two days |
  | `end_datetime` set + `lookback_window: P2D` (chatgpt-team today) | six one-day windows, then one covering two days |
  | no `end_datetime` + `lookback_window: P2D` (claude-team since manifest 3.1.0) | all windows one day |

- The bound is pinned to a calendar midnight. Without it the window ends at the read time, so the last one stays inside the current day.
[...]

PR #3030, opened 2026-09-01
fix(chatgpt-team): keep the day cursor request window inside one day

Gregory91G:
Refs #3029 — deliberately not `Closes`.

**Why.** A day's per-person AI usage could be stored under the previous date. The date is a label the connector stamps from the request window's start, and the shared `day_cursor`'s final window spanned two days.

**What changed.** The cursor loses its end bound, so every request window stays inside one calendar day; the daily step and the two-day lookback are untouched. Both connectors gain a regression suite over the real declarative cursor — claude-team's bound was removed earlier and is now pinned against coming back.

**The issue stays open on purpose.** This branch stops new misdated rows; the rows already stored are cleared operationally on the affected installations rather than by product code, so a one-off cleanup does not become permanent machinery in the medallion path.

**Verified.** 52 passed across both suites. Restoring the end bound turns 20 of them red, naming the two-day window.


Gregory91G:
@coderabbitai review

hello1101n (approved):

PR #3112, opened 2026-09-02
fix(frontend): say how long a running sync has been going, not that it took no time

### LINKED ISSUE #3036: Connector health shows near-zero duration for a sync that is still running
## Summary
On the Connector health page, a connector shown as "syncing" reports a Duration of a few seconds
or less regardless of how long ago its sync actually started, so an admin cannot tell a fresh sync
from one that has been running for hours.

**Version:** unknown — could not reach the instance's cluster from this session to confirm via
`helm list`; the page's own header read "Last checked 14 min ago" at capture time
**Reproduction rate:** 1/1 (one page load; two independent connectors showed the anomaly in that
same read)
**State:** an admin viewing the Connector health page while at least one connector's sync is in
progress.

## Steps to Reproduce
1. As an admin, open Manage → Connector health.
2. Find a row whose State is "syncing" and whose "Last sync started" timestamp is more than a few
   minutes before the page's "Last checked" time.
3. Read that row's Duration.

**Expected:** Duration reflects roughly how long the sync has been running so far (growing over
time while it stays in "syncing"), or is shown as not yet known.
**Actual:** Duration shows a small, effectively frozen value that does not reflect elapsed time.

## Additional information
- Row observed: connector `bitbucket-cloud`, State `syncing`, Last sync started
  `2026-09-01 15:12:33Z`, Duration `0 ms`, Records reported `357,723` — roughly 12 hours had
  elapsed since the start time shown, with the page itself dated "Last checked 14 min ago."
- Row observed: connector `jira`, State `syncing`, Last sync started `2026-09-02 03:00:34Z`,
  Duration `1.0 s`, Records reported `119,214`.
- Contrast, same page load: connectors in State `sync ok` show durations that scale with the
  connector and look plausible — `zoom` 2m 31s, `m365` 2m 10s, `compass` 6m 37s, `bamboohr`
  2m 19s, `confluence` 37m 16s, `slack` 2m 3s. Duration does populate with a real-looking value
  once a sync reaches a completed state; the anomaly is specific to rows still in `syncing`.

## Notes
Related to #2913 (Connector health page).

mozhaev-dev:
Closes #3036.

**Why.** A connector shown as `syncing` reported a Duration of `0 ms` however long ago it started, so a sync stuck for hours looked like the fastest row on the page.

**What changed.** The column read the mover's own reported duration for every row, and for a job it has not finished the mover reports a running total — a zero where it has none. The sweep now records no duration for a `pending` or `running` job, and the page derives the elapsed time for such a row from its start and the answer's own `as_of`.

**The page ignores a recorded duration for an in-flight row rather than trusting the writer.** Ledger rows written before this change still hold that zero, so the fix has to work without a backfill. Reversible on its own.

**Wording is "so far", not a bare duration.** Elapsed time and a finished measurement are different facts and share one column; the row says which it is instead of relying on the state badge beside it.

The `example-inbox` row, started 37 minutes before the page was dated, carrying the zero a ledger row written by the old sweep still holds:

| Before | After |
|---|---|
| before | after |

**Verified.** `vitest run --project=unit` (2352 pass, 9 new), `tsc -b`, `eslint --max-warnings 0`, and `python3 -m unittest discover -s src/ingestion/reconcile-connectors/tests` (87 pass, 6 new). Screenshots are the msw mock stand.

## Summary by CodeRabbit

* **Bug Fixes**
  * Connector health now displays live elapsed time for pending and running syncs instead of misleading zero-duration values.
  * Completed syncs continue to display their reported durations, including genuine zero-duration jobs.
  * Unavailable or invalid timing information is shown as unmeasured rather than a fabricated duration.

* **Improvements**
  * Active syncs are clearly distinguished from completed syncs.
  * Running syncs appear in health history with their start time and current status.

### DIFF
--- a/src/ingestion/reconcile-connectors/tests/test_sweep_plan.py
+++ b/src/ingestion/reconcile-connectors/tests/test_sweep_plan.py
@@ -1,13 +1,14 @@
 """What a tick decides to write.
 
-Run: python3 -m unittest discover -s src/ingestion/reconcile-connectors/tests
+Run: pytest src/ingestion/reconcile-connectors/tests
 """
 
 from __future__ import annotations
 
 import sys
-import unittest
 from pathlib import Path
+
+import pytest
 
 sys.path.insert(0, str(Path(__file__).resolve().parents[1] / "python"))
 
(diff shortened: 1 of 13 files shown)


ktursunov (approved):

src/ingestion/reconcile-connectors/tests/test_sweep_plan.py
```diff
@@ -372,6 +372,45 @@ def test_it_outranks_the_finished_job_it_follows(self) -> None:
...
+
+    def test_it_still_states_when_it_started(self) -> None:
+        """The elapsed time the page reports is derived from this stamp, so
+        withholding the duration must not withhold the start with it."""
+        planned = plan.sync_row(self.RUNNING, CONNECTORS, TICK)
+
+        self.assertEqual(planned["started_at"], "2026-08-27 08:00:30.000")
+
+
+class OnlyAnUnfinishedJobLosesItsDuration(unittest.TestCase):
```
ktursunov: Can we use only pytest - both unit and e2e tests?

ktursunov (changes requested):

ktursunov (approved):

PR #3125, opened 2026-09-02
docs(frontend): design for removing the legacy UI so the portal is the only interface

hello1101n:
**Why.** The pre-portal shell is still in the tree behind `insight.legacyShell`, a hatch only `tests/stand/ui/conftest.py` writes, and `/custom-metrics` and `/queries` have no portal home and no link pointing at them.

**What changed.** Adds the design: both consoles move into Manage, then the hatch, the old shell, four orphan routes and `tests/stand/ui/` go.

**Consoles ship gated as `planned`** via `frontend.nav.planned` in insight-gitops, all four clusters — they land without becoming visible. Reversible by dropping the entry.

**Portal automation scope stays open.** The spec records what the 13 removed stand journeys covered; rewriting them is separate work.

**Out of scope.** No code — delivery is two follow-up PRs here plus one in insight-gitops.

**Verified.** Docs only, nothing to run.

## Summary by CodeRabbit

- **Documentation**
  - Added a design specification outlining the planned transition to a portal-only interface.
  - Documents the proposed migration of metrics and query console surfaces into the Manage area, including navigation and access considerations.

### DIFF
--- a/docs/superpowers/specs/2026-09-02-remove-legacy-ui-portal-only-design.md
+++ b/docs/superpowers/specs/2026-09-02-remove-legacy-ui-portal-only-design.md
@@ -1,0 +1,223 @@
+# Remove the legacy UI — the portal is the only interface
+
+Design, 2026-09-02.
+
+## Goal
+
+Delete the pre-portal shell and its screens from the frontend, so the portal is
+the only interface the codebase can render. Move the two console surfaces that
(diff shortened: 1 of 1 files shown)


hello1101n:
Superpowers design docs do not belong in this repo — reopening as the implementation instead.

PR #3128, opened 2026-09-02
fix(frontend): order org-chart reports by display name

### LINKED ISSUE #3126: People pane org chart and scope picker list reports in arbitrary order
## Problem

In the People zone, WorkChart lists a manager's reports in the order the identity API returns them. The Scope picker in the topbar walks the same tree, so its manager list comes out in the same order. Neither order is alphabetical, so finding a name means scanning the whole list.

## Reproduce

1. Open the portal with an org-chart visibility policy → **People** → expand any manager in WorkChart.
2. Open **Scope** in the topbar.

**Expected:** siblings under each manager sorted alphabetically by display name, in both surfaces. The picker keeps its outline order (a lead, then that lead's leads); only siblings change.
**Actual:** siblings appear in person-id order — arbitrary to a reader.

## Root cause

- identity-resolution orders child edges by `insight_source_type, insight_source_id, child_person_id` (`current_children_for_parent` in `persons_repo.rs`) and the subchart builder keeps that order. Nothing sorts by name.
- The frontend preserves the wire order: `toIdentityPerson` in `src/frontend/src/api/identity-client.ts` maps `subordinates` as received, `PersonNode` in `org-tree.tsx` renders them verbatim, and the manager walk in `resolveScopeRoster` (`use-org-scope.ts`) pushes them verbatim.
- The flat-policy roster (`RosterList`) and the person header's peers list already sort with `personDisplayName(a).localeCompare(personDisplayName(b))`, so the tree and the picker are the two places that skipped it.

## Fix

- [ ] Sort `subordinates` by `personDisplayName` at each level in `toIdentityPerson`, so every consumer of the identity tree (WorkChart, Scope picker, Employees walk) shares one order and the two display sites need no local sort.
- [ ] Unit test in `identity-client.test.ts`: nested levels sorted; the display-name → username → email fallback is the sort key.

hello1101n:
Closes #3126.

**Why.** WorkChart lists a manager's reports, and the Scope picker its managers, in the order the identity API sends them — person-id order — so finding a name means scanning the whole list.

**What changed.** `toIdentityPerson` now sorts `subordinates` at every level by `personDisplayName`, the key the flat roster and the peers list already sort under, so every consumer of the identity tree shares one alphabetical sibling order. The picker keeps its outline order; only siblings move.

**Verified.** Frontend `pnpm test` (2366 passing), `pnpm lint`, `pnpm typecheck`. No screenshot: the mock fixture is already alphabetical, so a local run cannot show a before.

## Summary by CodeRabbit

* **Bug Fixes**
  * Subordinates are now consistently sorted by display name at every level.
  * When a display name is unavailable, sorting falls back to username and then email address.
  * Updated ordering improves consistency when viewing hierarchical person information.

### DIFF
--- a/src/frontend/src/api/identity-client.test.ts
+++ b/src/frontend/src/api/identity-client.test.ts
@@ -82,12 +82,12 @@ describe("getPerson", () => {
     const person = await getPerson("019e27bc-dec0-7626-81a9-c5524662a6aa");
 
     expect(person.subordinates.map((s) => s.person_id)).toEqual([
-      "019e27bc-dec0-7626-81a9-c5524662a6ab",
       "019e27bc-dec0-7626-81a9-c5524662a6a9",
+      "019e27bc-dec0-7626-81a9-c5524662a6ab",
     ]);
     expect(person.subordinates.map((s) => s.email)).toEqual([
-      "ic1@example.com",
       "",
+      "ic1@example.com",
     ]);
   });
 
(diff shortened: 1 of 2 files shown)


hello1101n:
Closing: the sort moves to identity-resolution so every client gets the same order as the roster listing. Backend PR to follow against #3126.

PR #3132, opened 2026-09-02
fix(identity): serve subordinates in label order

### LINKED ISSUE #3126: People pane org chart and scope picker list reports in arbitrary order

hello1101n:
Closes #3126.

**Why.** A manager's reports came back ordered by source and child person id, so WorkChart and the topbar scope picker both listed people in an order arbitrary to a reader.

**What changed.** `assemble_profile` and `assemble_person` now order their sibling nodes by the display label, case-insensitively, unlabelled people last. The key is the precedence the roster listing already pages by, so the tree and the listing agree and no client sorts for display.

**Ordered on the server, not in the UI.** Three surfaces read this tree, and each one sorting for itself is three chances to disagree with the listing. Reversible without touching a client.

Screenshots come from the local mock stand, whose fixture I temporarily reordered so each column shows what the UI does with the wire order named above it. The mock service worker answers in place of this service, so these prove the rendering, and the unit tests prove the ordering.

| Wire order by person id (before) | Wire order by label (after) |
|---|---|
|  |  |
|  |  |

**Out of scope.** `GET /v1/subchart` builds its forest in row order too; no client reads it today, so it is left alone.

**Verified.** `cargo test -p identity-resolution` 424 passing, `cargo fmt --check` clean, `cargo clippy --all-targets` clean under CI's `-D warnings`. Four of those tests are new and were watched failing on the old order first.

## Summary by CodeRabbit

* **Improvements**
  * Subordinate people are now displayed in a consistent, case-insensitive label order.
  * Names are prioritized using display name, full name, username, and email, with unlabeled people listed last.
  * Ordering remains stable when multiple people share the same label.


hello1101n:
The Identity Resolution OpenAPI drift gate here is red for a reason outside this branch: the committed spec is stale on main as it stands. Reproduced on 340f2d454 with an otherwise clean tree, and #3134 regenerates the one line. This branch goes green once that lands.

hello1101n:
Correction to my note above: this gate is advisory, not one of the required checks, so nothing here was blocked. #3134 is green and clears the red mark once it lands.

cyberantonz (approved):

PR #3167, opened 2026-09-03
test(datapath): run the data-path suite on a compose test-stand instance

ktursunov:
**Why.** The metrics data-path rig ran on a compose stack of its own, minted its own tokens, stubbed identity, and judged its expectations through a CEL engine nothing else used; the identity lane spawned the service against a throwaway database. Neither proved the deployed path, and both cost a Rust build per CI run.

**What changed.** Both lanes are now `tests/datapath`, run as pytest against a `dev-compose.sh test-stand minimal --instance=<name>` stand: the suite seeds bronze, dbt builds in process, persons-seed mints every person from the HR rows a spec seeds, corrections go through the operator API, and every read is a real persona session through the gateway. The old rig is deleted.

**Assertions are pytest, not CEL.** A fixture keeps what to seed; the module beside it holds one test per case. The coverage gate reads the ledger those assertions record at run time, not the YAML text, and blocks the required check on any builtin metric no spec asserts. A view counts as covered only once a test has asserted something in it, and a meta test fails on a spec no module claims.

**Callers are real personas chosen for the rule under test.** The admin operator makes corrections and sees nobody; a spec's people are grafted beneath the seeded dev lead by their HR record, and that lead reads them; a service principal, obtained at the authenticator's token endpoint, reaches identity's internal routes. Nothing writes an identity row directly.

**CI runs one shard per metric class plus identity**, each on a minimal stand of its own, and this workflow alone emits the required `Run E2E suite` check: a deferral on a pull request, the verdict in the merge queue. The runner-image build job went with the rig, and every registry and index fetch on a shard's path is pinned or retried.

**The endpoint-coverage gate keeps only the suite it still has.** Deleting the rig took the analytics and identity ledgers with it, so `scripts/ci/api_coverage.py` is now the authenticator gate alone — no recording half, no suite switch, and none of the identity suppressions the document has since made stale. The stand suite's own gate covers analytics and identity.

**Also here.** `--instance` lets two stands coexist; the enrich sidecar is a compose overlay; one assistant-usage staging model gains the empty-table guard its five siblings had.

## Open after this

1. Two stands side by side have no automated coverage: every CI shard raises the default instance, so the per-instance env file, realm, manifest and port shift are only exercised by hand.

**Verified.** Full suite green on a local instance: 292 data-path and 21 identity tests. Seventeen dispatched CI runs of the lane across five commits, no failed job; a run takes about 12 minutes, the identity shard 3. On a pull-request event only this workflow emits the required check, as the deferral. Coverage gate: 105 of 105 builtin metrics asserted. The authenticator gate ran green in CI against its real ledger after the rewrite.

https://claude.ai/code/session_01Dh5gzjkqaVsQ226rFnZAHx


cyberantonz (approved):

src/ingestion/tests/e2e/conftest.py
```diff
@@ -115,11 +115,19 @@ def ch_migrations_applied(compose_stack: SessionConfig) -> SessionConfig:
...
+    The closure build that follows makes relation existence uniform for the whole
+    session — see `DbtRunner.build_closure`. It runs on the primary worker only:
+    the relations it creates are shared, and a second worker creating them again
+    would race the first.
+    """
     cfg = ch_migrations_applied
     runner = DbtRunner(cfg)
     runner.setup()
+    if _IS_PRIMARY:
+        runner.build_closure(worker_ctx=worker_ctx)
```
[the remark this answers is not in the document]
ktursunov: Not reachable: the rig never runs under xdist. No `-n`/`--numprocesses` is passed anywhere (`e2e.sh test`, `pytest.ini` addopts, the runner compose files), and `e2e-bronze-to-api.yml:345-350` documents why — CI shards are separate matrix jobs with isolated stacks, and *within* a shard tests run serially because the session rig is not xdist-safe; "non-primary workers don't wait for ClickHouse migrations" is already listed there. `build_closure` sits in the same primary-only pattern as `apply_ch_migrations` and `truncate_data_tables` directly above it; if xdist is ever enabled, all three need one barrier together, not one of them now for a mode the rig does not run.

deploy/compose/docker-compose.enrich.yml
```diff
@@ -0,0 +1,37 @@
+# The connector enrich sidecar, as an overlay on the root docker-compose.yml.
+#
+# A connector whose descriptor declares `images.enrich` materializes part of its
```
ktursunov: Simplify comments for maintainability

deploy/compose/two-instance-check.sh
ktursunov, line 1: Custom check - not needed in the code

scripts/ci/api_coverage.py
```diff
@@ -0,0 +1,587 @@
...
+# limiter) is universally unobservable.
+IDENTITY_UNIVERSAL_BOILERPLATE = frozenset({429})
+# The spec declares a generic `200` on the mutating routes, but the handlers
+# actually answer 201 (create) / 202 (accepted) / 204 (delete) — the same
+# `.standard_errors`-style spec-fidelity gap as analytics
+# #1669. The wrong 200 is BLOCKED (never answered), and the REAL success code
+# is REQUIRED_EXTRA (must be observed even though the spec doesn't declare
+# it) — so the gate cannot report 100% while no mutation success was ever
+# seen. Both are self-cleaning once the spec is fixed.
+IDENTITY_BLOCKED: dict[str, frozenset[int]] = {
```
ktursunov: Doublecheck the status codes from openapi spec - fix openapi spec if needed

scripts/ci/api_coverage.py
```diff
@@ -0,0 +1,587 @@
...
+    "DELETE /v1/person-roles/{id}": frozenset({204, 400, 401, 403, 404}),
+    "POST /v1/visibility": frozenset({201, 400, 401, 403}),
+    "GET /v1/visibility": frozenset({400, 401, 403}),
+    "DELETE /v1/visibility/{id}": frozenset({204, 400, 401, 403, 404}),
+    "GET /v1/subchart": frozenset({400, 401}),
+    "GET /v1/subchart/{personId}": frozenset({400, 401, 404}),
+}
+
+IDENTITY_REQUIRED_EXTRA: dict[str, frozenset[int]] = {
+    # POST /v1/persons-seed no longer exists (see the SKIP entry below) — its
```
ktursunov: Check if it's removed from openapi spec. Crucial to minimise overrides in gate checks

scripts/ci/api_coverage.py
```diff
@@ -0,0 +1,587 @@
...
+
+# The deprecated persons lookup (approved removal, zero callers) and the
+# persons-seed POST trigger (#1690: the seed is CLI-only — CronJob / manual
+# Job via the `seed` subcommand; the GET journal routes remain) no longer
+# exist. While the gate universe still lists them, SKIP entries let these
+# operations be legitimately unexercised.
+IDENTITY_SKIP_LIST: list[tuple[str, str]] = [
+    ("GET /v1/persons/{email}", "removed (approved: zero callers)"),
+    (
+        "POST /v1/persons-seed",
```
ktursunov: Check if openapi spec is actual for this endpoint

src/ingestion/tests/e2e/lib/api_coverage.py
ktursunov, line 1: Shouldn't this file be relocated to /tests location in appropriate subfolder?

src/ingestion/tests/e2e/lib/dbt_runner.py
ktursunov, line 1: Why this file is not relocated? Is it still required to exist?

tests/lib/insight_datapath/metric_coverage.py
ktursunov, line 1: This file by some reason was not migrated

src/ingestion/tests/e2e/lib/metric_expect.py
ktursunov, line 1: Should be migrated to global test folder

src/ingestion/tests/e2e/lib/spec_runner.py
ktursunov, line 1: why this file is not migrated?

deploy/compose/docker-compose.enrich.yml
ktursunov: Trimmed to two lines: the file is an overlay rather than a service because compose interpolates `${VAR:?}` for every command regardless of profile, so a required variable in the root file breaks a plain `up`. Everything else the descriptor and `enrich.py` already say.

deploy/compose/two-instance-check.sh
ktursunov: Deleted, along with the comment in `dev-compose.sh` that pointed at it.

Worth knowing: it was the only thing exercising two instances side by side. Every CI shard raises the default instance, so the per-instance env file, realm, manifest and port shift now have no automated coverage. `test-stand env` has no caller either. Say the word if you want one shard aimed at `--instance=<shard>` to keep that path executed once per run.

scripts/ci/api_coverage.py
ktursunov: Checked against the committed document, and the table was stale: the identity spec now declares the real `201`/`204` on the mutating routes, so every `IDENTITY_BLOCKED` entry claiming they answer `200` was dead, and `POST /v1/persons-seed` and `GET /v1/persons/{email}` have left the document entirely.

The deeper problem was that nothing produced those ledgers any more — the rig this PR deletes was the only writer. The file is now the authenticator gate alone: analytics and identity tables gone, recording half gone, `--suite` gone. CI ran it against the real ledger on this branch and it passed.

scripts/ci/api_coverage.py
ktursunov: Yes — removed. What is left is one `BLOCKED` entry (back-channel-logout's `200`, answered to the IdP's own POST and never to the test client) and one `SKIP_LIST` entry. The universal-boilerplate and required-extra mechanisms were empty for the surviving suite, so they went with the suites that used them.

Analytics and identity coverage is not lost: `tests/lib/insight_stand/coverage.py` gates it against the stand suite's own ledger.

scripts/ci/api_coverage.py
ktursunov: It was not. The entries for that endpoint were written when the document declared only `200`; it declares the real codes now. The whole identity block is gone rather than corrected, because nothing writes an identity ledger for it to read since the rig's deletion.

src/ingestion/tests/e2e/lib/api_coverage.py
ktursunov: The shim is deleted with the rest of the rig. The gate itself sits at `scripts/ci/api_coverage.py` — it is a CI script the authenticator workflow runs over a ledger the Rust harness writes, not a test, so it belongs beside the other `scripts/ci/` gates rather than under `tests/`.

src/ingestion/tests/e2e/lib/dbt_runner.py
ktursunov: It is relocated — `tests/lib/insight_datapath/dbt_runner.py`. This comment was on the old path, which no longer exists on the branch.

Its tests were dropped in the move, which an independent review caught: `tests/datapath/meta/test_dbt_project.py` now brings back the relation-naming contracts and the two connector-pass regression guards.

tests/lib/insight_datapath/metric_coverage.py
ktursunov: That path is the migrated location: `tests/lib/insight_datapath/metric_coverage.py` is where the metric×view gate lives now, and the workflow runs it from there. Nothing of it remains under `src/ingestion/`.

src/ingestion/tests/e2e/lib/metric_expect.py
ktursunov: Migrated — `tests/lib/insight_datapath/metric_expect.py`. Its unit tests were not, which a review caught: `tests/datapath/meta/test_metric_expect.py` now covers the selection, completeness and refusal rules, plus the new rule that a view is booked as covered only by an assertion rather than by being read.

src/ingestion/tests/e2e/lib/spec_runner.py
ktursunov: Migrated — `tests/lib/insight_datapath/spec_runner.py`. The comment sits on the old path, deleted with the rest of the rig.

scripts/ci/api_coverage.py
```diff
@@ -0,0 +1,587 @@
+#!/usr/bin/env python3
+"""API endpoint coverage report — which analytics routes the e2e suite exercises.
+
+Recording half (imported by the rig): `record_response` is an httpx response
+event-hook on the single client every suite request flows through; it records
+`(method, path) -> {status codes}` into a module-level ledger that
+`conftest.pytest_sessionfinish` dumps to `.artifacts/observed_endpoints.json`.
+
+Gate half (`python3 lib/api_coverage.py`, stdlib only; blocking in `./e2e.sh
```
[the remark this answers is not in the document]
ktursunov: Fixed — the gate's docstring names its own path now. It also describes what the file actually is after this branch: the authenticator endpoint-coverage gate, with the analytics and identity halves removed because the rig that produced their ledgers is deleted.

src/ingestion/tests/e2e/conftest.py
```diff
@@ -213,10 +223,16 @@ def pytest_collection_modifyitems(config, items):
     items.sort(key=lambda i: 0 if "meta/" in str(i.path) else 1)
 
 
+def _has_cases(path: Path) -> bool:
+    """A spec whose cases were ported to a Python module keeps only its bronze."""
+    document = yaml.safe_load(path.read_text(encoding="utf-8")) or {}
+    return bool(document.get("cases"))
```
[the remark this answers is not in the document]
ktursunov: Moot: `pytest_generate_tests` and the YAML collection path are gone with the rig. Specs are reached through each module's `SPEC` constant, and a malformed one fails that module rather than collection.

src/ingestion/tests/e2e/lib/api_coverage.py
```diff
@@ -1,593 +1,24 @@
...
+    raise ImportError(f"cannot load api_coverage from {_TARGET}")
+_module = importlib.util.module_from_spec(_spec)
+# Registered before execution: dataclasses resolves string annotations through
+# sys.modules[cls.__module__], and the moved file has them all as strings.
+sys.modules[_spec.name] = _module
+_spec.loader.exec_module(_module)
+
+record_identity_response = _module.record_identity_response
+record_response = _module.record_response
+main = _module.main
```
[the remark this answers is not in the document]
ktursunov: The shim is deleted. The gate lives at `scripts/ci/api_coverage.py`, and its recording half went with the rig that was its only importer.

src/ingestion/tests/e2e/lib/fixture_loader.py
```diff
@@ -184,9 +177,9 @@ def load(path: Path, *, schemas_dir: Path | None = None) -> TestYaml:
         bronze[table] = resolved
         schemas[table] = schema
 
-    cases = doc["cases"]
-    if not isinstance(cases, list) or not cases:
-        raise FixtureError(f"{path}: `cases` must be a non-empty list")
+    cases = doc.get("cases") or []
+    if not isinstance(cases, list):
+        raise FixtureError(f"{path}: `cases` must be a list")
```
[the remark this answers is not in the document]
ktursunov: Good catch, and it was live: three git specs carried cases no module claimed, so thirteen assertions never ran. They are ported, and `tests/datapath/meta/test_spec_ownership.py` now fails on an unclaimed spec, a spec claimed twice, or a `SPEC` naming a file that is not there. Verified it fails by planting an unclaimed fixture.

src/ingestion/tests/e2e/README.md
```diff
@@ -85,7 +85,7 @@ e2e/
 │   ├── analytics.py        # builds + spawns the analytics binary
 │   ├── worker.py               # WorkerContext (resolves pytest-xdist worker id)
 │   ├── metric_coverage.py      # builtin metric coverage gate
-│   ├── api_coverage.py         # endpoint-coverage report + httpx recording hook
+│   ├── api_coverage.py         # shim: loads scripts/ci/api_coverage.py for the identity lane
```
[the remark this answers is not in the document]
ktursunov: Moot: that README is deleted with the rig. The suite's own documentation is `tests/stand/README.md`, which this branch corrects where it described the retired gate.

PR #3171, opened 2026-09-03
fix(ingestion): put committer_date in the gold DDL snapshot

Gregory91G:
**Why.** The connectors-ddl snapshot check has failed on every push to `main` since #3155, and on every pull request with it — a branch is checked against its merge with `main`, so the failure reads as the branch's own and each author spends the look finding out it is not.

**What changed.** #3155 added `committer_date` to `git_authored_commits` and `git_commit_file_changes` and regenerated the silver snapshot, but not the gold one. These are the two lines the check's own re-dump produces, in the position dbt-clickhouse writes them: between `observed_at` and `entity_id` in one model, between `observed_at` and `data_source` in the other — the snapshot is positional, so where they sit is part of being correct.

**Taken from the check's re-dump, not hand-authored.** `insight.sql` is generated and must not be edited manually. This change applies the `connectors-ddl-drift` output produced by the failing CI run; the DDL check regenerates it and verifies the snapshot.

**Verified.** `git diff` matches the failing run's diff byte for byte, including both blob hashes; the column position was cross-checked against each model's own `SELECT` order rather than only against the diff. The check on this PR is the real proof — it re-dumps from a fresh ClickHouse.

## Summary by CodeRabbit

* **New Features**
  * Added optional committer date information to authored commit and commit file change records.
  * This enables more precise tracking of when commits were made.


Gregory91G:
Folded into #3143 as a commit — same two lines, same resulting blob (`53c5455`). Merging one PR instead of two.

PR #3199, opened 2026-09-05
fix(frontend): order the roster by name

hello1101n:
**Why.** The People roster's Members list opens in an order nothing on screen explains — behind-peers first — while the org tree next to it is A→Z. Rows read as unordered.

**What changed.** The grid defaults to name order whenever the standing facet is off, which is exactly the roster: no chip names a standing, and the Person header offers only a name toggle. Views that do show the facet — AI & Cost, the group drilldowns — keep behind-first.

**Screenshots.** People roster on the MSW mock roster, same data either side.

| Before | After |
|---|---|
|  |  |

**Verified.** `pnpm test` (2357), `pnpm lint`, `pnpm typecheck` — green. Screenshots from `pnpm dev` on the mock roster, no deployed data.

### DIFF
--- a/src/frontend/src/components/portal/team-state-view.test.tsx
+++ b/src/frontend/src/components/portal/team-state-view.test.tsx
@@ -166,6 +166,12 @@ describe("TeamStateView", () => {
     expect(screen.getAllByText("boss")).toHaveLength(2);
   });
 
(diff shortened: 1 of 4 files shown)


hello1101n:
Follow-up on the dead facet this order came from: #3201.

cyberantonz (approved):

PR #3248, opened 2026-09-06
fix(git-cli-proxy): evict an entry whose purge keeps failing, and purge under pressure before the cap

aleksdotbar:
**Why.** An entry over its per-repository cap whose in-place repack cannot finish inside the heavy budget is retried on every drift cycle: the purge holds the entry's write lock for the whole budget, times out, and the readers queued behind it get an over-cap 429 the moment it releases — with their own retry budget long spent. The pack only grows between attempts, so the repository becomes permanently unservable while its blob weight sits on the volume.

**What changed.** The post-serve purge now counts consecutive failures per entry; the second one evicts the entry under the write lock the purge already holds (new `EvictionTier::PurgeExhausted`), and the next open re-clones the blobless skeleton. Separately, an entry at or past half its cap queues its purge on the first losing probe instead of yielding through three, so repacks run while the pack can still finish inside the budget.

**Eviction over a live walk is a deliberate trade.** It bumps the generation, so a continuation in flight on that repository answers 409 and the connector reruns it from its stored cursor — the documented contract. For an entry that can no longer be purged or served, one rerun beats an unbounded loop of 429s.

**Verified.** `cargo test -p git-cli-proxy` (220 lib + 20 boot); both new regression tests fail when their rule is disabled and pass with it; `cargo fmt --check`; clippy pedantic clean.

### DIFF
--- a/src/backend/services/git-cli-proxy/src/engine/metrics.rs
+++ b/src/backend/services/git-cli-proxy/src/engine/metrics.rs
@@ -22,13 +22,17 @@ pub enum EvictionTier {
     Blob,
     /// The whole entry deleted.
     Full,
+    /// The whole entry deleted because repeated purges could not shed its
+    /// blobs in place.
+    PurgeExhausted,
 }
 
 impl EvictionTier {
-    const fn as_str(self) -> &'static str {
+    pub(crate) const fn as_str(self) -> &'static str {
         match self {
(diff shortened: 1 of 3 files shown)


aleksdotbar:
Superseded by #3258, which carries this commit as its first one on main; the same fix reached release-2026.08 through #3249.

Issue #3050, opened 2026-09-02
Validate metric: git.default_branch_prs_created


hello1101n:
Part of #3039.

- [x] What does `git.default_branch_prs_created` do? Does the definition make sense?
- [x] Where is it used? Mark used / not used.
- [x] Write an e2e test: specific bronze → transform → request in the API, for each scenario.
- [x] Check the current GitHub and Bitbucket validation scope. Does it make sense with different data?
  - [x] Counted correctly for GitHub and Bitbucket?
- [x] Re-check related fixes and recent commits.

mozhaev-dev:
Need VZ validation

mozhaev-dev:
Counts authored pull requests whose destination is the repository's current default branch, dated by creation. An unreported destination and a repository whose default branch is unknown both read as not default, so default plus non-default stays exhaustive. A request the source names no author for, by address or by account, reaches nobody. The definition is sound.

mozhaev-dev:
Not shown directly: the UI uses `git.prs_created`. This scoped half is served through the catalog and the API with period, peer, timeseries and breakdown views.

mozhaev-dev:
Covered end to end from synthetic bronze fixtures through transformations to `/v1/metric-results`: a destination compared against its own repository's default, two repositories that disagree about the default name, a repository that reported no branches reaching one scope only, the destination breakdown, attribution by source account, the default plus non-default partition, creation dating, and an empty period.

mozhaev-dev:
Validated on a GitHub install and a Bitbucket install against the sources themselves: the request set recomputed from the silver layer per scope equals the served figures once requests with neither an author address nor an account are set aside, evidence equals observations per author and day, and every repository's default name equals the head of the connector's own clone. The API answers match the warehouse on GitHub, attributed through the source account. GitLab is deferred to a separate validation.

mozhaev-dev:
Yes for both. Default plus non-default equals all created requests for every author on both installs, and one evidence row stands for each request, dated by its creation. The requests that reach nobody are those the source names no author for: automation accounts on GitHub, a service account on Bitbucket.

mozhaev-dev:
Re-checked after #3353 and #3379 landed: this metric reads the request's destination, not the commit collapse, so neither changed it; the content-order and shared-hash rules do not apply to requests. Figures unchanged on both installs.

mozhaev-dev:
Nuances, not fixed or by design:
[...]

Issue #3039, opened 2026-09-02
metrics validation


hello1101n:
## **How to verify metrics?**

1. What does this metrics do? Does it make sense? 
2. Check where is it used? Mark if used or not.
3. Write e2e test that from specific bronze →  transform → request in API. (For each metric for different scenarios)
4. Check what is coming in v1 stand, f1 stand, and dev stand for this metric. Does it make sense? Verify with different data. (**git metrics:** Is it correctly counted for gh? gitlab? bitbucket?)

## **Metrics to validate**

 1. AI - skip
 2. Collab - skip
 3. git.active_days - @mozhaev-dev — #3041
 4. git.approval_to_merge_time_h -| @ktursunov  mark as planned on all stands please — #3090
 5. git.code_lines - @mozhaev-dev — #3043
 6. git.commit_size - @aleksdotbar — #3044
 7. git.commits_per_active_day - @aleksdotbar — #3045
 8. git.default_branch_code_lines - @mozhaev-dev  — #3046
 9. git.default_branch_commits - @mozhaev-dev — #3047
10. git.default_branch_lines_added - @mozhaev-dev  — #3048
11. git.default_branch_lines_removed - @mozhaev-dev — #3049
12. git.default_branch_prs_created - @mozhaev-dev — #3050
13. git.default_branch_prs_merged - @mozhaev-dev — #3051
14. git.first_review_time_h | @ktursunov  mark as planned on all stands please — #3091
15. git.lines_added — #3053 @aleksdotbar 
16. git.lines_removed — #3054 @aleksdotbar 
17. git.merge_rate — #3055 @mozhaev-dev
18. git.merges_without_approval_rate | @ktursunov  mark as planned on all stands please — #3092
19. git.multi_reviewer_rate | @ktursunov  mark as planned on all stands please — #3093
20. git.non_default_branch_code_lines — #3058 @mozhaev-dev 
21. git.non_default_branch_commits — #3059 @mozhaev-dev 
22. git.non_default_branch_lines_added — #3060 @mozhaev-dev 
23. git.non_default_branch_lines_removed — #3061 @mozhaev-dev 
24. git.non_default_branch_prs_created — #3062 @mozhaev-dev 
25. git.non_default_branch_prs_merged — #3063 @mozhaev-dev 
26. git.pr_abandonment_rate | @ktursunov  mark as planned on all stands please — #3094
27. git.pr_cycle_time_h — #3065
28. git.pr_size — #3066
[...]

PR #3286, opened 2026-09-07
test(e2e): pin that the default branch is resolved per repository, and from the flag

mozhaev-dev:
Part of #3039, and the coverage half of #3050. Follows #3185, #3186, #3253, #3254 and #3256.

**Why.** Only two specs assert this comparison — `git_branch_scope` across four views, `git_prs_merged` as the branch-scope split on merges — and both use one repository whose default is `main` with every request carrying a destination. Three wrong readings pass both: a comparison hardcoded to `main`, one letting an empty destination match an unknown default, and one that stops selecting branches on the default flag at all.

**The third is the one worth reading.** Removing `WHERE is_default = 1` from `repository_default_branches` — the line that makes that CTE mean *default* branch — leaves every test in the branch-scope family green today, because with the filter gone the CTE degrades to a `min()` over all branches and every existing fixture's alphabetically-first branch happens to be its default. With this fixture in place that mutation fails, and only this fixture fails on it.

**What changed.** Three repositories that disagree about their default — `main`, `master`, and none reported at all — with six requests, two of which also merge so the merged pair rides the same fixture rather than a near-duplicate one. Each wrong reading is caught by a different rule: the destination split catches the hardcoded name (`main` reads twice, `master` matches no row); the period total catches the empty-destination guard; and a non-default branch sorting *before* `main` catches the missing flag filter.

**The repository case needed both halves.** Its negative assertion — that the branchless repository contributes nothing to this scope — passes for the wrong reason on its own: strip that repository's author rows and it never reaches the breakdown at all. A positive rule on the opposite scope now holds those two requests alive.

**Recorded, not asserted.** The comparison uses the repository's *current* declared default, so renaming one reclassifies requests opened before the rename — which the metric's own explanation ("does not move afterwards") denies. The rig seeds one branch state per run, so no fixture can express a rename; the description says so where the next reader will find it.

**Verified.** `./e2e.sh test -k git_` — 28 passed. Three model mutations, each restored byte-identically: the hardcoded branch name, the dropped empty-destination guard, and the dropped default-flag filter.

**Out of scope.** `repository_default_branches` resolves a repository claiming two default branches by `min(branch_name)`, i.e. alphabetically rather than by recency. No fixture in the suite has two such rows, and this one does not add the case — worth its own issue.

### DIFF
--- a/src/ingestion/tests/e2e/metrics/git_default_branch_prs_created.test.yaml
+++ b/src/ingestion/tests/e2e/metrics/git_default_branch_prs_created.test.yaml
@@ -1,0 +1,231 @@
+spec_version: 1
+description: >
+  Metric: git.default_branch_prs_created — requests opened against the
+  repository's own default branch, #3050.
+
+  How it's computed (bronze → silver → gold):
+    • bronze: each request's destination branch, and the branch list the
+      connector reports per repository
+    • silver: one class row per request and per branch
+    • gold:   the destination is compared against THAT repository's declared
+      default branch, and the scope picks one of the pair — so default plus
+      other always equals the total, by construction rather than by
+      reconciliation
+
+  Only two specs assert this comparison: `git_branch_scope` across all four
+  views of both created metrics, and `git_prs_merged` as the branch-scope split
+  on merges. Both use ONE repository whose default is `main` and give every
+  request a destination — so a comparison hardcoded to `main`, one that let an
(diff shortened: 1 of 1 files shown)


mozhaev-dev:
Closing: the metric suite moved to `tests/datapath/metrics/<class>/` as a fixture + pytest module pair, so this branch targets a path that no longer exists — it would merge cleanly and land an orphan file. Reopened against the new structure.

PR #3293, opened 2026-09-08
fix(gold): give CI evidence the source the drilldown resolves its links against

mozhaev-dev:
Found while validating the git metrics under #3039 — the CI family shares the drilldown surface, and its repository column never linked anywhere.

## The defect

A drilldown row's repository becomes a link only when the row offers a `source` dimension naming the provider **and** a `source_id` among its details: the resolver reads the provider from the dimensions, the instance id from the details, and returns nothing at all when either is missing (`metric_drilldown/presentation.rs:296-328`).

CI evidence emitted neither, so the repository column was inert for every CI metric while the same column links out on the git side.

## The fix

Both emitting paths now carry them, derived exactly as the git evidence derives them:

- value — the connector's `data_source` with its `insight_` prefix stripped;
- label — the shared source-label macro;
- details — the row's own `source_id`.

Collected commits are left alone: that path emits no dimensions at all, so it shows no repository column to link from.

The relation is materialized as a table and the column set does not change, so this needs no migration — a gold rebuild is enough.

## Verified

On a compose test-stand instance:

- `--tree=tests/datapath/metrics/ci` → 14 passed, so no CI metric moved;
- both record kinds carry the pair in the built relation — run rows and deployment rows each report a `source` dimension (`github` / `GitHub`) and a details `source_id`, where before both were absent.

## Not asserted

The rendered link also depends on an external-source registry entry for the instance, and nothing in the data-path suite asserts a drilldown link today — the suite only calls `/v1/metric-results`. So this fixes the data half; the href itself stays unasserted. Giving the suite its first drilldown assertion is worth doing and is bigger than this change.

## Left out deliberately

The same finding has a second half: the repository value is a bare `owner/name`, so two CI sources holding a repository of the same name collide into one row. Fixing that means prefixing the value the way git does, which changes what groups together and what the interface shows, and needs the query reference and the front end checked with it. That belongs in its own change, not here.

## Summary by CodeRabbit

* **Data Reporting**
  * CI metric evidence now includes the originating source as a dimension for both CI runs and deployments.
  * Evidence details now include a source identifier for improved traceability.

### DIFF
--- a/src/ingestion/gold/ci_metric_evidence.sql
+++ b/src/ingestion/gold/ci_metric_evidence.sql
@@ -34,6 +34,7 @@ runs AS (
         branch,
         started_at,
         duration_s,
+        data_source,
         toDate(started_at) AS metric_date,
         -- Whether the run's head commit was ever collected by the commits
         -- stream. PR runs build synthetic merge refs and fork commits the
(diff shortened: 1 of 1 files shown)


mozhaev-dev:
Closing as out of scope: this repository's CI metric family is not part of the git metric coverage epic this branch came out of, and the change is a product fix rather than test coverage. Whoever picks up the CI metric will carry it.

The diagnosis stays available in the description above, and the branch `fix/ci-evidence-repository-link` is kept, so reopening this is a one-command job.

PR #3297, opened 2026-09-08
feat(datasets): give datasets a home, a registry and a discovery API

aleksdotbar:
**Why.** A data set is the unit everything else stands on: one kind of thing people can ask about, at one stated grain, with the counting rules already applied. Until now there was no such unit — only pre-built metrics over tables nobody outside the pipeline could interpret, so every new question meant a developer reading warehouse schemas.

**What changed.** Data sets get a home, a registry and a way to be discovered.

**A data set is one folder.** `src/ingestion/datasets/<key>/` holds the model that materializes it, its dbt properties and grain test, its declaration, and a README saying what one row is. Deliberately outside the medallion layout: a data set is a product-level thing, and the layers underneath it may change. dbt gains the folder as a model path; `.dbtignore` keeps `dataset.yaml` out of dbt's hands.

**The declaration says what may be asked.** Grain, the columns that may be grouped or filtered with the label that travels beside each, the columns that may be folded, the columns carrying event time, and how the relation must be read to be duplicate-free. One validator checks every declaration against the warehouse's column catalogue and refuses it by field name — a mismatch fails a test run rather than a request.

**Two data sets to start.** `git_commits` (one row per commit somebody wrote) and `git_file_changes` (one row per path a commit touched), both over a shared authored-change relation that applies the content dedup once, so line totals agree with the metric evidence.

**Discovery.** `GET /v1/datasets` and `GET /v1/datasets/{key}` describe what can be asked about, without revealing database, relation, read discipline, tenancy column or row identity. Admin-only until access rules land.

**Shipped declarations travel in the image.** The datasets folder reaches the analytics build as a named additional build context, so the build script can embed every `dataset.yaml` without widening the build context. Data sets people add at runtime will join the same registry from the service's own store — one loader, one validator, one list.

**Out of scope.** Asking questions: [#3187](https://github.com/constructorfabric/insight/pull/3187) stacks on this and adds `POST /v1/query`.

**Docs.** `docs/domain/datasets/` — PRD (what a data set is to a person, and the roadmap for ones people add) and DESIGN (anatomy, rules, registry, lifecycle, how to add one).

**Verified.** 872 unit tests, clippy and rustfmt clean, no build warnings; `dbt parse` and `dbt ls` see both data sets from the new path; `cfs validate` and `cfs validate-toc` pass for both documents; OpenAPI and stand schemas regenerated; ruff clean.

### DIFF
--- a/.cf-studio/config/artifacts.toml
+++ b/.cf-studio/config/artifacts.toml
@@ -107,6 +107,18 @@ path = "docs/components/backend/analytics/specs/connector-health/DESIGN.md"
 name = "Connector Health Design"
 traceability = "DOCS-ONLY"
 
+[[systems.artifacts]]
+kind = "PRD"
+path = "docs/domain/query-engine/PRD.md"
+name = "Query Engine PRD"
+traceability = "DOCS-ONLY"
+
+[[systems.artifacts]]
+kind = "DESIGN"
+path = "docs/domain/query-engine/DESIGN.md"
+name = "Query Engine Design"
+traceability = "DOCS-ONLY"
+
 [[systems]]
 name = "Identity Service"
 slug = "identity-svc"
--- a/docs/domain/query-engine/DESIGN.md
+++ b/docs/domain/query-engine/DESIGN.md
@@ -1,0 +1,932 @@
+---
+version: 2.1
+status: proposed
+date: 2026-09-08
+---
+
+# Technical Design — Query Engine
+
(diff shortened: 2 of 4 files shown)


aleksdotbar:
Superseded by #3321: the branch was renamed to `feat/datasets` as this became the datasets PR, and GitHub closed this one rather than following the rename. Same work, plus the dataset registry and discovery API.

PR #3341, opened 2026-09-09
ci(e2e): skip both suites when only charts or unit tests change

ktursunov:
**Why.** A diff confined to a service's Helm chart lands under `src/backend/`, so both e2e suites treated it as a backend change: five data-path stands plus the full compose stand, for every queue entry.

**What changed.** Each `changes` job now drops the paths nothing it raises or runs reads, then decides relevance on what is left.

**`*/helm/Chart.yaml` is exempt from the drop.** `dev-compose.sh` takes the pulled image's `appVersion` from it, so a pin bump stays a real change. Deleting one pattern undoes the exemption.

**`gateway`, `gateway_checks` and `authenticator` keep reading the raw diff.** They cross into `gateway.yml`, which helm-lints the gateway chart and runs its own e2e suite from `services/*/tests` — trees the compose stand never reads. A chart-only diff therefore skips both suites while `Gateway checks` stays armed.

**Markdown is ignored only where nothing parses it.** dbt's doc paths default to its resource paths and the data-path suite runs `dbt parse`, so markdown under `src/ingestion/**` is a parser input; the rule is anchored to `src/backend` and `tests` instead.

**Out of scope.** Per-shard relevance for the metric matrix — it needs the coverage gate scoped to the shards that ran, or a partial run reds a required check. A change to the gateway or authenticator e2e suite also still runs the data-path suite, since `^src/backend/` is what makes that workflow relevant.

**Verified.** Both `filter` scripts extracted from the YAML and replayed over 15 file lists with a stubbed `git` — 30 executions, 0 failures. A real chart-only file list (#3322) yields `relevant=false` on both workflows with `gateway_checks=true`; `dbt` data tests, markdown under the dbt trees, chart pin bumps and every service source change are unchanged. `actionlint` reports only a pre-existing `SC2001`, confirmed present at `HEAD`.

### DIFF
--- a/.github/workflows/e2e-stand.yml
+++ b/.github/workflows/e2e-stand.yml
@@ -150,6 +150,21 @@ jobs:
           fi
           changed="$(git diff --name-only "$base_point" "$HEAD")"
           printf '  %s\n' "${changed//$'\n'/$'\n  '}"
+
+          # Ignored because nothing this stand raises or runs reads these paths.
+          # INVARIANT: */helm/Chart.yaml is NOT ignored — dev-compose.sh takes
+          # the image the stand pulls from its appVersion.
+          IGNORED_PATHS='^src/(backend/services/[^/]+|frontend)/helm/|^src/backend/(services|tools)/[^/]+/tests/|^src/ingestion/tools/seed/tests/|^tests/datapath/|\.md$'
+          IMAGE_PIN_PATHS='^src/(backend/services/[^/]+|frontend)/helm/Chart\.yaml$'
+          # `|| true`: a grep selecting nothing exits 1 — set -e fails the step.
+          changes_to_test="$(printf '%s\n' "$changed" | grep -vE "$IGNORED_PATHS" || true)"
+          image_pins="$(printf '%s\n' "$changed" | grep -E "$IMAGE_PIN_PATHS" || true)"
+          changes_to_test="$(printf '%s\n%s\n' "$changes_to_test" "$image_pins" | grep -v '^$' || true)"
+          if [ -n "$changes_to_test" ]; then
+            printf '%s\n' "$changes_to_test" | sed 's/^/  affects this run: /'
+          else
+            echo "  affects this run: nothing — no changed file can reach this stand"
+          fi
           relevant="$forced"
           backend=false
           frontend=false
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-stand.yml
```diff
@@ -150,6 +150,21 @@ jobs:
           fi
           changed="$(git diff --name-only "$base_point" "$HEAD")"
           printf '  %s\n' "${changed//$'\n'/$'\n  '}"
+
+          # Ignored because nothing this stand raises or runs reads these paths.
+          # INVARIANT: */helm/Chart.yaml is NOT ignored — dev-compose.sh takes
+          # the image the stand pulls from its appVersion.
+          IGNORED_PATHS='^src/(backend/services/[^/]+|frontend)/helm/|^src/backend/(services|tools)/[^/]+/tests/|^src/ingestion/tools/seed/tests/|^tests/datapath/|\.md$'
```
aleksdotbar: This pattern also matches `src/backend/services/gateway/tests/`, which is the gateway e2e suite (`run-e2e.sh`, `conftest.py`, `docker-compose.e2e.yml`) that `gateway.yml` runs, and the queue only reaches `gateway.yml` through `gateway_checks=true` here. Suggest excluding the gateway tests directory, or listing the services whose `tests/` are true unit trees instead of `[^/]+`.

aleksdotbar (changes requested):
**Medium — the `e2e-stand.yml` ignore list drops the gateway e2e suite's own files.** `^src/backend/(services|tools)/[^/]+/tests/` covers `src/backend/services/gateway/tests/`, which is not a unit-test tree: it holds `run-e2e.sh`, `conftest.py` and `docker-compose.e2e.yml`, the suite `gateway.yml` executes. `gateway.yml` has no `pull_request` trigger; the queue reaches it only through `gateway_checks=true` from this filter. After this change, editing the gateway e2e scripts yields `gateway_checks=false`, so they first run on push to `main`. The authenticator's `tests/` is fine: `authenticator.yml` runs it on the pull request from its own paths. The same pattern in `e2e-bronze-to-api.yml` is harmless, that suite never runs gateway tests.

**Low — `tools/[^/]+/tests/` drops routegen tests from `gateway_checks`.** Previously any change under `src/backend/tools/routegen/` set it. ci.yml runs the routegen tests on the pull request, so only the queue-time rerun is lost. Worth a sentence in the body.

**Low, pre-existing — a `Chart.yaml` pin bump still forces a source build.** The exempted pin survives into `changes_to_test`, matches `^src/backend/` and the per-service patterns, so `backend=true` and, for example, `analytics=true`: the stand compiles from the tree instead of pulling the image the bump pins. Not introduced here; a follow-up issue rather than a change in this PR.

.github/workflows/e2e-stand.yml
ktursunov: Confirmed, and there is a second instance of the same class in the same list — thank you, this would have shipped as a silent gap.

`services/gateway/tests/` is the gateway e2e suite (`gateway.yml:183` runs `run-e2e.sh`), `tools/routegen/tests/` is exercised by `cargo test -p routegen` (`gateway.yml:75`), and `services/authenticator/tests/` is run by `authenticator.yml:94`. I dropped the `(services|tools)/*/tests/` rule from both workflows rather than narrowing it: `git-cli-proxy/tests/boot.rs` was the only tree it covered that no lane runs, and no e2e flag reads that path, so the rule bought nothing once the three real suites were carved out.

The second instance is `services/gateway/helm/` itself. It is the route table, not a deployment detail (DD-GW-02) — `gateway.yml:115-122` helm-lints the chart and pipes `helm template --show-only templates/configmap.yaml` through routegen. So ignoring it for stand relevance is right (the compose stand reads `deploy/compose/gateway/routes.yaml`), but it must not reach the flags that cross into `gateway.yml`. `gateway`, `gateway_checks` and `authenticator` now read the raw diff; only the stand's own decisions read the filtered list. A false "unchanged" on those three would also let `gateway.yml` accept a stale image, since they feed `prebuilt_images.py`'s `required` policy.

Net effect on the case that motivated the PR — #3322, a chart-only diff: both suites still skip, and `Gateway checks` still runs.

Replayed both `filter` scripts, extracted from the YAML against a stubbed `git`, over 15 file lists (30 executions, 0 failures):

| Changed files | stand relevant | gateway_checks |
| --- | --- | --- |
| #3322's real file list | false | true |
| `gateway/helm/values.yaml` | false | true |
| `gateway/tests/run-e2e.sh` | true | true |
| `routegen/tests/golden.rs` | true | true |
| `authenticator/tests/e2e_refresh.rs` | true | true |
| `identity-resolution/helm/values.yaml` | false | false |

One consequence I did not try to fix here: a change to the gateway or authenticator e2e suite still runs the data-path suite and a backend compile, because `^src/backend/` is what makes that workflow relevant. That is pre-patch behaviour, and narrowing it needs its own argument.

aleksdotbar (approved):

Issue #3059, opened 2026-09-02
Validate metric: git.non_default_branch_commits


hello1101n:
Part of #3039.

- [x] What does `git.non_default_branch_commits` do? Does the definition make sense?
- [x] Where is it used? Mark used / not used.
- [x] Write an e2e test: specific bronze → transform → request in the API, for each scenario.
- [x] Check what the v1, f1 and dev stands return for it. Does it make sense? Verify with different data.
  - [x] Counted correctly for github, gitlab and bitbucket?

mozhaev-dev:
Counts authored, non-merge commits that have not reached a default branch: work in flight, or abandoned. A commit leaves this side when its copy reaches its repository's default branch, a merged pull request into a default branch lists it, or default-branch content matches it in time order; since #3379 that is decided per commit, whichever connected repository holds the copy. Derived commits are excluded: a squash or rebase result whose originals were collected, and a later carrier of the same patch. Dated by the author date. The definition is sound.

mozhaev-dev:
Not shown as its own figure. It is the evidence carrier for the drilldown of the three non-default line metrics, and it is served with period, peer, timeseries and breakdown views, including the hour block.

mozhaev-dev:
Covered end to end from synthetic bronze fixtures through transformations to `/v1/metric-results`: the default plus non-default partition with this side's own values, a merged request promoting a commit out of this side, the content-order rule keeping a commit here when the trunk already held its content, a commit shared by two repository copies, the hour-block breakdown over both scopes, the composition with derived commits, and an empty period.

mozhaev-dev:
Validated on a GitHub install and a Bitbucket install against the sources themselves: per person-day equality between the metric rows and the served observations, one evidence row per commit dated on its author day, an independent recompute from the silver layer, and default-branch reachability from the connector's clone cache on both, so that nothing served here sits on a default branch. The API answers and the drilldown rows match the warehouse on GitHub. GitLab is deferred to a separate validation.

mozhaev-dev:
Yes for both. Default plus non-default equals all authored commits for every person on both installs, the hour blocks partition this side exactly, and no commit served here sits on its repository's default branch, apart from one that landed after its last collection and waits for the next read.

mozhaev-dev:
[...]

PR #3353, opened 2026-09-10
fix(gold): a commit has not landed on content the default branch already held

### LINKED ISSUE #3340: Default-branch content heal promotes a commit on content the trunk already carried
Found while validating `git.non_default_branch_commits` (#3059).

## What the rule does today

`git_default_branch_commits` decides whether a commit's work reached the default branch. Its second half — the content rule — exists because reachability cannot answer that question after a squash: the original commits never become ancestors of the default branch, however plainly their work landed. So the rule reads the content instead, and declares a commit landed when one of its file changes has the same **(path, content identity)** as a change on a default-branch commit.

The content identity is the file's object id after the change (or, for a deletion, a marker plus the id before it). So the test is "this path was, at some point in the default branch's history, in byte-identical shape".

## The defect

The rule never asks **when** that content reached the default branch. A commit that brings a file to a state the default branch **already carried** therefore reads as landed, even though nothing of it landed.

Three ordinary ways to trigger it, all of them normal developer behaviour:

- restoring a file to the state the trunk already has (`git checkout main -- LICENSE`);
- copying a file across from the trunk onto a working branch;
- reverting your own edit, so the file returns to the state the trunk holds.

Two things make the blast radius larger than the trigger:

1. **One file is enough.** The rule emits the commit hash as soon as any single change matches, so an incidental `.gitignore`, licence file or config can carry a commit of any size across.
2. **The whole commit moves.** Branch scope lives on the commit and its lines inherit it — deliberately, so a commit's lines and the commit itself never disagree. So every file and every line of the commit follows the one match.

## What it affects

Every metric built on the split, in both directions at once:

- overstated: `git.default_branch_commits`, `git.default_branch_code_lines`, `git.default_branch_lines_added`, `git.default_branch_lines_removed`;
- understated: `git.non_default_branch_commits`, `git.non_default_branch_code_lines`, `git.non_default_branch_lines_added`, `git.non_default_branch_lines_removed`;
- and the `branch_scope` breakdown of the unscoped line and commit metrics, plus the drilldown rows behind all of the above.

The unscoped totals (`git.commits`, `git.code_lines`, …) are untouched — only the split moves.

This is **not** #2846, which is the opposite direction: there a genuinely landed commit keeps reading non-default because `is_default_branch` went stale outside the sync lookback. The two errors can coexist in one install.

## Fix

Require the content to have reached the default branch **no earlier than** the commit claiming to have landed it. A squash or a fast-forward satisfies this — the trunk carrier is created at or after the branch commit. Content that was already there does not.

Concretely, in `git_default_branch_commits.sql`:

1. the landed-content set carries the arrival time of each (path, content identity) — the date of the default-branch commit that brought it, committer date with the author date as fallback;
2. the candidate set carries the same date for the commit itself;
3. the content join keeps a match only where an arrival is at or after the candidate. `join_use_nulls` rejects a non-equality predicate spanning both join sides, so the comparison belongs in the filter after the join, not in `ON`.
4. Missing date on either side means no order and therefore no promotion.

**Chosen by calibration**, against the commits whose landing is independently proven — those listed by a pull request merged into the default branch. On that population the strict rule costs almost nothing, while on the population where content is the only evidence it removes a much larger share; if those removals were genuine landings they would behave like the proven population and they do not. Any tolerance window makes the discrimination worse, which is the same result the close-time recovery in #3153 produced.

Two alternatives were measured and rejected:

- **requiring every file of the commit to match** — a squash carries only the surviving state of each path, so a branch that touched one path twice leaves its intermediate versions nowhere but the branch. Most commits whose landing is proven match on only part of their files, so this loses the majority of genuine landings.
- **matching on patch id instead of content** — a squash's patch id is the combined diff of the branch and equals no single commit's, so the rule would stop working at all.

## Cost

Gold only, in a table rebuilt on every run: no migration, no descriptor bump, no bronze re-ingestion. Gold restates in full, so historical figures for the eight metrics above move on installs that have any of these commits.

A spec case must pin both directions — content that predates the commit does not promote it, content that arrives afterwards does — and the guard should be mutation-checked: removing it turns the case red.

mozhaev-dev:
A commit read as having reached the default branch whenever any one of its files happened to hold content the default branch also holds — with no question of **whether the branch already held it**. So restoring a file to the state the trunk already has, copying a file across from the trunk, or reverting your own edit was enough to mark the commit as landed. One incidental file was enough, and the whole commit went with it: every file, every line.

The comparison now takes the first time that content reached the default branch and refuses a match the branch already carried. A squash or a fast-forward is unaffected — there the content arrives afterwards, which is the whole reason the rule exists.

Which reading to use was settled by calibration against the commits whose landing a merged pull request proves on its own: taking the latest arrival instead of the first, or the author date instead of the committer date, each discriminates worse at the same cost on that population.

**Historical figures move**: the four default-branch metrics read slightly lower, the four non-default ones slightly higher, and the `branch_scope` breakdown with them. The unscoped totals — commits, lines, code lines — do not move, because only the split does. Gold only: no migration, no descriptor bump, no re-ingestion.

Found while validating `git.non_default_branch_commits` (#3059). Closes #3340.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved branch-scope metrics to identify when commits first reach the default branch.
  - Prevented commits from being counted as landed when their content was already present before they were written.
  - Improved handling of undated commits, timestamp ordering, repeated or restored content, squash merges, and multi-file commits.
  - Improved attribution of code lines to the correct commit side.

- **Documentation**
  - Clarified the rules used to determine whether commits are counted as landed.

### DIFF
--- a/src/ingestion/gold/git_default_branch_commits.sql
+++ b/src/ingestion/gold/git_default_branch_commits.sql
@@ -22,7 +22,11 @@
 -- squash rewrites history, so the originals are unreachable from the default
 -- branch for good. The second half of this model reads the content instead: a
 -- change whose object id sits on a default-branch commit at the same path
--- landed, whichever commit carries it there.
+-- landed, whichever commit carries it there — unless the default branch
+-- already held that content when the commit was written, which makes the
+-- match no evidence about the commit at all. One change is enough and scope
+-- lives on the commit, so without that test a restored file carries a whole
+-- commit across. #3340
 --
 -- INVARIANT: content can only heal what SURVIVED into the default branch. A
 -- path a branch touched more than once keeps its intermediate versions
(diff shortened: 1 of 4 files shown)


mozhaev-dev:
Good catch on the undated carrier — that was a real hole and it is fixed in bafd32f.

The ordering took the earliest arrival with `min`, and `min` skips nulls, so a carrier the source never dated was simply invisible: a later dated carrier answered for the content and promoted a commit the default branch may well have preceded. Reachable, too — the class date is a parse of the reported one, and an unparseable date leaves the commit with none.

An undated carrier now makes the whole arrival unknown, and unknown refuses instead of guessing. The spec grew the case: three carriers of one content whose earliest is the one nobody dated, with the branch commit between them staying on the other side. Skipping the undated carrier rather than obeying it turns all four cases red, and so does each of the other four mutations (no ordering test, a strict comparison, the latest arrival, the author date). Full git data-path tree green.

aleksdotbar (approved):

Issue #3362, opened 2026-09-10
Pull-request size and cycle time: correctness, contract and documentation gaps


Gregory91G:
Found while validating `git.pr_size` (#3066) and `git.pr_cycle_time_h` (#3065) against their
definitions, their consumers and the three git connectors. Part of #3039.

Everything here that needed a change to the product, a contract or the documentation shipped
in #3363, merged as `642f68981`. What is left is the observations section, where no change is
proposed. Every number below comes from the synthetic fixtures in that PR, not from any
deployed environment.

Both validation tasks now carry their own report: #3066 and #3065.

The tests named against each item state the settled contract and are in main; each was first
executed against the PRE-fix pipeline, which failed them with exactly the values these changes
correct.

One item has since been overtaken by another change on main and is marked as such below. The
Bitbucket cases also stopped carrying a seeded account binding: #3423 states the account id
for a workspace participant, so those cases now reach their person the way a real
installation does and fail if that path breaks.

## Bugs

- [x] **A Bitbucket pull request's size could include files no longer in its diff.**
  `bitbucket_cloud__pull_requests.sql` summed the per-file `pull_request_diffstat` rows for a
  request. The row key is the file path and the table is a ReplacingMergeTree, so a row is
  replaced but never removed, and nothing marked a row as belonging to the diff the request
  has now: a request re-collected after a rebase dropped a file kept that file's row and its
  lines kept counting. GitHub and GitLab were unaffected — their diff-stat streams carry one
  total per request, which a re-fetch replaces.
  **Fixed** by taking the newest `pr_updated_on` generation for the request as a whole.
  Resolving each file to its own newest row does not work: the dropped file has no newer row
  to displace it. A row with no usable stamp sorts to the epoch, so a request collected
  before the stamp existed is still summed entire.
  Test: `git_pr_size` — request 802 serves 30 where summing the stale row reaches 100.

- [x] **A merged request took its close time from `closed_at` instead of `merged_at`.**
  `closed_at` describes a request closed WITHOUT merging, and a request closed, reopened and
  then merged reports the earlier close, so a duration ended before the merge happened and
  the merge was dated to the wrong day. **Fixed** by preferring `merged_at`: in
  `github__pull_requests.sql` by #3363, and on the GitLab side by #3250, which reached the
  same precedence while rebuilding that connector.
  Blast radius, corrected: every measure dated by the close or gated on the merge —
[...]

Issue #3288, opened 2026-09-07
A pull request that breaks a data-path spec is green until the merge queue rejects it


Gregory91G:
## Summary

A pull request that breaks a data-path spec shows every check green. The five `data path (shard …)` legs are skipped on `pull_request`, and the umbrella check that the `main` ruleset requires prints a deferral instead of a verdict. The merge queue is the only thing that runs the suite, so the first sign of a broken spec is the queue rejecting an approved pull request.

**Version:** `main` at `08e011f81`
**Reproduction rate:** 1/1

## Steps to Reproduce

1. Push a pull request that changes a spec under `tests/datapath/metrics/<class>/` so that one assertion is wrong.
2. Read the checks on the pull request. All green.
   - `.github/workflows/e2e-bronze-to-api.yml:148` — `if: ${{ needs.changes.outputs.relevant == 'true' && github.event_name != 'pull_request' }}`, so the shards skip.
   - `metric-coverage-gate` then skips with them (`:298` keys off the shards' result).
   - `Run E2E suite` — the required context — short-circuits at `:371-372`: `if [ "$EVENT" = "pull_request" ]; then echo "the data-path suite runs in the merge queue; deferred there."`
3. Merge it. The queue runs the shards, the broken one fails, and the pull request is ejected.

**Expected:** a pull request that changes a data-path spec gets that spec run before it is queued.

**Actual:** it is proven only inside the queue. An approved, green pull request is ejected there, and the author has to read a `gh-readonly-queue/…` run to find out why.

## Why this is worth an issue rather than a habit

`workflow_dispatch` already gives a real verdict on a branch at zero YAML cost, and that is the current workaround. It is not discoverable: nothing on the pull request says the suite did not run, and the umbrella's message reads like a pass. The same shape bit us on #3197.

The exclusion is deliberate and measured, so this is a trade-off to re-decide, not a bug to swat. `ae16c3f26` ("ci: run the e2e lanes after the merge, not on every pull request") recorded 21 min p50 / 44 min p90 for this lane over ~50 pull-request runs and moved it to the queue on purpose; `8059d6cf1` then made this workflow own the required check and state the deferral.

## How small the fix can be

Three sizes, all in `.github/workflows/e2e-bronze-to-api.yml`:

| Lines | What it buys |
|---|---|
[...]

Issue #3066, opened 2026-09-02
Validate metric: git.pr_size


hello1101n:
Part of #3039.

- [x] What does `git.pr_size` do? Does the definition make sense?
- [x] Where is it used? Mark used / not used.
- [x] Write an e2e test: specific bronze → transform → request in the API, for each scenario.
- [x] Check what the v1, f1 and dev stands return for it. Does it make sense? Verify with different data.
  - [x] Counted correctly for github, gitlab and bitbucket?

Gregory91G:
## `git.pr_size` — validation report

Validated together with #3065. Everything that needed a change is in #3363, merged as
`642f68981`; what stays open is the observations section of #3362. Every value below comes
from the synthetic fixtures in that PR.

### What it does — and does the definition make sense

`median(pr_change_size)`: one observation per pull request the person authored, valued at
`lines_added + lines_removed`, dated by the day the request was **opened** (UTC) — unlike its
close-dated neighbours in the same group — and counted whatever state the request reached,
open, merged or declined. A reviewability metric, so `lower_is_better` stays: how large a
request to open is the author's own choice.

The definition holds. Three things it left unsaid are now in the registry explanation: the
dating day, the timezone, and the whole rule for a zero — a source that never reported line
counts contributes no value, while counts reported as zero (a rename or a mode change)
contribute an observed diff of no lines.

### Where it is used — USED

- Metric card in the `git_output` collection: `period`, `peer`, `histogram`.
- `Development ▸ Overview` and `▸ Flow` "Typical values (median)" tiles, the "How large pull
  requests were" event histogram, and two team-level lenses.
- `direction` feeds personal attention flagging and peer standing, not just a label.
- No card offers a `breakdown`; the registry still declares the dimensions, so the API serves
  one when asked.

### Defects found, and what changed

1. **A Bitbucket request's size could include files no longer in its diff.** The per-file
   diffstat rows are a ReplacingMergeTree keyed on the file path, so a file dropped by a
   rebase kept its row and kept counting. Fixed by taking the newest `pr_updated_on`
   generation for the request as a whole — resolving each file to its own newest row does not
   work, because the dropped file has no newer row to displace it.
2. **A collected zero-line diff was discarded** by a `> 0` guard in gold. This measure was the
   NULL-versus-zero marker's only live consumer, so the distinction served nothing, and
   dropping zeros biased a `lower_is_better` metric upward. The guard is now `IS NOT NULL`.
[...]

Issue #3065, opened 2026-09-02
Validate metric: git.pr_cycle_time_h


hello1101n:
Part of #3039.

- [x] What does `git.pr_cycle_time_h` do? Does the definition make sense?
- [x] Where is it used? Mark used / not used.
- [x] Write an e2e test: specific bronze → transform → request in the API, for each scenario.
- [x] Check what the v1, f1 and dev stands return for it. Does it make sense? Verify with different data.
  - [x] Counted correctly for github, gitlab and bitbucket?

Gregory91G:
## `git.pr_cycle_time_h` — validation report

Validated together with #3066. Everything that needed a change is in #3363, merged as
`642f68981`; what stays open is the observations section of #3362. Every value below comes
from the synthetic fixtures in that PR.

### What it does — and does the definition make sense

`median(pr_cycle_hours)`: hours from opening a pull request to merging it, taken only for
requests in state `MERGED` that carry both timestamps and a forward-running interval, and
**dated by the merge** (UTC). `git.pr_cycle_time_p75_h` reads the same measure at q=0.75.

The definition holds, with one correction to what the number may say about a person. Cycle
time is mostly queueing — for a reviewer, for a build, for someone to press merge — so it
describes the path a change travels, not the person who opened it. In this product
`direction` is not a label: it feeds personal attention flagging, deciding whether someone is
ranked bottom-quartile and marked as moving the wrong way. Both cycle metrics are therefore
`neutral` now, which keeps the number, the distribution and the heatmap column and drops only
the verdict; `peer-standing.ts` already had `neutral_direction` as a first-class reason, so no
consumer changed.

The registry explanation now also states the merge-dating, the timezone, and that a source
reporting no merge time contributes no duration.

### Where it is used — USED

- Metric card in the `git_output` collection: `period`, `peer`, `histogram`; the p75 with
  `period` and `peer`.
- One of the ten heatmap columns.
- `Development ▸ Overview` and `▸ Flow` "Typical values (median)" tiles, the "How long pull
  requests stayed open" event histogram, and two team-level lenses.
- No card offers a `breakdown`; the registry still declares the dimensions, so the API serves
  one when asked.

### Defects found, and what changed

1. **A merged request took its close time from `closed_at` instead of `merged_at`.**
   `closed_at` is not a reliable merge instant: for a request that was closed, reopened and
   later merged, it can represent the earlier non-merge close — so the duration ended before
[...]

PR #3363, opened 2026-09-10
fix(git): correct pull-request size and cycle-time semantics

Gregory91G:
Part of #3362 — every finding there that needed a code, contract or documentation change.
What is left in the issue is observations with no change proposed.

**Why.** Validating `git.pr_size` (#3066) and `git.pr_cycle_time_h` (#3065) turned up three
defects and four unsettled semantics in the pull-request family, and neither metric had a spec
that would have caught any of them.

**What changed.** Two staging fixes and one gold fix, a new class-contract column with its
migration and staging heal, the median aggregation in every view, one metric's directionality,
the passports and the git silver README — plus a dedicated spec per metric covering GitHub,
GitLab and Bitbucket.

**A recovered Bitbucket close time counts the merge and yields no duration.** The recovery
(#3051, #3328) corroborates which DAY a merge landed on, not which instant, so the class
contract carries `closed_on_reported` beside `closed_on` and every interval measure reads the
former. `prs_merged` is untouched, so no merge is lost. Reversible by pointing the durations
back at `closed_on`.

**The rollout backfills GitHub and GitLab, and deliberately not Bitbucket.** Their close time
is the source's own, so the migration and the heal copy it and no historical duration is lost
between deploy and the next sync. Promoting a recovered Bitbucket time is the error the column
exists to prevent.

**`median` is the textbook median in every view, the peer one included.**
`quantileExactInclusive`, not `quantileExact` — the old index answered with the upper middle
value on an even sample, so a person with two observations was reported at the slower of them.
It reaches all twelve median metrics across git, ci and tasks. A percentile keeps
`quantileExact`, having no second definition to match, and peer `p25`/`p75` stay order
statistics over the cohort.

**`git.pr_cycle_time_h` and its p75 are `neutral`, not `lower_is_better`.** `direction` feeds
personal attention flagging here, and cycle time is mostly queueing for a reviewer, a build or
someone's merge button. `git.pr_size` keeps `lower_is_better` — how large a request to open is
the author's own choice.

**A collected zero-line diff is a value, not an absence.** Safe only once no connector writes a
literal `0` for counts it never collected, which for GitLab is now settled by #3250.

**The GitLab-specific NULL assertion is deleted rather than updated.** #3250 gave the connector
a real diff-stats stream, so a dbt test pinning those three columns to NULL for `insight_gitlab`
now asserts the opposite of the contract; the generic macro behind it had no other caller.

**Bitbucket reaches its person through #3423, not through a seeded binding.** Both new specs
drop their `identity_accounts` row and seed a `commit_authors` edge instead, so a Bitbucket
assertion fails if that path breaks. GitLab keeps its seeded binding — it exposes an address
only to an admin-scoped token, which is a different question.

**This branch carries no GitLab diff-stat change** — #3250 owns that, and `gitlab__pull_requests.sql`
here is main's file plus `closed_on_reported`. `union_by_tag` is a positional UNION, so all
three contributors place that column at the same position, straight after `closed_on`; the
contract therefore ships with a numbered silver migration and a guarded staging heal, and the
generated `connectors-ddl` snapshot is regenerated rather than hand-edited. Seven expectations
in five neighbouring specs read the old upper-middle tie rule; each is recomputed from its own
fixture, not copied off the new output. The data-path suite does not run on pull requests
(#3288), so PR CI will not exercise any of it.

**Out of scope.** The observation left in #3362: the same directionality question on four
reviewer-latency metrics that are attributed to the request's author.

**Verified.** On a compose test-stand with the backend compiled from this tree: the `git` tree
runs **144 passed, 0 failed**, `ci` **14 passed**, `tasks` **56 passed**, and the schema-drift
guard **148 passed, 3 skipped** — every median metric across the three domains, both new specs,
and the cases that state the corrected contracts. Against a silver-seeded instance
`test_drilldown.py` runs **136 passed, 1 skipped, 12 xfailed**, the skip and the xfails all
pre-existing (#2989), exercising the exact-median reconciliation over every `EXACT_MEDIAN`
metric the matrix covers. The Bitbucket cases in both new specs reach their person through the
account binding rather than a seeded one, so they now fail if that path breaks.

Backend and static gates: `cargo test -p analytics` **865 passed / 0 failed** including the
passport drift test; `cargo fmt --all --check` and `cargo clippy -p analytics --all-targets --
-D warnings` clean; `dbt parse` clean; `ruff`/`pycln` clean on every touched file;
`git diff --check` clean. Workspace-wide clippy could not run here (`cmake` absent, so
`rdkafka-sys` will not build), so CI's Rust job is the authority for the crates this branch
does not touch.

The `git` and `tasks` trees were completed in slices rather than one process each: a
long-lived pytest aborts inside a C extension on this machine, a local fault unrelated to the
suite — every module passes and the slice counts sum to the tree.

The migration and the heal were exercised on their UPGRADE path, not just a fresh create. The
class table and all three staging projections were forced back to the pre-change shape and
seeded with a merged request per connector; the shipped migration and the real
`apply-ch-migrations.sh` then put the column at position 19, directly after `closed_on`, and
backfilled the GitHub and GitLab rows while leaving the Bitbucket one null. A second pass is a
no-op, and a row whose reported time had been set to something other than `closed_on` keeps it.
The `connectors-ddl` snapshot differs from main by exactly the one expected line.


cyberantonz (approved):

Gregory91G:
@coderabbitai review

Issue #3368, opened 2026-09-10
CI audit: repeated compilation, queue pressure, coverage gaps — consolidated checklist


cyberantonz:
Consolidated findings from a full CI review (all 26 workflow files, cross-checked against the tree and recent Actions run data). Grouped by theme; checked items are already fixed, with the fixing change linked.

## A. Repeated compilation & broken caching

The backend workspace is compiled in 6 workflows under 6 mutually incompatible cargo profiles (dev, dev+coverage, release, e2e, musl-release, docker-release); a `Cargo.lock` bump triggers ~35–40 independent cold compiles of the same dependency graph across PR → merge queue → push.

- [ ] **R1 — `build-backend` (E2E bronze to API) has zero caching.** `CARGO_TARGET_DIR` is a docker named volume (fresh every run) and the cargo registry lives inside the container; no rust-cache/sccache/layer cache. Move the target dir to a bind-mounted path and add `Swatinem/rust-cache` (save on `merge_group`/`push`).
- [x] **Seed step rebaked backend images in every data-path shard** — `compose run --build seed-sample` also rebuilt every `depends_on` image, per shard. Fixed by #3342 (`--no-deps`).
- [ ] **R2 — each of the 5 data-path shards still cold-builds the gateway image** (musl `cargo build --release -p routegen`): `test_stand_prime_dev_images` skips the gateway and the compose `gateway` service has no `profiles:`, so `up -d` bakes it. Prime the gateway like the other services (verify first: `routes.yaml` is a bind-mount compiled at container start, so the "baked at image build" rationale for the exclusion looks stale) or build it once in `build-backend`.
- [ ] **R3 — e2e-stand source fallback compiles the workspace ~5× in one job**: when `backend=true` but none of the four per-service filters match (e.g. a `src/backend/libs/**` change), `stand-runner` runs `--build-backend` without `--skip-build`, so the e2e compile runs *and* compose bakes all four `:dev` images at `--release`. Widen the per-service filters (e2e-stand.yml `changes` job) to cover `libs/**`, or at minimum pass `--skip-build` + prime images on that path too.
- [ ] **R4 — openapi-specs fans 5 cold cargo builds per PR and uses a floating toolchain.** All five drift-gate jobs omit `toolchain:` (rolling `stable`) while every other lane pins 1.95.0, and the analytics job sets no rust-cache key. Pin the toolchain and build all five binaries in one job.
[...]

PR #3386, opened 2026-09-11
fix(e2e): stop pinned test-stand runs from compiling insight-v3-core from source

cyberantonz:
Part of #3368.

**Why.** Every `test-stand` bring-up without `--build` compiled insight-v3-core from source — the service is missing from the pinned-backend list, so each of the five data-path shards of every merge-queue entry paid a full release-workspace image build.

**What changed.** insight-v3-core joins `TEST_STAND_PINNED_BACKENDS`, so pinned runs pull `ghcr.io/constructorfabric/insight-v3-core:<appVersion>` like the other four; e2e-stand's merge-group prebuilt path loads it through prebuilt_images.py's existing `unchanged` policy.

**A mechanical guard now fails bring-up before compose builds anything.** After pinning, every service in the compose up set that still carries a `build:` section must resolve to a locally present image — so the next backend service added to docker-compose.yml without a pin entry trips loudly instead of compiling silently. Skipped when `--build`/`--build-backend` asks for a source build.

**Skip-build priming deliberately excludes insight-v3-core.** Like the gateway, its code runs from the image rather than a bind-mounted binary, so under `--build-backend` only an image built from this tree exercises a source change.

**Out of scope.** build-images exports no insight-v3-core e2e-prebuilt artifact, so a merge-group entry touching insight-v3-core alongside a prebuilt service runs the published image, not the PR's, and `--build-backend --skip-build` shards still bake it per shard; both want that export as a follow-up.

**Verified.** `bash -n` and shellcheck (no new findings); pulled the chart-pinned tag from GHCR; `docker compose config` resolves `INSIGHT_V3_CORE_IMAGE`; guard exercised both ways — all five pins pass, removing the v3-core pin trips:

```
ERROR: `up` would compile these services from source, in a run that did
       not ask for a source build:
         insight-v3-core -> insight-v3-core:dev
         insight-v3-core-migrate -> insight-v3-core:dev
       Every backend service must run a published image here. Add the
       service's "<VAR>|<chart path>|<name>" entry to
       TEST_STAND_PINNED_BACKENDS, or pass --build to build from source
       deliberately.
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved test-environment startup by ensuring the Insight V3 Core service uses the correct prebuilt or locally built image.
  - Added validation to prevent services from unexpectedly compiling from source when required images are unavailable.

- **Chores**
  - Updated test-stand setup, image preparation, and help text to consistently support the expanded backend service set.

### DIFF
--- a/.github/workflows/e2e-stand.yml
+++ b/.github/workflows/e2e-stand.yml
@@ -304,7 +304,8 @@ jobs:
             --image "insight-analytics:${ANALYTICS_POLICY}" \
             --image "insight-authenticator:${AUTHENTICATOR_POLICY}" \
             --image "insight-identity-resolution:${IDENTITY_POLICY}" \
-            --image "insight-gateway:${GATEWAY_POLICY}"
+            --image "insight-gateway:${GATEWAY_POLICY}" \
+            --image "insight-v3-core:unchanged" # build-images exports no insight-v3-core artifact — always the published image
 
       # Restore-only warm start for the --build-backend fallback: the cache is
       # saved by e2e-bronze-to-api.yml's push-to-main run; nothing here saves.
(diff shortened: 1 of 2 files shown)


hello1101n (approved):

mozhaev-dev (approved):

cyberantonz:
The CI failures were a wrong image-name derivation: `test_stand_pinned_image` prefixes `insight-` itself, so the pin entry's name field must be `v3-core` — the previous field produced `insight-insight-v3-core` (manifest unknown). Fixed in `e0cc355cc` (pin entry + the prime-skip case), verified by pulling the derived ref `ghcr.io/constructorfabric/insight-v3-core:2026.09.11.18.58-6245304` successfully. Branch rebased onto current main.

PR #3424, opened 2026-09-14
fix(frontend): stop the retry story depending on how many requests the view makes

hello1101n:
**Why.** `TestErrorAndRetry` fails on `main`. It asserts the error card, but the recovering handler spent a budget of one failure per request — and since the vitest 5 / Storybook 10.6 bump the view issues two requests before `play` runs, so the first spends the 503, the second serves data, and the error card never renders.

**What changed.** The handler fails while `state.failing` is set, and the story clears the flag before clicking Retry. That is what the story means — fail until I retry — and it no longer depends on how many requests reaching the error state takes.

**Why it was invisible.** The frontend job is gated by change detection, so a backend-only commit skips it. The failure has been sitting on `main` since the bump, waiting for the next PR that touches a frontend file.

**Bisect.** Passes at `23f8f0906`, fails at `2ecb2784d` (the npm group bump). Raising the budget to two made it pass again, which is what identified the second request.

**Verified.** `pnpm test` 255 files / 2579 tests, `pnpm test:storybook:ci` 8 files / 32 tests, `pnpm typecheck` and `pnpm lint` — all clean locally, matching CI's 263 / 2611 totals.

## Summary by CodeRabbit

- **Tests**
  - Improved the metric time-series retry scenario so recovery is triggered explicitly when the retry action is selected.
  - Made error-state simulations consistent regardless of how many requests the view makes before displaying the retry option.


hello1101n:
Measured the request count directly rather than inferring it, and the finding is bigger than the test.

I instrumented the MSW handler to record every `POST /metric-results` the view makes before `play` clicks Retry:

| commit | requests | bodies |
|---|---|---|
| `23f8f0906` (pre-bump) | 1 | grouped query, 531 bytes |
| `2ecb2784d` (the npm bump) | 2 | ungrouped `dimensions: []`, 343 bytes — **then** the grouped one, 531 bytes |

So the bump did not just change timing. The view now issues an **extra, different** request on mount: an ungrouped query that fires before the grouped one. No view code changed between those commits — only dependencies.

That is a real behaviour change for anyone opening this widget, not a test artifact, and the old test was catching it by accident. This PR stops it catching it, because a test for "error card, then Retry" should not be the thing that pins request count.

Worth someone who owns this widget deciding whether the extra ungrouped request is intended. If it is not, it wants its own fix plus a test that asserts request count on purpose rather than as a side effect.

mitasovr (approved):

PR #3425, opened 2026-09-14
feat(clickhouse): deliver operator value-map rows per environment

cyberantonz:
**Why.** Rows in `config.task_value_map` have no delivery path — the design assumes a hand-written INSERT and the authoring surface is deferred, so per-environment issue-type decisions cannot ship.

**What changed.** A post-upgrade hook Job (weight 300, after the migrate job) reads a per-environment TSV from a ConfigMap and inserts it. Operators author business columns only; `valid_from`/`recorded_at`/`recorded_by` are loader-owned, and the two timestamps are computed server-side rather than from the pod clock.

**Insert-only by construction.** The table is a ReplacingMergeTree whose key includes `valid_from` and `recorded_at`, so a plain INSERT creates a newer version that silently wins over a stored decision. The loader reads the stored business keys, inserts only the ones with no row — including keys whose stored row is a retraction, since re-seeding one would resurrect a removed mapping. Editing an already-loaded line therefore does nothing; that is deliberate, and the README says so.

**Python, not shell.** `src/ingestion/scripts/` is predominantly Python with a pytest suite, and parsing and validating operator input is exactly the part that should be unit-tested: 30 tests cover the bundle format, the in-bundle duplicate collapse, and the key filtering, with no ClickHouse needed.

**Off by default.** `clickhouse.taskValueMap.enabled=false`; the ConfigMap mount is optional and an empty directory is a no-op.

**Verified.** Against a local ClickHouse holding a real `config.task_value_map`: fresh insert works; a re-run inserts nothing and adds no physical version; a stored row with a different `canonical_value` is neither overwritten nor superseded; a new key alongside it is inserted; a key whose only stored row is a retraction stays untouched. A malformed line fails the hook naming `file:line`. Scripts suite 110 passed; `helm template` renders the Job only when enabled and the toolbox image is set; `helm lint` and ruff clean.

## Summary by CodeRabbit

* **New Features**
  * Added optional GitOps support for seeding ClickHouse task-value mappings from TSV configuration files.
  * Seeds only missing mappings, preserving existing decisions and safely handling empty or unavailable bundles.
  * Added configuration for enabling the feature, naming the ConfigMap, and recording the seed source.
  * Added documentation covering the TSV format and deployment workflow.

* **Tests**
  * Added coverage for parsing, validation, deduplication, existing-key handling, and SQL generation.


cyberantonz:
Superseded: issue-type decisions move to a dedicated `config.task_issue_types` table instead of the generic value map, so the loader is being rebuilt against that shape. Replacement PR follows.

Issue #3075, opened 2026-09-02
Validate metric: tasks.bugs_fixed


hello1101n:
Part of #3039.

- [ ] What does `tasks.bugs_fixed` do? Does the definition make sense?
- [ ] Where is it used? Mark used / not used.
- [ ] Write an e2e test: specific bronze → transform → request in the API, for each scenario.
- [ ] Check what the v1, f1 and dev stands return for it. Does it make sense? Verify with different data.

cyberantonz:
## Validated on a deployed stand (gold layer, read-only)

Checked the ClickHouse gold layer of a deployed stand running a build that includes #3483. Every query was a `SELECT` in a server-enforced read-only session.

### Does the definition make sense? Yes, with one product question

`tasks.bugs_fixed` sums the `bugs_fixed` measure: one unit per closed issue whose `issue_kind` resolves to `bug`. It is a subset of `tasks.closed`, not a separate total, which the registry `explanation` states.

One question worth settling, since it is a definition choice rather than a data fault: `issue_kind` comes from an operator-configured mapping row, falling back to a tenant default and then to `unknown` (`gold/task_issue_state.sql`). An operator can therefore map incident-, problem- or vulnerability-style issue types onto the `bug` kind, and those closures then count toward `bugs_fixed` alongside plain bug types. Whether ITSM and security-scanner issue types belong in a developer-facing bug metric is a product call, and it moves both this metric and `tasks.bugs_ratio` materially. Worth a documented default.

### Where is it used? Used

- `src/frontend/src/lib/insight/groups.ts` — quality group, and a member of the task-kind metric partition.
- `tests/stand/api/analytics/drilldown_matrix.py` — drilldown expectation at `Tier.EXACT_COUNT`.

### e2e coverage: already present

No new test needed; the bronze to transform to API path is covered by
`tests/datapath/metrics/tasks/tasks_bugs.test.yaml`,
`tasks_bugs_lifecycle.test.yaml`,
`jira_tasks_bugs_default_kind.test.yaml` (tenant-default fallback),
`jira_tasks_bugs_value_map_precedence.test.yaml` and
`github_tasks_bugs_value_map_precedence.test.yaml` (mapping-row precedence over the default).

### Gold-layer findings

- Evidence is exactly one row per closed bug: row count, distinct `record_id` and summed contribution all agree, so there is no fan-out or double emission.
- Evidence totals reconcile exactly with the observation totals.
[...]

PR #3426, opened 2026-09-14
feat(jira): classify issue types by id, no name fallback

cyberantonz:
Part of #3075. **Draft — must not merge until the value maps are loaded on every stand with Jira** (#3425 delivers them). Merging earlier sends every unmapped type to `unknown` and zeroes bug metrics.

**Why.** Classification fell back to name lists, so renaming a type re-wrote history into the wrong bucket: `Bug` renamed to `Task` moves closed bugs into the non-bug side, silently and retroactively.

**What changed.** The Jira map keys on the issue-type id, which never changes, and the name fallback is gone — an unmapped id is `unknown`. GitHub keeps its own keying and the shared name lists.

**Renames become visible instead of silent.** `value_display` holds the name as seen when the decision was made, and a new `data_quality` check flags a type whose catalog name has drifted from it. Jira type mappings also enter `assert_bound_values_are_mapped`, which the `issuetype`/`type` field_id mismatch had excluded.

**Cost this accepts.** A new team-managed project mints fresh type ids that need explicit rows; unmapped means `unknown`, never a wrong class.

**Verified.** Seven datapath specs now carry explicit map rows; the precedence spec gains a renamed-type case (same id, different catalog name, mapping holds). Full tasks tree green (52 passed); pre-commit clean.


cyberantonz:
Superseded: the classification now reads a dedicated `config.task_issue_types` table rather than the generic value map. Replacement PR follows once the table and its loader land.

PR #3433, opened 2026-09-14
ci(build-images): let a dispatch rebuild every layer from scratch

### LINKED ISSUE #3432: Image builds always hit the cached apt layer, so a fixed base package never reaches a published image
**Problem.** Every image build reads a registry layer cache (`$BUILDCACHE:<component>-<arch>`). A layer's cache key is its command plus its parent, so `RUN apt-get …` keeps hitting cache for as long as the base image digest and the command text stay unchanged. A base-distribution package that has been fixed upstream therefore never reaches the image, however many times the build re-runs.

**Impact.** When the image CVE gate sees such a package it fails the `merge-<component>` job. That job is the one that assembles and publishes the multi-arch manifest, so the component simply stops publishing while the rest of the build stays green — easy to miss, and the change that triggered the build never reaches any environment. It currently blocks `toolbox` and `git-cli-proxy`.

**Evidence.** From a `build-images` run, the toolbox build log:

```
#7 importing cache manifest from ghcr.io/constructorfabric/insight-buildcache:toolbox-amd64
#9 [2/9] RUN apt-get update && apt-get upgrade -y --no-install-recommends && ...
#9 CACHED
```

**The Dockerfiles are already correct.** Checked against the current base images: `apt-get upgrade` in the toolbox image, and `apt-get install git` in git-cli-proxy, both resolve the fixed package version straight from the archive. No pin and no Dockerfile change are needed — the layer only has to actually run.

**There is no way to make it run.**

- `no-cache:` and `pull: true` appear nowhere in `.github/workflows/`.
- `workflow_dispatch` exposes a single input, `frontend_only`.
- A source commit does not help: `COPY` sits after the apt layer, so it invalidates only what follows it.
- Re-running the failed merge job does nothing — it scans and joins per-arch digests that already exist.
- `pull: true` alone would not help either: the hit is on the `RUN` layer, whose key follows the base digest, and that digest has not changed.

**Proposed fix.** Add a `no_cache` boolean input to `workflow_dispatch` and pass it to every `docker/build-push-action` step, so a dispatch can rebuild each layer from scratch. This is the only lever that reaches those layers.

**Alternative considered.** Deleting the cache tags from the registry has the same effect, but it is a manual action needing package-delete rights, leaves no trace in the repository, and has to be repeated by hand every time.

Gregory91G:
Closes #3432.

**Why.** A base package fixed upstream never reaches a published image: the apt layer keeps hitting the registry cache, the CVE gate then fails the merge job, and that component silently stops publishing while the rest of the build stays green.

**What changed.** A `no_cache` boolean on `workflow_dispatch`, threaded into all 13 `docker/build-push-action` steps — the only lever that reaches a cached layer.

**Dispatch-only, and `cache-to` still runs.** The expression is false on every other event, so pull requests, the merge queue and ordinary pushes read the cache exactly as before; on a no-cache dispatch the rebuilt layers become the cache later builds read, so one run heals it for everyone.

**Left alone: the Dockerfiles.** Checked against the current base images — `apt-get upgrade` in toolbox and `apt-get install git` in git-cli-proxy already resolve the fixed package straight from the archive. The layer only has to run.

**Verified.** actionlint with `main` as the baseline: 33 findings before, 33 after, none new. `build-images` has no trigger on this branch — its `push` and `pull_request` filters are `main` and `release-**` — so the `no_cache` path first executes on a dispatch after merge, not on this PR.

## Summary by CodeRabbit

- **New Features**
  - Added an optional manual workflow setting to force Docker images to rebuild without using cached layers.
  - Applied the no-cache option consistently across backend, frontend, toolbox, seed, and connector image builds.


cyberantonz (approved):
D

PR #3439, opened 2026-09-14
fix(analytics): zero-fill sum and distinct-count metrics where the source covers the entity

### LINKED ISSUE #3438: Sum and distinct-count metrics serve null where the contract says zero
`sum` and `distinct_count` metrics serve `null` for an entity the source **does** cover but which produced no event of the measure in the period. The code states the opposite in two places:

- `metric_definitions/definition.rs`, `ComputationSpec::DistinctCount` — *"Zero-filled like `Sum` — no subjects is a genuine zero, not unknown."*
- `metric_results/compiler.rs`, `item_value_expr` — *"The builder zero-fills distinct counts … exactly as it does sums"*, deliberately contrasted with *"the builder never zero-fills medians (honest-null)"*.

No such zero-fill exists. `build_period_view` takes the definition as `_def` and passes the SQL NULL straight through, so every computation kind is honest-null. Nothing asserts a served value either way — the tests pin only that the SQL emits `…OrNull`, which is the correct *input* to a zero-fill that never happens.

## Reproduction

A person the git source covers for the period — they have observations for a sibling measure of the same source — who merged no pull request:

```
git.prs_created   <n>     the source observed them
git.prs_merged    null    <- 0 is the truth: they merged nothing
```

The same request distinguishes the two zeros correctly for a median: `git.pr_size` serves `0` where line counts were reported as zero and `null` where the source never reported them. Only the sum/distinct-count kinds collapse "zero events" onto "no evidence".

## Scope

62 `sum` + 3 `distinct_count` of 105 metrics. `median`, `percentile`, `stddev` and `ratio` are honest-null by design and behave correctly.

## Why "the entity appeared in the batch" is not the signal

The period batch scans only the requested metrics' `(source_key, measure_key)` pairs, so whether an entity produces a row at all depends on which **other** metrics share the request. A metric's answer must not change with the company its request keeps.

## Expected

`0` where the metric's own source observed the entity in the period, `null` where it did not — which keeps `docs/product/SCENARIOS.md` rule 1 ("never a zero for missing data") intact, because an evidence gap stays a gap.

Gregory91G:
Closes #3438.

**Why.** A `sum` or `distinct_count` for an entity the source covers, but which produced no event of the measure, serves `null` where two comments in the same tree promise `0` — so "merged nothing this period" is indistinguishable from "no git data at all". This was originally reproduced on `git.prs_merged`.

**What changed.** `build_period_view` now reads the definition it was already handed as `_def` and fills a *covered* absence with `0`; every other computation kind stays honest-null.

**Coverage is read off the source, not off the batch.** The period batch scans only the requested `(source_key, measure_key)` pairs, so "the entity produced a row" would make one metric's answer depend on which others shared its request. Instead: one `countIf(source_key = ?)` column per zero-filling item, over a scan widened to those sources. A batch that zero-fills nothing compiles unchanged, and an absent coverage column reads as no coverage.

**Out of scope.** Peer pools keep every aggregate `…OrNull` — a fabricated zero must never enter another person's distribution.

**Verified.** `cargo test -p analytics`: 869 pass, 6 new over covered-zero, uncovered-null, measured-zero and per-window coverage; fmt and clippy (pedantic) clean. The emitted SQL shape was also run by hand against a local stand, where the widened scan returned the same aggregate as the pair-scoped one.


cyberantonz (approved):

Gregory91G:
Closing: this was based on a misreading of the contract.

The PR's own checks skipped every `Data path` shard — the diff touched only Rust analytics — so the data-path suite first ran in the merge queue, and four of its five shards failed on the same assertion:

```
tasks.bugs_fixed/period     value: expected None, got 0.0
collab.channel_posts/period value: expected None, got 0.0
ai.seat_cost/period         value: expected None, got 0.0
ai.cost/period              value: expected None, got 0.0
```

`tests/datapath/` asserts, across three sources, that an absent observation means `null` **even for an entity the source demonstrably covers** — `test_zero_bug_member_has_no_observation` pins `tasks.bugs_fixed = None` for a member whose `tasks.closed = 2` in the same request and period.

A blanket zero-fill also breaks metrics whose absence is not "no events happened". `ai.seat_cost` is a `sum`, and `test_ai_seat_cost_for_a_seat_with_no_tier` ("a seat with no tier at all was never billed for") expects `null`: what is missing there is the price binding, not the events. Reporting `0` would be a false money figure — the exact thing `docs/product/SCENARIOS.md` rule 1 forbids.

So neither the computation kind nor source coverage separates "the events were counted and there were none" from "the value could not be computed", and the executable contract already answers the question: `null`.

What remains is a documentation defect, not a behaviour one — `definition.rs` (`ComputationSpec::DistinctCount`) and `compiler.rs` (`item_value_expr`) still describe a builder zero-fill that has never existed. Leaving that for a separate change rather than carrying this branch.

PR #3451, opened 2026-09-16
fix(metric): make json metrics compile, store and run as written

hello1101n:
**Why.** A json metric could be stored and still never run: an array selector broke on a nullable payload column, a key naming anything but an identifier was refused, and `put_metric` answered "stored" for bodies that cannot compile.

**What changed.** The selector unwraps its payload before reading the array, a json key is escaped into its SQL literal instead of being charset-restricted, and `check_metric` now reports the deserialize error and compiles the query before storing it.

**Escaping replaces the charset check rather than widening it.** The charset was the guard against interpolating a key into SQL; the injection test now asserts a neutralised literal instead of a refusal. Reversible on its own.

**Out of scope.** `put_metric` still stores a metric naming a table or column that does not exist — that needs a catalog lookup on every write.

**Verified.** `cargo test -p insight-v3-core` 336 pass, `cargo clippy --all-targets` clean. Two new tests run their SQL against a local ClickHouse 25.7 container; each goes red when its fix is reverted.

## Summary by CodeRabbit

- **New Features**
  - Metric queries can now extract values from JSON arrays stored in nullable columns.
  - JSON keys containing spaces, symbols, or other non-standard characters are supported.

- **Bug Fixes**
  - Improved handling of JSON paths, including safe escaping of special characters.
  - Invalid metric definitions are now rejected before storage, with clearer validation for unsupported or malformed queries.


cyberantonz (approved):

hello1101n:
@coderabbitai not adding docstrings for the coverage check. This repo sets the opposite rule in `AGENTS.md`: "Doc comments only for genuinely public/external APIs; keep them brief", and "`///` doc comments live in shared library crates; binaries and services get none. No `missing_docs` enforcement anywhere."

`insight-v3-core` is a service, so the 80% default threshold does not apply here. Leaving that check unresolved on purpose.

Issue #3408, opened 2026-09-14
Create datasets


hello1101n:
Refactor the current API for insight-v3-core to upload new data.

**Acceptance criteria:**
1. Should be able to create dataset (IU , API) (create by admin only, and in API only using specific token)
2. Should be able to remove dataset (IU , API)
3. Should be able view data set in UI
4. LLM should be able to get dataset data

**Etc:**
- This is most probably should be a part of insight-v3-core
- In future data will be restricted by datasets
- Review scratches and PRD from @aleksdotbar 
- Required before implementation create feature doc for review and send it to review

**Q:**
1. What to do with analytics data? Should they also be datasets? Think about it

PR #3462, opened 2026-09-17
feat(insight-v3): add declared datasets and dataset-backed metrics

mozhaev-dev:
Custom metrics now read declared datasets instead of naming warehouse tables and repeating JSON extraction rules. This adds the backend dataset lifecycle and reorganizes insight-v3-core into API, domain, store, chat, configuration, and MCP layers.

## Changes

- Store dataset declarations and operation ownership in MariaDB, with a separate ClickHouse table for each dataset generation.
- Add administrator-only dataset management, declaration validation, record previews, dependency lookup, and guarded replacement/removal.
- Route ingestion to ready datasets and compile metrics from declared fields, types, row identity, and effective clocks. Collapse repeated records before filtering and windowing.
- Expose datasets to the assistant and MCP, validate metric definitions before storing them, and update the OpenAPI contract.
- Pin removal to the table captured when taking the operation; preserve tables when publication has an uncertain outcome.

## Compatibility and scope

This is a breaking change to custom ingestion and metric definitions: metrics naming physical tables are refused. There is no automatic migration or adoption of legacy stream tables; callers must declare datasets, resend records, and rewrite metric definitions. Portal UI changes are separate. Dataset administration uses the administrator role; a separate administration token is not implemented here.

## Validation

- `cargo test -p insight-v3-core --quiet`: 501 passed.
- `cargo fmt -p insight-v3-core --check`: passed.
- `RUSTFLAGS='-D warnings' cargo clippy -p insight-v3-core --all-targets --quiet`: passed.
- `git diff --check`: passed.

MariaDB concurrency and ambiguous-commit behavior were reviewed in code; the test suite does not exercise these against a running MariaDB instance.

Related to #3408.


mozhaev-dev:
Two service-side companions for the frontend (#3475) are now on this branch: `?limit` and a total on the records preview, and `GET /v1/{kind}/{name}/dependents` — every definition that names this one, as the dataset endpoint already answered for the metrics reading it.

cyberantonz (approved):

PR #3465, opened 2026-09-17
feat(task): name the task-kind metrics as a partition

cyberantonz:
**Why.** The members heatmap put Issues, Bugs and Non-bug side by side as if they were a partition, but the two subsets do not add up to the total: a closed issue whose type maps to no kind joins neither column and is accounted for nowhere. In the `tasks_closed` fixture, 2 of one person's 6 closes sit in no column at all.

**What changed.** The measure `closed_non_bug` becomes `closed_task` (metric `tasks.closed_task`, "Tasks closed"), and a new `closed_unknown_type` measure feeds `tasks.closed_unknown` ("Unclassified issues closed"). The gold ARRAY JOIN in `task_metric_evidence` mints exactly one kind measure per closed issue, so `bugs_fixed + closed_task + closed_unknown_type == tasks_closed` per entity-date. The heatmap now reads Issues | Bugs | Tasks | Unclassified.

**The dbt subset guard now asserts the equality, not a bound.** With the unknown measure present the three subsets are an exact partition, so a kind that falls through every branch is a failure rather than an under-count nobody sees.

**Rollout.** The rename leaves the old `tasks.closed_non_bug` row in an already-seeded analytics database. No migration is needed: `disable_missing_builtin_rows` flips `is_enabled = FALSE` at startup for any builtin key the registry no longer names — the same path that handled the removal of `tasks.closed_unknown_type` in c31b906, which also shipped without one. Verified on a local stand: after the sync the old key is present and disabled, the two new keys enabled.

**Out of scope.** Resolution classes (`closed_fixed` / `closed_duplicate` / `closed_wontfix`) stay in the open #3463.

**Verified.** `dbt parse` clean; the RMT read-dedup audit byte-identical to main; analytics `cargo test` 865 passed including the passports-sync test; seed suite 264 passed; frontend `tsc` and `eslint` clean with 69 vitest tests; the `tests/datapath/metrics/tasks` tree green on a local compose stand.

## Summary by CodeRabbit

- **New Features**
  - Added analytics for closed issues with unknown classifications.
  - Updated task-delivery views, heatmaps, drilldowns, and metric results to distinguish known tasks from unknown classifications.
  - Added validation to ensure task, bug, and unknown classifications partition closed issues correctly.

- **Breaking Changes**
  - Renamed `tasks.closed_non_bug` to `tasks.closed_task`. Update saved queries or integrations using the previous name.

- **Documentation**
  - Updated metric definitions, design documentation, alert text, and terminology.


mitasovr (approved):

cyberantonz:
CodeRabbit triage — both findings valid, both fixed in b8e3c3ef9.

**1. Document the unknown closure measure** (DESIGN.md:96-99, §11.5, Grafana annotation) — FIXED. The PR renamed `closed_non_bug` in all three places without updating the claim that unknown closures carry no signal. Now:
- produced-measures list gains `closed_unknown_type`;
- §11.5 says the unlisted type lands in `closed_unknown_type` (served as `tasks.closed_unknown`), so the bug share moves with a readable cause rather than silently;
- the alert description names `closed_unknown_type` / `tasks.closed_unknown` instead of "mints neither … degrade silently".

**2. Assert `tasks.closed_unknown` per unknown-type path** — FIXED, with one correction. `test_tasks_closed::test_kind_subsets_partition_the_closed_issues` already covers the metric (period + peer + timeseries) on the JIRA path, so the gap was the other classification paths, not the metric. Added:
- `test_github_tasks_bugs` — `closed_unknown` = 1 (the unmapped Incident), so 3 = 1 + 1 + 1;
- `test_github_tasks_bugs_value_map_precedence` and `test_jira_tasks_bugs_value_map_precedence` — `closed_unknown` = 3 (deleted, future and absent rows), so 6 = 2 + 1 + 3;
- `test_tasks_bugs::test_unified_task_bug_metrics_empty_window` — `closed_unknown` = None.

Each of those tests now closes its own partition instead of leaving the remainder implicit.

Adjacent, not for this PR: #3467 (independent branch) replaces the hardcoded `unknown` terminal in gold with `__undecided__` plus a blocking dbt test. The agreed follow-up when the two meet is a fourth measure `closed_undecided`, not widening `closed_unknown_type`.

Verified on a compose datapath stand rebuilt from this head: `tests/datapath/metrics/tasks` — `54 passed in 835.48s`. `SKIP=cargo-clippy pre-commit run` clean (clippy skipped: it fails on untouched files under the local rustc 1.98).

cyberantonz:
Dropped the redundant `bugs_fixed + closed_task > tasks_closed` leg from the subset guard: with non-negative counts the exact partition `bugs_fixed + closed_task + closed_unknown_type = tasks_closed` already implies it, so it could never fire on its own. Renamed the test to `assert_task_kind_subsets_partition_closed` to match what it now asserts; tags, severity, store_failures and selector membership are unchanged.

cyberantonz:
Rebased onto `main` after #3463 (resolution classes) landed; single signed-off commit, force-pushed.

**Conflicts, one line each**

- `task_metric_evidence.sql` — kept both axes in the ARRAY JOIN: the kind branches (`bugs_fixed` / `closed_task` / `closed_unknown_type`) and #3463's resolution branches (`closed_fixed` / `closed_duplicate` / `closed_wontfix`).
- `gold/schema.yml` — accepted measure keys are the union of both sets.
- `registry.yaml` — `tasks.bugs_fixed` keeps the `resolution` breakdown #3463 gave it; `tasks.closed_task` / `tasks.closed_unknown` follow it with no dimensions.
- `builtin.rs` — metric count recomputed against main's new total: 108 (105 + #3463's three) + `tasks.closed_unknown` = 109.
- `passports.md` — regenerated with `analytics passports`, never hand-edited; the in-sync test passes.
- `drilldown_matrix.py` — both metric sets listed.
- `assert_task_bug_subsets_bounded.sql` (delete/modify) — this PR replaces the kind half with the exact-partition guard `assert_task_kind_subsets_partition_closed`. #3463 had added a resolution bound to the same file, so that half now lives in its own test, `assert_task_resolution_subsets_bounded`, unchanged in meaning: the unknown resolution class mints no measure, so the resolution axis is a bound and only the kind axis closes. The partition guard is deliberately NOT extended to the resolution measures — different axis.

**Follow-up when #3467 lands too**: its `__undecided__` sentinel means a fourth kind branch, `closed_undecided` (`issue_kind = '__undecided__'`), gets added to the evidence and to this guard's sum so the partition stays exact. Not widening `closed_unknown_type` to `NOT IN ('bug','task')` — the sentinel is a configuration fault, not a kind.

PR #3467, opened 2026-09-17
fix(task): require an operator default decision for every standardized field

cyberantonz:
**Why.** `task_issue_state` ends its issue-kind chain at a hardcoded `'unknown'`. A tenant that *decided* a type is unknown and a tenant whose `config.field_value_defaults` row was never authored produce identical rows, so a lost decision looks like a real answer.

**What changed.** Nothing in the data model — `issue_kind` stays `bug` / `task` / `unknown` and the `'unknown'` terminal stays. Two signals now catch the missing decision instead:

- `assert_task_field_value_defaults_exist` fails the gold build when a live `(tenant, source, field)` has no current `config.field_value_defaults` row.
- The Grafana rule `insight-field-value-defaults-missing` reports the same condition in a deployed install.

**Scope of the test is CI, not production.** It is `severity='error'` and untagged, so `dbt build --select tag:gold` picks it up (the data-path rig and stand builds) and the in-cluster `dbt test --selector data_quality` job does not. A permanently-red scheduled job stops being read and masks real breakage, and a deployment that deliberately ships no field-value-map bundle would keep it red forever.

**The alert has no share and no floor.** `insight-issue-types-unclassified` measures how much of a source lands unclassified and rides out a backfill at `for: 6h`. A decision that does not exist is a fault at one occurrence, so the new rule is `count > 0` at `for: 5m` — long enough to cover the gap between a deploy and the first sync. It is config-only: it counts no issue rows, so it fires on the fault rather than on its consequences.

**Which fields are demanded.** A source is asked for a default only where the role is bound: Jira binds both `issuetype` and `resolution` by product default, other vendors bind what they authored. A field nothing binds is not standardized and needs no decision.

**Bitemporal reads follow gold's shape** — latest row per key first, tombstone and domain filter after. Filtering before `LIMIT 1 BY` can resurrect a retracted decision. `insight-issue-types-unclassified`'s CTEs are corrected to that shape while here.

**Seeds and fixtures carry the decision they would demand of an operator.** The seed tool writes `field_value_defaults` rows for every task source and field; every task data-path spec seeding issues gets one.

**Out of scope.** The Grafana file here is the public sample; the private gitops repo mirrors it and needs the same rule.

**Merge order.** #3471 adds `insight-task-config-tables-missing` at a nearby insertion point and this rule's `execErrState: OK` depends on it, so #3471 merges first; the remaining conflict is one hunk at the tail of `insight-issue-types-unclassified`'s description.

**Verified.** `dbt parse`; `dbt ls` confirms the test is in `tag:gold` and absent from the `data_quality` selector; `audit_rmt_read_dedup.py` gap list identical to `upstream/main`; seed tests 267 passed; data-path `tests/datapath/metrics/tasks`.

## Summary by CodeRabbit

* **New Features**
  * Added required default classifications for unmapped issue types and resolutions.
  * Added monitoring for missing or invalid field defaults.

* **Bug Fixes**
  * Improved classification accuracy when configuration decisions are updated or retracted, preventing older decisions from being restored incorrectly.
  * Gold builds now fail clearly when required defaults are missing or outside supported value domains.

* **Tests**
  * Updated task-processing scenarios and seed data to include explicit default classifications.


cyberantonz:
Rebased onto `main` after #3463 (resolution classes) landed; single signed-off commit, force-pushed.

**Conflicts**: `task_issue_state.sql` (header comment — both narratives merged) and `gold/schema.yml` (the `issue_kind` accepted values sat next to the `resolution_kind` column #3463 added).

**Semantic adjustment the rebase required.** #3463's resolution chain ended at the same hardcoded `'unknown'` this PR exists to remove, so the sentinel extends to it:

- `resolution_kind`'s terminal is `'__undecided__'` — a resolution value that reaches neither a map row nor a default row is nobody's decision, exactly as for the kind.
- The deliberate carve-out from #3463 stays: an issue with NO resolution value at all resolves to `'unknown'` directly, without consulting the default. An absent value is not a missing decision.
- `assert_task_kind_decision_exists` now covers both fields (a second UNION arm, grouped by `field`, reporting type ids for `issue_type` and issue refs for `resolution`).
- `accepted_values` for `resolution_kind` gains `__undecided__`.
- The seed emits a `config.field_value_defaults` row per (source, field) rather than for `issue_type` alone. No datapath fixture needed a new resolution default — the one spec that seeds resolution values already carries one.

**Not done, on purpose**: the Grafana rule `insight-issue-kind-undecided` still measures the issue-type axis only; the resolution axis is covered by the blocking dbt check. Worth a follow-up if we want the production half symmetric.

**Follow-up when #3465 lands too**: a fourth measure `closed_undecided` (`issue_kind = '__undecided__'`) gets added to the evidence and to that PR's partition guard's sum, so the kind partition stays exact. Not widening `closed_unknown_type` to `NOT IN ('bug','task')` — the sentinel is a configuration fault, not a kind.

cyberantonz:
Rebased onto main (post-#3465) and **re-scoped down** after owner review. The `__undecided__` sentinel is gone.

**What was dropped.** The sentinel value in `task_issue_state` (both axes), the accepted_values additions, and the sentinel-shaped blocking test. `issue_kind` stays three values, the hardcoded `'unknown'` terminal stays, and nothing touches `task_metric_evidence.sql`, either kind/resolution guard, the registry, builtin.rs, passports, the FE or drilldown_matrix. #3465's exact-partition equality is untouched and still holds.

**Why.** "No decision exists" is a configuration fault, not a category of work, so it does not belong in a served metric — a fourth measure or a fourth `issue_kind` value would have paid model and UI cost for a state that must never occur.

**What replaces it.** The same fault, caught twice, without a new data value:

- `assert_task_field_value_defaults_exist` — severity error, fails the gold build when a live `(tenant, source, field)` has no current `config.field_value_defaults` row. A source is asked only for the fields whose role is bound, so GitHub is not asked for a resolution default.
- `insight-field-value-defaults-missing` — the deployed signal, config-only, no share and no floor, `for: 5m`.

**The test is CI-scope on purpose.** Untagged, so `dbt build --select tag:gold` picks it up (data-path rig, stand builds) and the in-cluster `dbt test --selector data_quality` job does not. A deployment that ships no field-value-map bundle would hold that job permanently red, and a permanently red job stops being read.

**Merge order.** #3471 first: it adds `insight-task-config-tables-missing`, which this rule's `execErrState: OK` relies on. One conflict remains after it lands, at the tail of `insight-issue-types-unclassified`'s description.

mozhaev-dev (approved):

cyberantonz:
CodeRabbit's review ran against `7f0c963d`, the pre-rescope commit. The `__undecided__` sentinel it reviewed is gone — the owner rejected it as over-engineering, `issue_kind` stays bug/task/unknown, and the fault is caught by a dbt test and an alert instead. Findings triaged against the current tree anyway.

**1. `role_bindings`: resolve the latest bitemporal row before filtering tombstones — real, not fixed here.**

Correct on the merits. `config.task_field_roles.unique_key` carries `valid_from` and `recorded_at`, so `FINAL` collapses only a decision written twice; `WHERE is_deleted = 0` ahead of `argMax(role, (valid_from, recorded_at))` drops the tombstone and lets `argMax` return the role it retracts. That is the same bug this PR fixes for `field_value_map` and `field_value_defaults`.

It is not fixed in the alert because the alert deliberately mirrors gold. `src/ingestion/gold/task_field_roles_current.sql`'s `authored` CTE has the identical shape, and both the new dbt test and `task_issue_state` read `task_field_roles_current`. Correcting the alert alone would make it demand a default for fields gold does not standardize, and disagree with the model it guards. The fix belongs in one change over `task_field_roles_current.sql` plus both alert CTEs, with a tombstone fixture to prove it — its blast radius is `task_issue_state`, `task_status_spans` and the datapath expectations, none of which this PR touches.

**2. `role_bindings`: include `data_source` in the grouping and join — declined.**

`insight_source_id` is a connector-instance identifier (`docs/shared/glossary/README.md:162`: "Connector instance identifier -- distinguishes between two GitLab instances for the same tenant"), so it functionally determines `data_source`; two sources cannot share one. The rule never joins on it either — `data_source` comes out of `observed` (`SELECT DISTINCT insight_source_id, data_source FROM silver.class_task_field_history`), and `authored_roles` is matched on `(insight_source_id, role)`. The new dbt test resolves it the same way. Gold's join carries `data_source` as defensive redundancy, not because the key needs it.

**3. `execErrState: OK` — the flip stands, the comment was wrong and is fixed.**

The rule the finding names (`insight-issue-kind-undecided`) no longer exists, but the objection carries to `insight-field-value-defaults-missing`. `Error` is exactly what #3471 exists to stop: below umbrella 0.7.36 the `config` tables are absent, every evaluation throws `UNKNOWN_TABLE`, and Grafana pages a `DatasourceError` naming no action. `insight-task-config-tables-missing` (#3471) reads only `system.tables`, resolves with the `config` database absent, and keeps `execErrState: Error` — so a datasource outage still pages there.

The claim you disproved was mine, in the comment: "every other breakage by insight-clickhouse-failed-queries". That rule fires above 0.5 failed queries/s over 10m and one failed evaluation per 5m is ~0.0033/s, so it covers nothing here. Comment corrected, and the annotation now states that `execErrState=OK` makes silence non-evidence for any evaluation error, not just a missing table (5ad7d70d3).

**4. Report the vendor issue-type ID as `sample_key` — moot.**

`assert_task_kind_decision_exists.sql` was dropped in the rescope. `assert_task_field_value_defaults_exist` returns `(tenant_id, insight_source_id, data_source, field)`, which is the `config.field_value_defaults` key an operator writes the row against — no display name involved.

**Merge order.** #3471 first. `git merge-tree` reports one conflict, at the tail of `insight-issue-types-unclassified`'s description, and this PR's `execErrState` comment references #3471's rule.

Head is 5ad7d70d3. YAML-only change: `yaml.safe_load` round-trip, 13 rules, placement and per-rule `execErrState`/`noDataState`/`for` identical to 38e6a2b42.

cyberantonz:
The `role_bindings` CTE in `insight-field-value-defaults-missing` now uses the corrected latest-row-then-tombstone shape from #3478.

cyberantonz:
Rebased onto main after #3471: kept #3471's added sentence on the insight-issue-types-unclassified description verbatim and appended insight-field-value-defaults-missing after it, so the insight-data group now reads guard rule -> coverage -> missing-default; also trimmed this PR's execErrState comment to defer to the neighbouring rule instead of restating it, and extended #3471's guard-rule runbook text to name both rules its missing tables blind.

cyberantonz:
Merge-queue run [35376624726](https://github.com/constructorfabric/insight/actions/runs/35376624726) failed in **Data path (rest)**, step *Run the data-path shard*:

```
FAILED tests/datapath/meta/test_reset.py::test_the_relations_the_stand_seed_owns_are_the_ones_it_declares
E   AssertionError: assert {('bronze_bam...tories'), ...} == frozenset({('...ories'), ...})
E     Extra items in the left set:
E     ('config', 'field_value_defaults')
```

Mine, not a collision and not a flake. This PR added `("config", "field_value_defaults")` to the seeder's `RESET_TARGETS` but not to its mirror `SEED_OWNED` in `tests/lib/insight_datapath/reset.py`; the meta test exists to catch exactly that divergence. It lives in the `rest` shard, which the tasks-tree run I validated against does not cover.

Fixed by adding the relation to `SEED_OWNED`, and rebased onto current `main` (which now carries #3478). The two PRs did collide in `insight-issue-types-unclassified`: #3478 rewrote its `role_bindings` CTE in lines adjacent to mine. Both rules now carry the identical tombstone-safe CTE keyed on `(insight_source_id, data_source, field_id)` — verified byte-equal.

Validated on the rebased branch:

- `dbt parse` — clean (pre-existing `MissingArgumentsPropertyInGenericTestDeprecation` warnings only)
- `audit_rmt_read_dedup.py` — output byte-identical to a baseline captured from `main`; no new gaps
- seed tests — `267 passed, 80 subtests passed in 6.34s`
- datapath `tests/datapath/metrics/tasks` — `59 passed in 1053.33s (0:17:33)`
- datapath `tests/datapath/meta` (the shard that failed) — `149 passed, 5 skipped in 55.89s`
- `values.yaml` `yaml.safe_load` round-trip plus a programmatic rule comparison against `main`: 13 -> 14 rules, one added (`insight-field-value-defaults-missing`, `insight-data` group, immediately after `insight-issue-types-unclassified`, `for: 5m`, `noDataState: OK`, `execErrState: OK`, `severity: critical`), none removed; the only other differences are `insight-task-config-tables-missing` annotations and `insight-issue-types-unclassified` SQL, both intended, plus the index shift of the three rules after the insertion point
- `SKIP=cargo-clippy pre-commit run` — passed

Not re-queued.

PR #3468, opened 2026-09-17
fix(gateway): make the gateway the sole author of X-Forwarded-For

cyberantonz:
**Why.** `/auth/*` and the SPA forwarded a caller-supplied `X-Forwarded-For` unchanged, so the client IP the authenticator records against a session was caller-controlled.

**What changed.** The six locations that run no access-phase Lua author `X-Forwarded-For` from `$remote_addr` instead of appending, and the authenticator reads the last hop. `/api/*` is unchanged — its Lua clear already produced this.

**Verified.** routegen 24 and authenticator 103 tests; the new golden test fails, naming `/auth/`, when the defect is reinstated; generated config passes `openresty -t` in the gateway image. Workspace clippy already fails on `main` independently of this branch, so the pre-commit hook was bypassed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved client IP attribution by using the gateway-authored forwarded address.
  * Prevented client-supplied forwarding data from being carried through authentication, metadata, and SPA routes.
  * Added validation to ensure trusted gateway network ranges and proxy addresses are correctly identified.

* **Deployment**
  * Gateway trusted IP ranges are now resolved from the live cluster when not explicitly configured.
  * Added support for passing additional Helm settings during rendering and deployment.

* **Tests**
  * Added coverage verifying forwarded-address handling and generated proxy configuration behavior.


cyberantonz:
Both findings reviewed. The first is correct and is fixed in the force-pushed revision.

**1. Node listing in CI — valid, fixed.** `deploy-test-stand.yml` runs `make -C deploy/gitops deploy ENV=test-stand`, no public-sample env pins `gateway.setRealIpFrom`, and the CI deployer is a namespace-scoped `Role`. `nodes` is cluster-scoped, so that call could never succeed and the deploy aborted before Helm.

The design error was hard-failing on *no answer* rather than on a *wrong answer*. Discovery is now best-effort: when the pod network cannot be read it warns and leaves `setRealIpFrom` unset, so the deploy proceeds exactly as it does today. A resolved-but-unsafe answer — non-private, implausibly wide, or not covering the running Envoy pods — still blocks. Verified all three paths, plus that the Makefile writes an empty args file rather than `setRealIpFrom={}` when discovery is skipped. No RBAC change, since granting a namespaced deployer cluster-scoped node read to populate an optional value is the wrong trade.

**2. In-cluster forging — accurate, and a known limit rather than a regression.** The gateway Service is ClusterIP with no NetworkPolicy, so a workload pod sits inside the trusted range and can forge `X-Forwarded-For`. Trusting only Envoy source CIDRs does not fix it: Envoy pod addresses change on every restart, and `routegen` reads this once at gateway startup, so a pinned address goes stale silently. A NetworkPolicy admitting the listener only from Envoy pods is the real mitigation and is larger than this branch.

Worth weighing against what it replaces: before this change `setRealIpFrom` was unset everywhere, so realip was off and *any external caller* could forge the address the authenticator records, while the flood guard was one global bucket that a single client could exhaust for everyone. After it, forging requires in-cluster pod access and the value is used for rate-limit keying and session attribution only, never authorization. Net narrower, not wider — but the residual vector is real and the NetworkPolicy is worth tracking separately.

mozhaev-dev (approved):

PR #3471, opened 2026-09-17
fix(grafana): report a missing task-config table instead of paging

cyberantonz:
**Why** — `Closed issues landing unclassified` fired on cfabric as a `DatasourceError` at 2026-09-16 00:26: `Unknown table expression identifier 'config.field_value_map'`. The rule shipped 2026-09-14; the DDL (dbt macro `create_task_config_tables`, insight#3430) first reached the umbrella in chart **0.7.36** on 2026-09-15. cfabric pins **0.7.34**, which creates only `config.task_field_roles` and `config.task_value_map`. The rule is ahead of the deployment — nothing failed, and the page repeats every interval naming no action.

**What changed** — new rule `insight-task-config-tables-missing` (group `insight-data`, folder `Insight Alerts`) queries `system.tables` only, so it always resolves — including when the `config` database itself is absent. It names the missing tables and carries the chart floor as the remedy. `insight-issue-types-unclassified` moves to `execErrState: OK`.

**Why not guard the query itself** — ClickHouse resolves table names at analysis time; there is no short-circuit. `merge()` over a non-matching regexp throws too (`CANNOT_EXTRACT_TABLE_STRUCTURE`, and `UNKNOWN_DATABASE` for a missing database), verified against a local dump on 25.7.5.34.

**Why `execErrState: OK` is not silence** — an errored evaluation yields no series, so a missing table can never read as "0% unclassified"; the guard rule is the standing signal that coverage is UNMEASURED. Every other query breakage still surfaces in `insight-clickhouse-failed-queries`, which rides the VM datasource and does not depend on the ClickHouse query succeeding.

**Out of scope** — `insight-issue-kind-undecided` (#3467, still open) needs the same one-line `execErrState` flip once it lands; it inherits the guard rule automatically. The private gitops copy of this file gets the same change, alongside the resolution-coverage sibling.

**Verified** — YAML round-trips through `yaml.safe_load`; group/folder unchanged; the guard SQL extracted back out of the YAML runs against a local ClickHouse dump and returns zero rows when all three tables exist and the missing names when they do not. No cluster access used.

## Summary by CodeRabbit

* **New Features**
  * Added monitoring to detect missing configuration tables and report which tables are unavailable.

* **Bug Fixes**
  * Prevented issue-type classification alerts from entering an error state when their underlying query fails.
  * Clarified alert guidance for deployments using older chart versions where the required configuration tables are unavailable.


mitasovr (approved):

mozhaev-dev (approved):

cyberantonz:
**Keep persistent query failures observable — half right, and the half that is right was my comment.**

The flip to `execErrState: OK` stands. `Error` is the behaviour this PR exists to remove: below umbrella 0.7.36 (0.6.102 on the 0.6 line) the `config` tables do not exist, ClickHouse resolves names at analysis time so every evaluation throws `UNKNOWN_TABLE`, and Grafana pages a `DatasourceError` that names no action. That is the realistic failure and it is now reported with its remedy by `insight-task-config-tables-missing`, which reads only `system.tables`, resolves with the `config` database absent, and keeps `execErrState: Error`. A datasource outage or an unreachable ClickHouse therefore still pages — through the guard rule, on the same datasource.

Your arithmetic on `insight-clickhouse-failed-queries` is correct and disproves a claim of mine, not the design: it evaluates `sum(rate(ClickHouseProfileEvents_FailedQuery[10m])) > 0.5` over 10m, and one failed evaluation per 5m is ~0.0033/s. The comment that routed "any other query error" there was wrong. Fixed in b73ff163c — the comment now says only what is true, the commit message drops the same claim, and the annotation states that `execErrState=OK` makes silence non-evidence for any evaluation error, not just a missing table.

What is left uncovered after that is narrow: ClickHouse healthy, `system.tables` readable, the config tables present, and this query alone erroring. It yields no series, so the rule reads silent, never clean, and the annotation says so. A dedicated per-evaluation health alert is a reasonable follow-up across every ClickHouse-backed rule in this group; it is not this PR.

Head is b73ff163c. YAML-only change: `yaml.safe_load` round-trip, 13 rules, placement and per-rule `execErrState`/`noDataState`/`for` identical to 2a9671d85. Approvals survive — the ruleset has `dismiss_stale_reviews_on_push: false`.

PR #3477, opened 2026-09-18
perf(jira): speed up Jira field history tests

mitasovr:
**Why** — the Jira field-history transformation lane took about half an hour and sat against its job's 45-minute timeout. Each of the 73 tests truncated bronze, ran `dbt` as a subprocess and rebuilt the chain.

**What changed** — a module's scenarios declare their bronze with a `@case` decorator, are seeded together and built **once**, each under a source id of its own. `dbt` is invoked in the test process rather than as a command. No coverage was traded away: the same 73 tests assert the same things.

**The cost was never the data.** A build costs ~15 s whatever it builds. `dbt-clickhouse` materializes a table as `CREATE TABLE … EMPTY AS (query)` and then `INSERT … (query)`, so the query is analysed twice, and analysing this one takes ~7 s on **zero rows** — `EXPLAIN AST` 0.08 s, `EXPLAIN QUERY TREE` 6.0 s. Its CTEs are inlined at each use, so 1070 lines of SQL expand into 150 query-tree nodes. The suite's runtime was the *number* of builds, not their content. A further ~9 s per call went to starting an interpreter and parsing the project.

**Isolation is `insight_source_id`** — every model in the chain carries it through its joins and windows, so scenarios sharing a build never see each other. Two reads that are not addressed by source were scoped explicitly: the unclassified-field registry by `insight_source_id`, and the content-addressed text table by the address the journal hands back, since it carries no source at all.

**Scenarios that cannot share** carry no `case` and get bronze to themselves, exactly as every test used to. Five need it, each with the reason written beside it: one builds twice, one expects a build to fail, one rewrites bronze half way through, and two are meant **not** to reconcile — `round_trip_holds()` asks about the whole warehouse, so those two would otherwise report themselves as a defect in every neighbour sharing their build. The README now names the tell: a test that does not assert `round_trip_holds()` is the one to check.

**Out of scope** — the double analysis and the CTE fan-out are paid by every production build too, twice per model. Making that query cheaper to analyse is a separate change against the model, not the tests.

**Verified** — 73 passed in 8:26 against a pinned ClickHouse 25.7.5 and the pinned dbt, down from ~30 min. Per-test cost inside a batched module fell from 15–27 s to 0.01–1.4 s.

One correctness note worth keeping: every dbt call gets a **freshly parsed manifest**. Reusing one is about a second faster and wrong — dbt marks a node compiled as it builds, so a second run of a model with an ephemeral parent stops getting the parent's CTE injected and references a name that was never defined. That is recorded as an `INVARIANT:` in `conftest.py`.

## Summary by CodeRabbit

- **Tests**
  - Expanded Jira field-history validation across multi-value fields, links, ordering, deduplication, retired fields, synthetic rows, and unclassified fields.
  - Added coverage for empty values, repeated changes, re-added elements, project moves, and stable field identifiers.
  - Standardized scenario setup and improved test execution reliability.

- **Documentation**
  - Updated test instructions, commands, scenario examples, and guidance for shared test runs and expected round-trip exceptions.


src/ingestion/dbt/tests/jira/transform/conftest.py
```diff
@@ -64,6 +66,52 @@ def _env(name: str) -> str:
...
+        self._log.clear()
+        runner = dbtRunner(callbacks=[lambda event: self._log.append(event.info.msg)])
+        result = runner.invoke([*args, *self._base])
+        statuses = tuple(str(node.status) for node in getattr(result.result, "results", ()))
+        log = "\n".join(self._log)
+        if result.exception is not None:
+            log = f"{log}\n{result.exception}"
+        return Invocation(success=result.success, statuses=statuses, log=log[-8000:])
+
+    def invoke(self, *args: str) -> Invocation:
```
cyberantonz, line 111: What the reason to have this method?

cyberantonz (approved):

PR #3478, opened 2026-09-18
fix(task): let an operator retract a field-role binding

cyberantonz:
**Why.** An operator cannot unbind a task field. `config.task_field_roles` is bitemporal — `unique_key` carries `recorded_at`, so a retraction is a newer separate row, not an overwrite — and role resolution filtered `is_deleted` before picking the latest row per key, throwing the tombstone away and returning the binding it retracts.

**What changed.** `task_field_roles_current`'s `authored` CTE and the Grafana rule's `role_bindings` CTE now pick the latest row per key first (`ORDER BY valid_from DESC, recorded_at DESC` + `LIMIT 1 BY`) and read `is_deleted` on the winner — the idiom `task_issue_state`'s decision CTEs already use.

**The sample alert moves with the model.** Leaving it on the old shape would have it disagree with the model it guards. The private gitops repo mirrors `deploy/gitops/system/grafana/values.yaml`, so a mirror MR is needed there before a deployed Grafana picks this up.

**Out of scope.** The singular dbt tests under `src/ingestion/dbt/tests/task/` resolve bindings with the same pre-filter shape. They are guards rather than data path and correcting them changes what CI fails on, so they want their own PR.

**Verified.** New datapath spec `github_tasks_role_retraction` — a retracted `issuetype` binding leaves both closures unclassified, and it fails against the pre-fix model. `tests/datapath/metrics/tasks` 58 passed (57 before this branch, none changed). `dbt parse` clean; `audit_rmt_read_dedup.py` output byte-identical to the `upstream/main` baseline; seed tests 265 passed; `values.yaml` round-trips with the same 12 rule uids; pre-commit clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected role-binding resolution so newer retractions properly override earlier assignments.
  - Updated task metrics to stop applying withdrawn classifications while preserving unaffected bindings.
  - Improved handling of unclassified task activity when no active type mapping remains.
  - Fixed alert evaluation to respect the latest configuration state, including deleted bindings.
  - Ensured data-source-specific bindings are resolved independently and accurately.

- **Tests**
  - Added coverage for retracted GitHub task-role bindings and resulting metric classifications.


cyberantonz:
CodeRabbit triage, one finding.

**Preserve `data_source` in `role_bindings` — REAL, fixed in 7106197af.**

Confirmed on the merits, not on the bot's say-so. `insight_source_id` is not a minted key: `reconcile_compute_source_id` (`src/ingestion/reconcile-connectors/lib/connector-naming.sh`) falls back to the literal `main` when no Secret carries a `source-id` annotation, so in a default deployment every connector — GitHub, Jira, Slack, Zoom — ingests under the same `insight_source_id`. One instance also emits more than one `data_source` by design (`insight_github` in the git class models, `github` in the task class models). Nothing in DDL, tests or a registry enforces the dependency the bare key assumes.

So a newer tombstone under one `data_source` could suppress a live binding under another. Both the CTE key and the `hist` join now carry `data_source`, matching what `task_field_roles_current`, `task_issue_state` and `task_status_spans` already do. Mirrored into #3467's `insight-field-value-defaults-missing` rule (`role_bindings` -> `authored_roles` -> `bound`), which carried a fresh copy of the same omission; whichever PR merges second resolves the textual conflict in `values.yaml`.

**Out of scope, named in the PR body.** The singular dbt tests under `src/ingestion/dbt/tests/task/` (`assert_every_field_is_bound_or_ignored`, `assert_one_field_per_role`, `assert_bound_values_are_mapped`, `assert_canonical_values_match_their_role`) carry the same pre-filter shape. They are CI guards on raw relation names, the datapath rig never runs them, and correcting them changes what CI fails on. Own PR.

**Also in this push, unprompted:** the four-line comments above the corrected CTEs are compressed to one tagged `INVARIANT` line each, in `task_field_roles_current.sql` and both `values.yaml` rules; the same pass trimmed fix-narration from the new spec yaml, its test module docstring and the `role_type_retracted` template comment. Alert `summary`/`description` annotations untouched.

**Validation.** `dbt parse` clean. `audit_rmt_read_dedup.py` byte-identical to the `upstream/main` baseline (both exit 1 on the same pre-existing gaps). `values.yaml` round-trips on both branches with rule count, uids, `for`, `execErrState`, `noDataState`, labels and annotations unchanged — the only rules whose bodies differ are the two intended. pre-commit clean. No datapath run: the code change is confined to a Grafana alert definition, which the rig does not load, and the model changed only in comments — the 58-passing `tests/datapath/metrics/tasks` run on 5077f006 still stands.

mitasovr (approved):

